PR #192, opened 2014-06-26
RFC: dexpect and container manager tools

ldoktor:
Hi guys,

this is a newer version of https://github.com/autotest/autotest-docker/pull/179 redesigned using @jzupka's (and others) input.

The idea is to use existing commands as much as possible and add some helpers, to significantly simplify container initialization, cleanup and interaction in general.

Main functions are:
* keep track of all created objects (containers, images for now)
* automatic cleanup for those objects
* deterministic way to get new container
* interactive variant of `AsyncDockerCmd`
* possibility to get pexpect-like `ShellSession` control of `InteractiveAsyncDockerCmd`-like command

Differences from previous version:
* `dexpect` doesn't execute commands itself, but accepts existing `InteractiveAsyncDockerCmd`-like commands (requires support some basic functions)
* `dexpect` doesn't create additional `tty` layer, which makes `non-tty` container interaction work correctly.
* each (Sub)Subtest has `DockerContainers`, `DockerImages` and `DockerManager` initiated during `__init__` phase (almost every test does this during `initialize`, it's easier to put it there for every test)
* `DockerManager` is separate module, which focuses only on the 90% of the usage without nasty hacks or complicated interfaces.

Booth demonstration tests are modified to work with the latest version. The `interaction` test fails due of know bug: https://bugzilla.redhat.com/show_bug.cgi?id=1113085

I left the original patches there for people, who already saw previous version to see only the differences. Please use the `files changed` view to see the overall changes. I'll add docstrings, unittests and rebase everything once we agree on the design.

### DIFF
--- a/config_defaults/subtests/docker_cli/detach.ini
+++ b/config_defaults/subtests/docker_cli/detach.ini
@@ -9,10 +9,12 @@ container_command = bash
 [docker_cli/detach/basic]
 
 [docker_cli/detach/stress]
+# TODO: BUG: When sleep is removed, change this to 100
 no_iterations = 3
 session_log = STORE
 
 
 [docker_cli/detach/multi_session]
+# TODO: BUG: When sleep is removed, change this to 100
 no_iterations = 3
 session_log = STORE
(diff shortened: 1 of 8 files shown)


cevich:
Does not meet current design style / goals for project.

PR #257, opened 2014-08-08
Daemon dns, group, ip, iptable, bridge etc

jzupka:
based on https://github.com/autotest/autotest-docker/pull/256


cevich:
Docker daemon tests are currently out of scope, will re-open if that ever changes.

PR #436, opened 2014-12-18
Pipedkrcmd

cevich:
This adds a common class for automatically opening and setting up a pipe through to a child-process running a docker command.  At least five current tests use something very similar, which I plan to replace with this new version, in this PR (eventually).  Several changes/updates are also made to the more basic ``DockerCmd`` stuff to help control them being overly verbose, yet retain info for debugging purposes.  Finally, it sets up a abstract method for the actual command execution phase, de-coupling it from the logging and preparation details.  This should make future expansion easier.


cevich:
Moving to 0.8.2 because there are no dependent tests in 0.8.1

PR #444, opened 2015-01-19
docker_cli/build:  Implement command pattern for postprocessing

cevich:
The specific behavior details of the postprocessing steps
is highly sub-subtest dependent.  Assign simple "commands"
for each step, then let configuration specify which and in
what order they must all pass.


cevich:
**NOTE: This is incomplete / work in progress.**

Issue #409, opened 2014-11-12
Add pylint disable comments


cevich:
Several places in ``subtest`` module have to use the catch-all ``Exception``, however they are queued and re-raised later.  These pylint disable cases simply need to be better commented (why it's disabled) so future maintainers aren't confused.  Suggest:

```python
# Catch-all exceptions here so they can be re-raised later, after all stages complete.
# pylint: disable=W0703
```

cevich:
Fixed as part of #466

PR #466, opened 2015-02-02
Introduce additional test trees for pretests, intratests and posttests

cevich:
These trees contain tests that run before, during, and after all other subtests.  This allows host and environment setup/tear-down to become part of the overall testing process.  The control file was re-structured to be more object-oriented, and iterate over each test step in turn instead of adding them all first.  Finally, envchecks has been replaced by a new ``garbage_check`` intratest test that does the same job.

resolves: #409


cevich:
**WORK IN PROGRESS**

PR #476, opened 2015-02-11
Fix cgroups

cevich:


lmr:
Hey Chris, I've looked at the output of your Travis job and couldn't figure out exactly what was wrong. I just want to let you know that a new PEP8 python checking program was out and some new checks started to find problems in virt-test, test providers and avocado. I've fixed them, then came to your jobs to try to figure out the problem and fix it, but your failures don't look related to the PEP8 problems.

If you need any help, let me know.

cevich:
@lmr Yep, I noticed those pep8 problems in other PRs of mine also.  You're right though, the failures here are unrelated.  They happen because the pep8/pylint versions in Ubuntu used by Travis differ from what's available on RHEL.  A headache I'm use to dealing with.  Thanks for checking though!

lmr:
@cevich, one way to deal with it would be to install the tools via `pip` rather than the package manager. Not clean, I know, but it gets the job done.

cevich:
@lmr clean is more important than the time it takes to fix Travis problems I'd need to fix anyway :grinning:

PR #487, opened 2015-03-05
Dockerfile: Support building/testing in Atomic

cevich:
SPC: Super Provileged Container


cevich:
Note: Still need to test this on RHEL Atomic Host and add support for LABEL INSTALL & UNINSTALL

PR #495, opened 2015-04-21
Fix attach case issue

luwensu:
In this case, it will execute docker run command with a 5s timeout and no pre-check
whether the image exist.The patch will introduce two functions to check and pull the image
if it does not exist.

*Add pre-check of whether image exist before run
*Add pull image function if the image is not exist

Signed-off-by: Time Su <lsu@redhat.com>

### DIFF
--- a/subtests/docker_cli/attach/attach.py
+++ b/subtests/docker_cli/attach/attach.py
@@ -36,10 +36,32 @@ class attach_base(SubSubtest):
                      % wait)
         time.sleep(wait)
 
+    def image_exist(self, image_name):
+        di = DockerImages(self)
+        image_list = di.get_dockerimages_list()
+        exist_status = DockerImages.filter_list_full_name(image_list,
+                                                          image_name)
+        return exist_status
+
+    def pull_image(self, image_name):
+        dkrcmd = AsyncDockerCmd(self, 'pull', [image_name],
+                                self.config['docker_timeout'],
+                                verbose=True)
+        self.loginfo("Executing background command: %s" % dkrcmd)
+        dkrcmd.execute()
+        while not dkrcmd.done:
+            self.loginfo("Pulling...")
+            time.sleep(3)
+        self.failif(dkrcmd.exit_status != 0,
+                    "Fail to download image %s"
+                    % image_name)
+
     def initialize(self):
         super(attach_base, self).initialize()
         self.sub_stuff['subargs'] = self.config['run_options_csv'].split(',')
         fin = DockerImage.full_name_from_defaults(self.config)
+        if self.image_exist(fin) == []:
+            self.pull_image(fin)
         self.sub_stuff['subargs'].append(fin)
         self.sub_stuff['subargs'] += self.config['bash_cmd'].split(',')
         self.sub_stuff['subargs'].append(self.config['cmd'])


subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
                      % wait)
         time.sleep(wait)
 
+    def image_exist(self, image_name):
+        di = DockerImages(self)
+        image_list = DockerImages.get_dockerimages_list(di)
```
cevich: Use object-oriented syntax here:
``image_list = di.get_dockerimages_list()``

subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
...
+        exist_status = DockerImages.filter_list_full_name(image_list,
+                                                          image_name)
```
cevich: here too:
``exist_status = di.filter_list_full_name(image_list, image_name)``

Also, this way if API changes to *not* ``@staticmethod``, no test code update will be needed.

subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
...
+        return exist_status
+
+    def pull_image(self, image_name):
+        print "pull_image"
```
cevich: Okay for debugging, but no ``print`` in final version.  If pulling is critical to the test, use ``self.loginfo()`` otherwise if it's only handy when something breaks, use ``self.logdebug()``.  Those methods make sure the logs are updated and they give control over verbosity level.

subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
...
+        dkrcmd = AsyncDockerCmd(self, 'pull', [image_name],
+                                self.config['docker_timeout'],
+                                verbose=True)
+        self.loginfo("Executing background command: %s" % dkrcmd)
+        dkrcmd.execute()
+        while not dkrcmd.done:
+            self.loginfo("Pulling...")
+            time.sleep(3)
```
cevich: Just FYI: AsyncDockerCmd does not use ``timeout``, but it's okay in this case since test cannot continue w/o image and autotest will timeout the test if it takes too long.  One day I should fix this :confused: in ``AsyncDockerCmd``

subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
...
+        result = dkrcmd.wait()
+        self.failif(result.exit_status != 0,
+                    "Fail to download image %s"
+                    % image_name)
```
cevich: Also FYI (your code is fine):  dkrcmd can represent it's own result, so this would work also:
```python
    self.failif(dkrcmd.exit_status != 0,
                "Fail to download image %s"
                % image_name)
```

For convenience, there are also attributes for ``stdout``, ``stderr``, and ``duration``.

subtests/docker_cli/attach/attach.py
```diff
@@ -36,10 +36,34 @@ def wait_interactive_cmd(self):
...
+
     def initialize(self):
         super(attach_base, self).initialize()
         self.sub_stuff['subargs'] = self.config['run_options_csv'].split(',')
         fin = DockerImage.full_name_from_defaults(self.config)
+        if self.image_exist(fin) == []:
+            self.pull_image(fin)
```
cevich: Oooohhh! I see what you're doing here now WRT e-mail on the subject, it's not just for this one test right?  Oh man, duhhhhh :blush: now I understand!   We **DO** have an API for this!  Adding this stuff as a new "test" (same format) as a [*pretest*](https://github.com/autotest/autotest-docker/blob/master/pretests/pretest_example/pretest_example.py) and/or [*intratest*](https://github.com/autotest/autotest-docker/blob/master/intratests/intratest_example/intratest_example.py]).

A *pretest* will run before any other tests (unless excluded), and an *intratest* will run between every other test.  Then you can also have a single configuration file (or default) that says: ``pull_missing_test_fqin = yes`` and globally control this behaviour for every test that's about to run.

Also, tell me if you think I should [include running all *intratest* after *pretest*](https://github.com/autotest/autotest-docker/compare/autotest:master...cevich:preintra) but before first test.  I'm thinking this would help avoid duplicating same 'pull' test code in both.

Otherwise, if I'm wrong and you just want to fix this one test, it's fine also :smile:

cevich: Hmmm, in fact, maybe this should only be a *pretest* that only runs once at beginning.  Docker-autotest already has common ``clean_all()`` code to protect this default FQIN image.  Then if docker breaks and the image is deleted, the next test will correctly fail.  What do you think?  Ya, I don't think [including intra-into-pre](https://github.com/autotest/autotest-docker/compare/autotest:master...cevich:preintra) is good idea.

luwensu: Yeah, i agree with you after a more deeper play around, a [pretest](https://github.com/autotest/autotest-docker/blob/master/pretests/pretest_example/pretest_example.py) is enough for handle this kind of issue.  
But it may not fit for this case now, since it put the `AsyncDockerCmd(self, 'run', self.sub_stuff['subargs'], verbose=True)` in every base clase's `def initialize`. Yeah, run command executes in initialize.....
Thus i'd like to fix the case now to let it run successfully and use the [pretest](https://github.com/autotest/autotest-docker/blob/master/pretests/pretest_example/pretest_example.py) as the next stage plan. What about it ? :)

luwensu: The another thing i'm concern is about the `class DockerImages(object):`
1.Do we need a `def image_exist()` in the Image/Images class, it just like `def image_exist()` in my patch, to check whether an image exist? I don't see a docker command can do that directly, nor in auotest-docker library. I would write the check in the library but just in case introduce more problems, so i put it in this single case only now.

2.About the `clean_all()`,  assume there is a image full name is `luwensu.com/rhel7:latest` and i give this in `default.ini`:
`preserve_fqins = rhel7`, then `clean_all()` will remove `luwensu.com/rhel7:latest` as it thinks only `rhel7` is preserve.
I see there is a document to explain it should be the full name, but i wonder whether  we need to do more check or give more warning ?
Just not sure about it, if it deserves a talk, we can more it into a single issue.

cevich:
> Thus i'd like to fix the case now to let it run successfully and use the pretest as the next stage plan. What about it ? :)

No problem, let's do it.

cevich:
> 1.Do we need a def image_exist() in the Image/Images class

Yes, and I wouldn't be opposed to also adding a 'pull_if_not_exist()' method there as well :smile:   Make that as a separate PR and I can help out writing unittest and docstrings for it.

cevich:
> 2. About the clean_all(), assume there is a image full name is luwensu.com/rhel7:latest and i give this in default.ini

Sadly it does have to be the complete, full name that you want to protect.  The main reason, is I wanted to keep this method as simple as possible.  At the time I needed it quickly, and more features meant more complex unit-tests.  If you'd like to expand it, maybe to accept regular-expressions and/or IDs I'm open to that improvement (it won't break existing use == no API change).  The brain-hurting work is always in writing the unittests for the new features.  Both the containers and images modules use a lot of mocking and black magic to avoid external and system-state dependencies.  Though I can help writing unittests if you want to get the PR together :smile: (unless you want to try them also, go for it!)

cevich:
Okay, changes here LGTM, nice job.

luwensu:
Haha, i'd like to try all of them, it sounds excited and remains me the time code fp, short lines but every line make the brain run out :). But  the start date may have to delay, since what i'm doing is make all the test case have a green pass, i'll run all tests again tomorrow and show your the result, it left little failed.
Anyway, i'll let you know if i start to work on the library part, and thanks again for your lots of help and i need be more care when push, make lots of simple mistakes this time :(

PR #503, opened 2015-04-23
Fix cp case

### LINKED ISSUE #502: cp test fails
*simple case fails due to it tries to write file to /tmp container which mounted as a tmpfs,
docker changed to this behaviour recently.
*every_last case lack of a postprocess to check the result.

luwensu:
Resolve #502 

*simple subcase: change file directory to /
*every_last: add postprocess function


cevich:
Nicely done, I can't find anything wrong at all :grinning:

PR #513, opened 2015-04-29
docker_cli/logs use utc instead of LC_ALL

luwensu:
The LC_ALL method still fails in my machine, then i have a idea that
use utc instead of config the LC_ALL env.

*Use utc as timestamp thus we can ignore the effect of LC_*


cevich:
Perfect!

PR #532, opened 2015-09-11
new function for checking whether the default image exists

luwensu:
What i see is all kill* sub-case lack of an image check before
the ``container.execute()``, thus the case would fail once the default image doesn't exist.

Using default_image_exist_check() in DockerImage
instance to check whether default image exists before run it.

Or we can move this from the library to a pretest sub case.
But it makes me feel a little bit redundant.

### DIFF
--- a/dockertest/images.py
+++ b/dockertest/images.py
@@ -483,6 +483,30 @@ class DockerImages(object):
         return [di for di in image_list if di.cmp_greedy(repo, tag,
                                                          repo_addr, user)]
 
+    def image_exist_check_by_full_name(self, full_name):
+        """
+        Check whether ``full_name`` has existed.
+        :param full_name: an image's full name.
+        :return: ``True`` if it exists, otherwise ``False`` value will be given
+        """
+        if self.repo_addr_check_by_full_name(full_name):
+            if self.filter_list_full_name(self.list_imgs(), full_name):
+                return True
+            else:
+                return False
+        else:
+            return False
+
+    def image_exist_check_by_default(self):
+        """
+        Check whether default image the sub case uses has existed.
+        **Note** The ``default`` here means the sub case's default,
+        if it's empty, it will be override by the ``Defaults`` section.
+        :return: ``True`` if it exists, otherwise ``False`` value will be given
+        """
+        default_image = self.full_name_from_defaults()
+        return self.image_exist_check_by_full_name(default_image)
+
     def list_imgs(self):
         """
         Return a python-list of DockerImage-like instances
@@ -558,6 +582,22 @@ class DockerImages(object):
         dis = self.list_imgs()
         return [di for di in dis if di.cmp_id(image_id)]
 
+    def pull_default_image(self, timeout=None):
+        """
+        Pull the ``default_image`` from ``full_name_from_defaults()``
+        **Note** The default_image used here is NOT from ``Defaults`` section
+        :param timeout: Override self.timeout if not None
+        """
+        if self.image_exist_check_by_default():
+            return True
+
+        default_image = self.full_name_from_defaults()
+        self.docker_cmd_check("pull %s" % default_image, timeout)
+        if self.image_exist_check_by_default():
+            return True
+        else:
+            return False
+
     def remove_image_by_id(self, image_id):
         """
         Use docker CLI to removes image matching long or short image_ID.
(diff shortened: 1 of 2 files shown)


dockertest/images.py
```diff
@@ -430,6 +430,17 @@ def default_image(self):  # pylint: disable=R0201
         # DICLS may have overriden this method
         return DockerImage.full_name_from_defaults(defaults)
 
+    def default_image_exist_check(self):
```
cevich: How about calling this ``pull_default_image(self)``?

dockertest/images.py
```diff
@@ -430,6 +430,17 @@ def default_image(self):  # pylint: disable=R0201
...
+        """
+        Retrieve default image if it does not exist in
+         ``self.filter_list_full_name()``
+        :return: ``autotest.client.utils.CmdResult`` instance
+        """
+        img_list = self.list_imgs()
+        default_image = self.full_name_from_defaults()
+        if not self.filter_list_full_name(img_list, default_image):
```
cevich: Just use the ``default_image`` instance property (like an attribute), so ``if not self.filter_list_full_name(img_list, self.default_image)``.

dockertest/images.py
```diff
@@ -430,6 +430,17 @@ def default_image(self):  # pylint: disable=R0201
...
+            return self.docker_cmd("pull %s" % default_image, self.timeout)
```
cevich: I understand why you want to return a CmdResult.  However, this API is intended for convenience, and not to be used for test subjects (so 'docker pull' in this case).  Therefore, I think it may be more practical to return the default image's ``DockerImage`` instance instead (the 0-th item from filter-list...() call).  This way the caller has ready-access to the image ID, size, broken-down name, etc.  What is your opinion?  Is there a reason returning CmdResult is better idea?

cevich:
I like this idea in general, esp. for tests that are not exercising 'docker pull' (like the kill tests).  Your code is clean, I just made a few general comments.  Also, do you think you could add a unittest for this one method?  If not, I can help write one as well.

cevich:
Hmmm, I think writing a unittest might be very challenging.  Let me investigate and see if I can come up with one.

cevich:
Okay done, here's a pair of unittests: https://github.com/luwensu/autotest-docker/pull/2

dockertest/images.py
luwensu: It may depend on how we using it
If name it as ``pull``, i'd like to use it like
```
if image not exist:
           pull(self)
container.execute()
```
otherwise, it will looks like
```
images.check
container.execute()
```
i'd rather use the second one since it will appear every case basically, 
just save the duplicate thing, even though the first one looks like better.
Any suggestion?

dockertest/images.py
luwensu: I just wanted to wrap ``check`` and ``pull`` into one single function.
Looks like it's better to split info different part, so you mean we'd like to call ``docker pull`` in the test case and just return the image instance from this function?

luwensu:
Ah...thanks for the unittest, i'm still figuring out how writing it, i'd like to try next time :)
And i will push them together after finish some basic testing in local

cevich:
> I just wanted to wrap check and pull into one single function.
> Looks like it's better to split info different part, so you mean we'd like to call docker pull in the test 
> case and just return the image instance from this function?

Yeah, then two methods would work better;  however, the image objects don't have any context (i.e. connection back to the subtest), they're just very simple objects.  So what I'd suggest then is maybe add another property similar to ``default_image``, maybe ``default_image_obj``?  Then modify your "pull" (whatever it's name is) method to accept ``image_obj`` as a parameter, returning True if it was pulled and False if it already existed (just an idea).

Then it can work like your preferred (first way):

```python
...
di = DockerImages(self)
if di.default_image_obj not in di.list_imgs:
    self.failif(not di.pull_image(di.default_image_obj,
                  "Unexpected background pull of %s" % di.default_image_obj)
...
```

I like this way better now also, you are right to suggest it, I agree.   Please feel free to update the unittests to work with the changes, it will be a good-enough exercise for you :smile: 

In this case, the unit-tests are particular are particularly tricky because they must mock (i.e. "simulate") the configuration, the returned list of images, and all ``utils.run()`` calls so the results can be verified.  That's what RUN_CACHE is for, and why I also added IMAGE_TABLE.  So I suggest maybe accept the PR I sent you, then modify things how you like them, and --force push the changes back to this PR.

(In other words, I'd like to keep the IMAGE_TABLE addition :smile: because it will be helpful in other testing later.)

luwensu:
Sorry for late, i kept changing my mind day after day....
So for now,
1.I decide use ``full_name_from_defaults()`` as if a sub case would use the image in config  that is different from the one in ``default.ini``, it will not miss it.
like:
``docker.io/fedora`` in ``default.ini`` and ``docker.io/centos`` in ``kill.ini``, the original method would trigger an error.

2.Adding a function for checking if the default image is full name.
For instance:
pulling ``fedora`` is not an expect result, but ``docker.io/fedora``, if tries the ahead one, we will not know which registry the image from, it depends on current docker daemon's parameter, the ``--add-registry`` and ``--block-registry``

cevich:
Ignore that comment, it wasn't me, I swear it :grinning:

cevich:
closing this due to waiting over a year for an update.  Please re-open when followup is ready.

Issue #538, opened 2015-10-21
Improve conf.py's compability for Sphinx v1.3.1


luwensu:
If apply the current conf.py for Sphinx v1.3.1, it will trigger few errors:

`Warning, treated as error:
WARNING: the config value 'latex_show_urls' has type `bool', defaults to `str.'
`
`
Warning, treated as error:
WARNING: the config value 'html_short_title' has type `NoneType', defaults to `str.'
`
`Warning, treated as error:
WARNING: 'default' html theme has been renamed to 'classic'. Please change your html_theme setting either to the new 'alabaster' default theme, or to 'classic' to keep using the old default.
`

cevich:
Travis has Sphinx locked down to version v1.2.2 because that's what matches RHEL also.  If you're seeing this error in Fedora then I'd just not worry about it.  The unittests that matter are the ones that run in Travis with the older version.

cevich:
closing this unless it becomes a problem for ``readthedocs.io`, then we can re-open.

PR #539, opened 2015-10-21
Fix conf.py, improve compability for Sphinx 1.3

luwensu:
Improve compatibility and resolve https://github.com/autotest/autotest-docker/issues/538

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -104,7 +104,7 @@ autodoc_member_order = 'bysource'
 
 #: The theme to use for HTML and HTML Help pages.  See the documentation for
 #: a list of builtin themes.
-html_theme = 'default'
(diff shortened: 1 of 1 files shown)


cevich:
Travis is failing with:

```
Theme error:

no theme named 'classic' found (missing theme.conf?)

make: *** [html] Error 1```

I haven't messed with themes at all, maybe adding a theme.conf that somehow binds ``clasic==default`` could work around this.  The older version didn't complain at all about the true/"true"/false/"False" fix.

cevich:
Closing this for now since it's not broken too badly for readthedocs.com nor builds in travis.  Please re-open if more work is needed.

PR #550, opened 2016-01-07
Pre-test reporting occurs before control imported

cevich:
This change resets the sysinfo directory and re-populates it with
test-specific details.

Signed-off-by: Christopher Evich <cevich@redhat.com>


cevich:
Not entirely sure I'm happy with this "solution".  Perhaps a better method is to just add a pretest (existing support) to gather these docker/RHEL/Centos/Fedora specific details.

PR #557, opened 2016-02-04
import subtest: remove --preserve-order tar option

edsantiago:
GNU tar 1.27 [*] allows --preserve-order (aka --same-order)
only with archive reading commands; the import test was
using it with --create, causing test to fail on f22 & f23.

  [*] commit 74ce228


cevich:
Awesome, thanks!

PR #560, opened 2016-02-11
Fantasticly simplify/hone run_env subtest

cevich:
Simple for now but w/ room for future expansion

### DIFF
--- a/subtests/docker_cli/run_env/run_env.py
+++ b/subtests/docker_cli/run_env/run_env.py
@@ -2,494 +2,140 @@ r"""
 Summary
 ---------
 
-This test checks function of `docker run -e VARIABLE=VALUE ...`.
-Also verifies the rm --link operation with ICC.
+This subtest checks functionality of varios environment variable related
+options / operations of docker.
+
 
 Operational Summary
 ----------------------
 
-#. Setup client & server containers w/ exposed port and env. vars.
-#. Run code in containers to verify communication and env. vars
-#. Check results match expectations
+#. Run container with various environment-var related options.
+#. Verify healthy container exit state/status.
+#. Verfiy container environment variables match expectations.
+
 
 Operational Detail
-----------------------
+-------------------
 
-The port test
-~~~~~~~~~~~~~~~
-#.  Starts server with custom -e ENV_VARIABLE=$RANDOM_STRING and opened port
-#.  Starts client linked to server with another ENV_VARIABLE
-#.  Booth prints the env as {}
-#.  Client sends data to server (addr from env SERVER_PORT_$PORT_TCP_ADDR)
-#.  Server prints data with prefix and resends them back with another one.
-#.  Client prints the received data and finishes.
-#.  Checks if env and all data were printed correctly.
+"spam" sub-subtest:
 
-The rm_link test
-~~~~~~~~~~~~~~~~~
-#.  Same as port (above), however after container is started, the
-    rm --link command is issued to remove the connecting link.
-#.  Verify communication between containers was prevented
-#.  Also conclude with same checks as in port (above)
-
-Prerequisites
----------------
-
-Docker daemon and host-networking setup to allow ICC between containers.
-If iptables is involved, and bridge-networking is in use, verify these
-sysctl's are set to ``1``:
-
-    * net.bridge.bridge-nf-call-arptables
-    * net.bridge.bridge-nf-call-ip6tables
-    * net.bridge.bridge-nf-call-iptables
+#. Generate hundreds of random short-named env. vars with short values
+   (so as to not exceede max command-line length)
+#. Execute docker run command with -e options for list, volume mounting
+   the sub-subtest's results dir into the container.
+#. Dump out container env into a file in mounted results dir
+#. Verify healthy container exit, no oopses or non-zero exit codes
+#. Verify generated environment vars match number & content in dump file.
+#. Ignore any container environment variables that don't contain a special
+   identifying prefix/suffix.
 """
-import ast
+
+import os
 import os.path
-import random
-import re
-
-from autotest.client import utils
-from dockertest import config
-from dockertest import docker_daemon
-from dockertest.containers import DockerContainers
+import shutil
+from dockertest.subtest import SubSubtestCaller
+from dockertest.subtest import SubSubtest
 from dockertest.dockercmd import DockerCmd
-from dockertest.dockercmd import AsyncDockerCmd
+from dockertest.images import DockerImages
 from dockertest.output import mustpass
-from dockertest.config import get_as_list
-from dockertest.images import DockerImage
-from dockertest.subtest import SubSubtestCaller, SubSubtest
-from dockertest.output import wait_for_output
-
-
-class InteractiveAsyncDockerCmd(AsyncDockerCmd):
-
-    """
-    Execute docker command as asynchronous background process on ``execute()``
-    with PIPE as stdin and allows use of stdin(data) to interact with process.
-    """
-
-    stdin_by_lines = True
-    guarantee_newlines = True
-
-    def __init__(self, subtest, subcmd, subargs=None, timeout=None,
-                 verbose=True):
-        super(InteractiveAsyncDockerCmd, self).__init__(subtest, subcmd,
-                                                        subargs, timeout,
-                                                        verbose)
-        self._stdin = None
-
-    def execute(self, stdin=None):
-        """
-        Start execution of asynchronous docker command
-        """
-        ps_stdin, self._stdin = os.pipe()
-        ret = super(InteractiveAsyncDockerCmd, self).execute(ps_stdin)
-        os.close(ps_stdin)
-        self.stdin = stdin
-        return ret
-
-    @property
-    def stdin(self):
-        """
-        Return open file instance connected to processes stdin
-        """
-        return self._stdin
-
-    @stdin.setter
-    def stdin(self, data):
-        """
-        Sends data to stdin (partial send is possible!)
-        :param data: Data to be send
-        :return: Number of written data
-        """
-        if data:
-            if self.stdin_by_lines:
-                for line in str(data).splitlines(True):  # Preserve '\n'
-                    if self.guarantee_newlines:
-                        # Make sure every (or only) line ends with '\n'
-                        if line[-1] != '\n':
-                            line += '\n'
-                    os.write(self._stdin, line)
-            else:
-                os.write(self._stdin, data)
-
-    def close(self):
-        """
-        Close the pipes (when opened)
-        """
-        if self._stdin:
-            os.close(self._stdin)
-            self._stdin = None
 
 
 class run_env(SubSubtestCaller):
-
-    """ Subtest caller """
-
-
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago mind taking a quick peek, double-check if I've done anything foolish?

cevich:
FWIW, my thinking w/ this one is to add a sub-subtest later for checking docker's ``--env-file=`` option.  That's why I made 'spam' the way it is :smiley:

edsantiago:
Happy to take a look; will continue tomorrow morning.

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
 This test checks function of `docker run -e VARIABLE=VALUE ...`.
-Also verifies the rm --link operation with ICC.
 
 Operational Summary
 ----------------------
 
-#. Setup client & server containers w/ exposed port and env. vars.
-#. Run code in containers to verify communication and env. vars
+#. Run container with -e options
 #. Check results match expectations
```
edsantiago: When writing comments I like to ask myself: who is my target audience? What's their background, what do they know, what do they need to know? For a comment at this level I might expect the maintenance programmer to be familiar with the overall test structure but perhaps not this particular module. Also, "results match expectations" doesn't spell out what the expectations are.

So, given all that, I might write something more like:

    #. Run container with a configurable number of "-e <envariable>=<value>" options
    #. From within container, dump environment
    #. Confirm that each -e pair is present in container environment

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
+
+
+class spam(SubSubtest):
+
+    # Allows quickly rejecting container env. supplied vars
+    KEYPFX = '_x'
+    VALPFX = 'x_'
+
+    @classmethod
+    def make_test_dict(cls, howmany, keyiter=None, valiter=None):
```
edsantiago: This is confusing: we're never invoked with keyiter/valiter, so that's just dead code. And if we were ever called that way, the signature would be confusing: howmany would be ignored, making for a hard-to-understand call. Have I missed something? If I haven't, might it make sense to remove the keyiter/valiter code?

While I'm at it: what would you think of renaming to just test_dict() ? "make", to me, implies an action with a side effect. "make" is the sort of thing I expect to be invoked as an imperative:

    self.make_test_dict(...)

Without make, the invocation in line 112 becomes the slightly-more-readable (to me):

    self.sub_stuff['envd'] = self.test_dict(...)

...but that's just a minor nit.

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
+            keyiter = (keyfmt % x for x in xrange(0, howmany))
+            valiter = (valfmt % x for x in xrange(howmany, 0, -1))
+        return dict(zip(keyiter, valiter))
+
+    def save_testdata(self):
+        self.sub_stuff['output'] = os.path.join(self.tmpdir, 'env.output')
+        resultdir = self.parent_subtest.resultsdir
+        inputcopy = os.path.join(resultdir, "env.input")
+        with open(inputcopy, 'wb') as inpf:
+            for key, value in self.sub_stuff['envd'].iteritems():
```
edsantiago: One of my habits is to make heavy use of sorting. It makes no difference whatsoever to the tool, but when things go wrong the human debugger will appreciate being able to 'diff env.input env.output', or just scan through them, or perhaps even compare against previous runs. Minor nit, and unlikely ever to be needed, but unlikely != impossible.

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
+                inpf.write("%s=%s\n" % (key, value))
+
+    def initialize_cmdline(self):
+        subargs = ['--rm', '-v', '%s:/x:Z' % self.tmpdir]
+        for key, value in self.sub_stuff['envd'].iteritems():
+            subargs.append("-e")
+            subargs.append("%s=%s" % (key, value))
+        subargs.append(DockerImages(self).default_image)
+        subargs.append("bash -c 'env > /x/env.output'")
```
edsantiago: See comment on line 61 re: sorting

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
-        self.failif(exp in client_res.stdout, "Data '%s' sent from server "
-                    "were not received on client:\n%s" % (exp, err_str))
-
-    def cleanup(self):
-        self.record_iptables('final')
-        super(rm_link, self).cleanup()
-        if self.sub_stuff.get('dd') is not None:
-            self.logdebug('Recovering docker daemon to original state')
-            docker_daemon.restart_service(self.sub_stuff['dd'])
+        super(spam, self).run_once()
```
edsantiago: This is very likely my own unfamiliarity with autotest, but is this necessary here?

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
+                    self.logdebug("Ignoring %s", line)
+                    # These are not the droids we're looking for
+                    continue
+
+                output[key] = value
+        return output
+
+    def validate_testdata(self, expected, actual):
+        # More helpful to debugging if done this way
+        for key, value in actual.iteritems():
```
edsantiago: I really like this level of consideration for the person running the test. Nice work. But is 'actual' the best loop to run? This would catch envariables in the docker container that match the KEYPFX/VALPFX requirement but were not assigned via -e. Is that likely? My gut feel is that the error we're trying to catch is "command line specified -e FOO=BAR but FOO didn't get passed to the container". That would suggest looping over expected, not actual. But of course perhaps I've missed something.

subtests/docker_cli/run_env/run_env.py
```diff
@@ -3,493 +3,123 @@
...
+        output = {}
+        with open(self.sub_stuff['output'], "r") as outf:
+            for line in outf:
+                key, value = line.split('=', 2)
+                key = key.strip()
+                value = value.strip()
+                if (not key.startswith(self.KEYPFX) or
+                        not value.startswith(self.VALPFX)):
+                    self.logdebug("Ignoring %s", line)
+                    # These are not the droids we're looking for
```
edsantiago: Love the comment, but in reviewing the debug log I find the "Ignoring" lines distracting; more so because of the unstripped newline. How about just removing this logdebug, and adding a simple count before returning?

    if key.startswith() and value.startswith:
        output[key] = value
    self.logdebug("Found %d envariables matching %s*=%s... in %s" % (len(output),
           self.KEYPFX, self.VALPFX, outputcopy)

subtests/docker_cli/run_env/run_env.py
cevich: Oh right, you are correct, and I might add as FYI, these subtest-module docstrings are what shows up in documentation for the test.  So the summary does tend to be very generic and high-level (for managers to read).  The detail you're suggesting I'd actually put into another (otherwise optional) section 'Operational Detail', that's intended to describe the individual sub-subtests more thoroughly.  I'll add in that section for what I've done here...

subtests/docker_cli/run_env/run_env.py
cevich: I was intending to subclass  this ``spam`` class for sub-subtests where the usage would make more sense.  I simply didn't implement the iterators (which could have side-effects).  I suppose it's just wasted code if/until there actually are more sub-subtests, so np killing it for now.

subtests/docker_cli/run_env/run_env.py
cevich: GMTA!  I actually spent some time thinking about this, and even considered doing something with the ``difflib`` module.  Then I considered the unlikeness, and decided the next-best thing is to make sure both the input and output mappings were available in the results directory.  That's automaticly included in all the reporting, so accessible if ever needed.  For anything more complex though, I agree with you, having the setup/input be explicitly deterministic in content and order makes problem finding super easy.  Even when things need/should be "random", I insist that the random seed be visible and configurable to aid in problem reproduction.

subtests/docker_cli/run_env/run_env.py
cevich: Yeah, it's part of the API to do this.  There are some, well at least one call for ``run_once()`` in the super classes which log a message.  The other standard methods take other necessary actions (depending).  For example, I know ``initialize()`` computes the subtest "version" and sets up some private variables for tracking sub-subtest state and such.

subtests/docker_cli/run_env/run_env.py
cevich: Nope, you're correct.  I remember struggling with this while writing it.  It is checking that keys didn't get passed, and also that keys and values didn't get mangled in some way by docker on input or on dump into results file.  Perhaps the thing to do here is check it both ways, then there can be no doubt.

subtests/docker_cli/run_env/run_env.py
edsantiago: I think what I mean is: does run_once() belong in postprocess()? That makes two invocations (one in self.run_once) which just doesn't seem right.

subtests/docker_cli/run_env/run_env.py
edsantiago: The failif() at the end leaves no doubt (that's the actual test). The key/value loop here is just for the benefit of humans. Can you imagine a scenario where Docker creates spurious envariables of the form _x123=x_321? And even if it does, would we care? IMO what we care about is that a -e option gets passed into the container, no?

subtests/docker_cli/run_env/run_env.py
cevich: With all the different images and environments we test in, it's handy to have an idea of what else is in the environment (especially the horrible-use, but critically important 'container=whatever' variable).  I don't expect anyone to use the debug log unless something's gone wrong, and couldn't be bothered to write out the extraneous stuff to a file, so into the debug log it goes :smile: 

I've updated it though to just dump the dictionary of extras instead of each line.

cevich:
@edsantiago thanks for your input, I've implemented most if not all your suggestion, very helpful.  Travis is happy as well, so going to merge this and open up issues for creating tests to exercise the parts I nuked in re-writing this.

cevich:
Oops, you're absolutely right about the ``run_once()`` not belonging in ``postprocess()``.  I'll fix that.

Re: Can you imagine a scenario where Docker creates spurious envariables of the form _x123=x_321? And even if it does, would we care?

Yes, if docker is modifying variables I'm passing in to values I didn't set, I certainly do care.  To your point though, it's probably overly pedantic.  But I'm too lazy to remove it now :smile:

edsantiago:
> if docker is modifying variables I'm passing in to values I didn't set, I certainly do care.

That would be _modifying_, not setting. And it would be caught by the comparison. I'm still not convinced the loop_actual=True case is useful, and I actually think it decreases maintainability, but I'll let it be for now.

cevich:
@edsantiago oops, didn't mean to drop that failif, okay stuck that back in and suck it into #563

PR #561, opened 2016-02-15
info subtest: fix verify_pool_name()

edsantiago:
verify_pool_name() checks that the Docker Pool Name listed
by 'docker info' is present in the output of 'dmsetup ls'.
We cannot rely on the output order of the latter: on one
RHEL7 system it outputs the desired pool name first, on
another system last. So let's just test for the presence
of the desired pool name anywhere in the output.

Also, create a unit test for the new code. Invocation is
ugly but that should be fixable by someone more python-
proficient than I. Fixes welcome.

Unit test requires python-nose and python-unittest2, both
of which are available on RHEL7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
     - pip install Pylint==1.4.5 Pep8==1.6.2 Sphinx==1.2.2
     - pip install Autotest==0.16.2
     - pip install inspektor==0.2.0
+    - pip install unittest2==0.8.0
 
 script:
     - SPHINXOPTS="-W" make
--- a/run_unittests.sh
+++ b/run_unittests.sh
@@ -14,5 +14,13 @@ do
     wait %- &> /dev/null
     RET=$?
 done
+
+# Unit tests for subtests.
+# FIXME: find a way to have nosetests recurse, or move the tests into
+#        a directory structure that nose can handle.
+for unittest in $(find subtests -name 'test_*.py'); do
+    nosetests -v $unittest
+done
+
 echo ""
 echo ""
--- a/subtests/docker_cli/info/info.py
+++ b/subtests/docker_cli/info/info.py
@@ -88,23 +88,27 @@ class info(subtest.Subtest):
             metadata_name = 'Metadata file'
             # TODO: Checks based oninfo_map['Backing Filesystem:']
 
-    def verify_pool_name(self, docker_pool_name):
-        read_pool_names = utils.run("dmsetup ls | grep 'docker.*pool'")
+    def verify_pool_name(self, expected_pool_name):
+        """
+        Pool Name reported by docker info should be listed in
+        results from dmsetup ls.
+        """
+        read_pool_names = utils.run("dmsetup ls")
         raw_pools = read_pool_names.stdout.strip()
-        pool_names = [x.split()[0] for x in raw_pools.split('\n')]
+        pool_names = [x.split()[0] for x in raw_pools.split('\n')
+                      if 'pool' in x]
 
         # make sure there is at least one pool
         self.failif(len(pool_names) < 1,
-                    "There are fewer than one docker pools.")
-        self.logdebug("Docker pool found.")
+                    "'dmsetup ls' reports no docker pools.")
+        self.logdebug("Docker pool(s) found: %s" % pool_names)
 
-        read_pool_name = pool_names[-1]
         # verify pool names
-        self.logdebug("Read Pool Name: %s , Docker Pool Name: %s",
-                      read_pool_name, docker_pool_name)
-        self.failif(docker_pool_name != read_pool_name,
-                    "Docker pool name does not mach dmsetup pool.")
-        self.logdebug("Docker pool name matches dmsetup pool.")
+        self.failif(expected_pool_name not in pool_names,
+                    "Docker info pool name '%s' (from docker info)"
+                    " not found in dmsetup ls list '%s'" %
+                    (expected_pool_name, pool_names))
+        self.logdebug("Docker pool name found in dmsetup ls.")
 
     @staticmethod
     def size_bytes(num_str):
--- a/subtests/docker_cli/info/test_info.py
+++ b/subtests/docker_cli/info/test_info.py
@@ -1,0 +1,94 @@
+# -*- python -*-
+#
+# Tests for the docker-autotest 'info' subtest
+#
+# As of 2016-02-12 this only includes verify_pool_name(); we'll try to
+# extend coverage as needed.
+#
+# RUNNING:
+#
+#   $ nosetests -v subtests/docker_cli/info
+#
+# This assumes that your autotest-docker is checked out underneath
+# the client/tests subdirectory of a checked-out autotest repo,
+# *and* that the autotest repo is called "autotest" (e.g. not "my-autotest").
+# Also: you can't just run nosetests and hope it'll recurse. It won't.
+#
+# Yeah. Sorry. There's indubitably a better way. Please fix if you know how.
+#
+from unittest2 import TestCase, main
+from mock import Mock, patch
+
+import sys
+sys.path.append('../../../..')
+
+import info
+
+class TestVerifyPoolName(TestCase):
+    # Standard docker pool name to expect
+    docker_pool = 'rhel-docker--pool'
+
+    # Typical output from 'dmsetup ls'. Note that order is arbitrary.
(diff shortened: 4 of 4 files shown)


cevich:
Originally, the run_unittests.py was only ever intended to check the API stuff.  It's fine to extend that, however, you'll need to tweak ``.travis.ymp`` so it installs the modules your unittests depend on (it's an ubuntu environment).

It's also going to run into a problem with ``run_pylint.sh`` thinking that your unittest is part of a subtest module.  You can get around that by adding a filter that excludes 'test_*.py' from this script.

Finally, you may need to actually go in and add similar filtering into ``dockertest/subtest.py`` in the ``SubSubtestCaller`` class, ``import_if_not_loaded`` module.  It loads sub-subtests assuming the directory is a python module (so it's going to process ``__init__.py`` and anything else in that directory).

For safety's sake, you may want to look through the rest of that class and it's subclasses, because having ``Mock`` get loose accidentally inside a test run is very very very very bad.  I've actually had this bite me before, and it was extremely difficult to debug (because Mock was hiding it).  This is why we originaly wrote and hard-coded a mock-like helper into each of the ``dockertest`` module's unittests.

One idea might be, split this PR in two.  One for updating the test and a second for dealing with all the unittest updates/additions.  That way we can get the fixed test running in production earlier :smile:

edsantiago:
Wow! Thanks for that feedback. I have a lot to look into; it might take me a few days because I'm working on other projects right now. Thanks again, I really appreciate your being aware of these issues.

cevich:
Ya, she's a complicated beast.  They ultimately descend from Autotest's ``test.test`` class, which brings some non-obvious requirements.  For example, how we store runtime data in ``stuff`` and ``sub_stuff``, that's because the test class must be pickle-able to support mid-test rebooting (which still only works in theory).  It's best to think of the subtests more like "plugins" rather than python modules.  Anyway, I'm not saying unit-tests for the tests are bad, they just come with some heavy baggage.

If you split out the test changes from the unittest stuff, I'll just merge in the former and then you can continue the later as you have time.

edsantiago:
Thanks for your patience. I've spent today looking into this in depth. I hope to push a new commit soon but:

> It's also going to run into a problem with run_pylint.sh thinking that your unittest is part of a subtest module. You can get around that by adding a filter that excludes 'test_*.py' from this script.

pylint was helpful, and found some issues which I've fixed. Why would I want to disable it?

> Finally, you may need to actually go in and add similar filtering into dockertest/subtest.py in the SubSubtestCaller class, import_if_not_loaded module. It loads sub-subtests assuming the directory is a python module (so it's going to process __init__.py and anything else in that directory).

None of the subtests are modules. There's no `__init__.py` anywhere nor any code (AFAICT) that does wildcard imports. The only way I can see test_<anything> being imported is if some .ini file explicitly lists it in subsubtests. Unless there's something I've missed, it seems safe to create any random foo.py in a subtest directory. (Obvious question is now: what have I missed?)

> For safety's sake, you may want to look through the rest of that class and it's subclasses, because having Mock get loose accidentally inside a test run is very very very very bad.

In reading through f9f258d it looks like your use of mock is very different from mine. In particular, all my uses of mock are within test functions, never at module level. If there's some way for this to "escape" it's beyond me and not mentioned in any of the references I've perused.

edsantiago:
Ugh! Looks like I screwed up by rebasing onto master. Should I close and create a new pr?

cevich:
> The only way I can see test_ being imported is if some .ini file explicitly lists it in subsubtests. Unless there's something I've missed, it seems safe to create any random foo.py in a subtest directory. (Obvious question is now: what have I missed?)

Hmmm, no, actually you may be correct.  Let me take another quick peek...

> In reading through f9f258d it looks like your use of mock 

Oh right! Now I remember what happened.  It wasn't a test module, it was ``conf.py`` (part of sphinx doc. building).  There use to be a check inside each subtest to make sure the docs were up to date.  But on reading ``conf.py``, sys.modules got all snafu'd (it was conf.py's fault).  Wow, you really went digging, good for you!

cevich:
Ahh okay, and travis is happy now also, cool.

It's possible for a subtest to have sub-subtests w/o them being explicitly listed in configuration.  This allows tests to find/load them dynamicly.  To support that pattern along with unittests, maybe we just need a check to make sure there are no dunder-init.py files used in subsubtest paths.

I needa do some reading on ``mod = imp.load_module(name, *imp.find_module(name, pkg_path))``.  I'm mostly worried hidden/implicit load problems and the dynamic-case above.  You've squashed my other concerns, so we're probably good, I just wanna be sure...

cevich:
...okay, refreshed.  Been a while since I looked in the bowels of the imp, I think we're good.  It seems ``./run_pylint`` may be lying to us:

```
************* Module info
W0511: 89,0: : TODO: Checks based oninfo_map['Backing Filesystem:']
Checking: subtests/docker_cli/info/test_info.py 
subtests/docker_cli/info/test_info.py:25:1: E402 module level import not at top of file
```

That E402 should have raised a flag, it's probably something stupid I did in the silly script...

edsantiago:
> subtests/docker_cli/info/test_info.py:25:1: E402 module level import not at top of file

Interesting. I didn't see that, and don't see it in my runs on F22 or RHEL7. I wonder if it's pylint-version-dependent? I could imagine fixing it by setting PYTHONPATH in the run_unittests loop...

cevich:
While pylint/pep8 is out of the question, I was thinking about running the unittests inside RHEL/CentOS/Fedora Atomic as part of a CI job for Docker Autotest.  in fact, I'm frustrated that I added a dependency on docutils.  Otherwise the unittests run fine on Atomic.   But  those features don't exist yet, and this one does.  Is it a terrible pain to drop the dependencies on nosetests, unittest2 and mock?  If so, I can overlook that and leave it to "the future" to work out.

cevich:
> Interesting. I didn't see that, and don't see it in my runs on F22 or RHEL7. I wonder if it's pylint-version-dependent?

pylint has always been a bit of a PITA.  and run_pylint is pretty fuggly as well.  It doesn't help one single bit that travis runs Ubuntu either :sa: 

If I had to guess, it's the secondary 'show TODO/FIXUP' pylint runs that cause it not to return an exit code.

cevich:
Ohh, wait, in fact...there is "another" mock you could use...lemmie see if it's still there...

cevich:
...ya! here it is ``autotest/client/shared/test_utils``, maybe a little different but should already be setup inside the autotest namespace.

edsantiago:
Ugh. I really need to think about this, but my gut reaction is that I'm reluctant to drop unittest2 and mock and nosetests. They seem to be the de facto standards for python testing. What would be the purpose of running unit tests on Atomic?

cevich:
Because all the Atomic's are "special" :smile:  The standard everywhere in Autotest and Docker Autotest is to use the built-in stuff rather than add anything fashionable.  We like to keep the framework stuff as plain-jane as possible because the test subjects are often complex enough.  Anyway, they're just developer tools so I'm willing to over look non-existing features and philosophy if it's a major pain.

cevich:
Okay, ya, the E402 must be an ubuntu/pep pylint thing.  In Atomic and RHEL we're only on pylint 1.3.1.  Though really, if I can get this docker-autotest-ci thing running properly, we could entirely drop travis.

cevich:
heheh, ya RHEL pylint-1.3.1 doesn't care, I stuck 'import info' at the bottom of the file and it doesn't complain.  No clue how/why ubuntu's pylint from pip doesn't exit with an error, anything I do on RHEL to make pylint throw an ``E`` code, and the script returns a '1' :confounded:

cevich:
I was able to rebase your branch against ``upstream/master`` but had to fiddle with ``into_tests.py``.  I noticed you've got a few commits w/o any signatures also.  So just fix up those few things and I can call it good-nuff.

Bonus points if you can get it working w/o the extra dependencies, but as I said, if it's a major PITA, then forget it.  Probably just as good to pave the way for all the other unittests to get updated sometime anyhow.

edsantiago:
I can't find a way to make tests run on Atomic: no pylint, no docutils. May I back out 4fa23fc?

cevich:
> I can't find a way to make tests run on Atomic: no pylint, no docutils. May I back out 4fa23fc?

oh yes, please, go ahead.  It was a good try, qualifies for 'PITA' in my book.  Though with these new dependencies, it means docs need updating as well.

cevich:
Thanks for all your help @edsantiago I just left the one little broken E402, we can catch/fix that some other time.

PR #563, opened 2016-02-15
Tt flex

cevich:
Enhance TextTable class to give callers more flexibility in use.  Updated version test to write out the client/server version into sysinfo directory.  Finally, two minor fixups for the run_env test.

### DIFF
--- a/subtests/docker_cli/run_env/run_env.py
+++ b/subtests/docker_cli/run_env/run_env.py
@@ -103,15 +103,11 @@ class spam(SubSubtest):
         self.logdebug("Environment variables NOT included in test: %s", extra)
         return output
 
-    def validate_testdata(self, expected, actual, loop_actual=True):
-        if loop_actual:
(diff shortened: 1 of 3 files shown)


subtests/docker_cli/run_env/run_env.py
```diff
@@ -120,6 +116,8 @@ def validate_testdata(self, expected, actual, loop_actual=True):
             if value != compareto[key]:
                 self.logwarning("var %s value %s does not match %s's %s",
                                 value, key, name, compareto[key])
+        # Now fail after checking all the keys/values
+        self.failif(expected != actual)
```
cevich: @edsantiago like this, ya?

edsantiago: Yes, and likewise with the postprocess(). Thanks!

cevich: No, thank you!

PR #565, opened 2016-02-22
Fix garbage_check not preserving image IDs

cevich:
* Additional regex magic & processing is needed when a FQIN is specified
without the repo-address component.
* Images module now uses key-words to reference components of FQIN
group matches
* Added new unittests to verify above changes
* Updated garbage_check to be more flexible when comparing against
configured images to preserve.

### DIFF
--- a/dockertest/images.py
+++ b/dockertest/images.py
@@ -54,8 +54,11 @@ class DockerImage(object):  # pylint: disable=R0902
     #: parsing, spec defined in docker-io documentation.  e.g.
     #: ``[registry_hostname[:port]/][user_name/]
     #: (repository_name[:version_tag])``
-    repo_split_p = re.compile(r"(.+?(:[\d]+?)?/)?([\w\-\.\+]+/)?"
-                              r"([\w\-\.]+)(:[\w\-\.]+)?")
+    repo_split_p = re.compile(r"(?P<repo_addr>.+?"  # optional
+                              r"(?P<repo_addr_port>:[\d]+?)?/)?"  # optional
+                              r"(?P<user>[\w\-\.\+]+/)?"  # optional
+                              r"(?P<repo>[\w\-\.]+)"  # required
+                              r"(?P<tag>:[\w\-\.]+)?")  # optional
 
     # Many arguments are simply required here
     # pylint: disable=R0913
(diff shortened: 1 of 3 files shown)


dockertest/images.py
```diff
@@ -54,8 +54,11 @@ class DockerImage(object):  # pylint: disable=R0902
     #: parsing, spec defined in docker-io documentation.  e.g.
     #: ``[registry_hostname[:port]/][user_name/]
     #: (repository_name[:version_tag])``
-    repo_split_p = re.compile(r"(.+?(:[\d]+?)?/)?([\w\-\.\+]+/)?"
-                              r"([\w\-\.]+)(:[\w\-\.]+)?")
+    repo_split_p = re.compile(r"(?P<repo_addr>.+?"  # optional
```
edsantiago: If you add an extra layer of parens:

     r"((?P<repo_addr> [...] )?)/)?"  # new paren after r" and immediately before the slash


...you can remove the awkward trim-slash at line 145-146.

dockertest/images.py
```diff
@@ -136,15 +139,26 @@ def split_to_component(cls, full_name):
...
-            if user:
-                user = user[:-1]  # remove trailing slash
-            if tag:
-                tag = tag[1:]  # remove beginning colon
-        except TypeError:  # no match
+            full_name = full_name.strip()
+            mobj = cls.repo_split_p.match(full_name)
+            repo_addr = mobj.group('repo_addr')
+            if repo_addr is not None:
+                repo_addr = repo_addr.replace('/', '')
```
edsantiago: See above.

dockertest/images.py
```diff
@@ -54,8 +54,11 @@ class DockerImage(object):  # pylint: disable=R0902
     #: parsing, spec defined in docker-io documentation.  e.g.
     #: ``[registry_hostname[:port]/][user_name/]
     #: (repository_name[:version_tag])``
-    repo_split_p = re.compile(r"(.+?(:[\d]+?)?/)?([\w\-\.\+]+/)?"
-                              r"([\w\-\.]+)(:[\w\-\.]+)?")
+    repo_split_p = re.compile(r"(?P<repo_addr>.+?"  # optional
+                              r"(?P<repo_addr_port>:[\d]+?)?/)?"  # optional
+                              r"(?P<user>[\w\-\.\+]+/)?"  # optional
```
edsantiago: Likewise: ((?P<user> ... )/)?

dockertest/images.py
```diff
@@ -136,15 +139,26 @@ def split_to_component(cls, full_name):
...
-                tag = tag[1:]  # remove beginning colon
-        except TypeError:  # no match
+            full_name = full_name.strip()
+            mobj = cls.repo_split_p.match(full_name)
+            repo_addr = mobj.group('repo_addr')
+            if repo_addr is not None:
+                repo_addr = repo_addr.replace('/', '')
+            user = mobj.group('user')
+            if user is not None:
+                user = user.replace('/', '')
```
edsantiago: Likewise.

cevich:
@edsantiago Good idea!  Bot I miss someone(s) looking over my shoulder.    Unf. I'm in a bit of a rush with this change (for CI's sake) and I *think* what's here now is 'working'.   It's a case of ugly-but-working > pretty-but-need-to-re-test.  In fact, I've squeezed in testing along with https://github.com/cevich/autotest-docker-enabled-product-testing/pull/26 to try and save some time...uggg...so rushed!  Though I'll open an issue on your suggestion, because it's a good one.

PR #568, opened 2016-02-23
Pylint cleanup

edsantiago:


cevich:
@edsantiago instead of ``Merge remote-tracking branch 'upstream/master'`` can you just rebase your branch from upstream/master instead?  That way the only merge commits are coming from actual PR merges.  Looking at the code here quickly, looks like a fine update, not much to comment about, except maybe one thing I'm testing now...

run_pylint.sh
```diff
@@ -140,72 +129,42 @@ check_subtest() {
...
-            record_return 1
-        fi
+        run_ff $PEP8 --ignore=$PEP8IGNORE "$WHAT"
     fi
 }
 
 check_subtests() {
     for thing in pretests subtests intratests posttests
     do
-        trap "break" INT
```
cevich: I'd like to keep this here because it allows ctrl-c'ing out of the loop to abort the run.  Without this, it's most likely to kill pylint instead of the for loop (b/c more time spent in pylint), which means the remaining pylint runs just keep going.   Only other way out is to spam ctrl-c and get lucky with the timing.

cevich:
Yep, rest of it looks good.  Just that trap change and rebasing is needed, then I'm happy to merge this.

edsantiago:
INT: my bad. Fixed.

Rebase: git co master && git rebase upstream/master && git co pylint_cleanup && git rebase master -- all done, but git push claimed 'already up to date'.

edsantiago:
Abandoned original branch, created new one, cherry-picked. Now it's just 1 commit.

cevich:
Excellent, thanks!  Maybe we won't have to look at this damn script for another two months :smile:

PR #569, opened 2016-02-24
fix rmi subtest failure

edsantiago:
rmi with_blocking_container_by_tag test was failing:

   repository name component must match "[a-z0-9]+(?:[._-][a-z0-9]+)*"

Root cause: DockerImage.get_unique_name("FOO") generates a name
of the form "<testname>_FOO_<rand>". If FOO is ":tag" (colon-tag),
as it was in this test, we get "<testname>_:tag_<rand>" and
docker doesn't like the trailing underscore in "<testname>_".

Fix: use "suffix:tag" instead of ":tag" for unique_name prefix.

Also: removed confusing test descriptions, making it clearer (I
hope) what this subtest does.

Also: fixed what I believe was use of incorrect variable name
in error message.

### DIFF
--- a/subtests/docker_cli/rmi/rmi.py
+++ b/subtests/docker_cli/rmi/rmi.py
@@ -147,10 +147,7 @@ class with_blocking_container_by_tag(rmi_base):
 
     1. Create new image with full_name (tag) from image (base_image)
     2. Use new image by new container (docker run image...)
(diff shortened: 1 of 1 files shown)


cevich:
Fantastic Ed, thanks!  Let's see...

cevich:
Great! Nice simple fix and good debugging.

PR #571, opened 2016-02-26
Fix test failure in double_tag subsubtest.

edsantiago:
subtest tag:double_tag was failing. Root cause: too many
to list. I don't think this test could have ever completed
successfully.

Fixes:
 + Principally, reworked the way we check exit status
   when we're expecting the command to fail (i.e.
   two 'docker tag' commands in a row without -f).

Related fixes:
 + moved some logic from the config file to the code:
   1) docker_expected_result (PASS/FAIL): this absolutely
      does not belong in a config file. There is no
      circumstance under which an end user would want
      to say "oh hey let's expect double-tag to pass"
      or "let's expect other-test to fail". This is
      code logic, not configuration logic.
   2) -f flag (in double_tag_force): again, this is
      code logic. It's the whole defining factor of
      the test itself. It is harmful to imply that
      an end user should be able to change this via
      config file.

Unrelated fixes:
 + differentiation of two previously identical debug
   messages, so human debugger can identify which part
   of the code a message is coming from;
 + variable name change, fixing a misleading reuse situation
 + make first docker command in double_tag verbose, making
   test flow clearer in debug log
 + slightly more informative comments
 + minor typo fix

### DIFF
--- a/config_defaults/subtests/docker_cli/tag.ini
+++ b/config_defaults/subtests/docker_cli/tag.ini
@@ -4,8 +4,6 @@ tag_force = false
 #: generate names using lowercase only
 gen_lower_only = True
 subsubtests = change_tag,change_repository,change_user,double_tag,double_tag_force
-#: expected results
-docker_expected_result = PASS
 #: testing repo_name_prefix
 tag_repo_name_prefix = test
 #: Works best with remote registry provided repository
@@ -24,7 +22,5 @@ docker_registry_host =
 [docker_cli/tag/change_user]
 
 [docker_cli/tag/double_tag]
-docker_expected_result = FAIL
 
 [docker_cli/tag/double_tag_force]
-tag_force = true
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago life would have been much easier if you were the author of most of these tests :smirk_cat: 

TBH, you'll probably see a bunch more of those 'expected to fail" configuration options.  Remember many of these tests were written before even docker 1.0.  We literally had no idea whether or not some feature was expected to work or not,  and plenty of examples of really odd behaviour that turned out as expected.  Also, it was normal to have a good 1-3 month window, for any replies to bugzillas we opened.  So flipping a config. value was often seen as the lesser of many evils.  Anyway, just some history.

Don't forget about the "Ripley Option" as well, I know from hind-sight there are many tests _waaaay_ over-engineered.

edsantiago:
[shamefaced] Thanks for the history lesson; that makes sense for rapid development in uncertain times.

cevich:
It was actually much worse, when you consider some of the bugs that went upstream have taken over a year to get merged.  For example, recently (for docker 1.9.x) daemon and client will have their own unique exit codes, so we can distinguish a rc=1 from the contained command vs failure in the client.  In fact, that's going to break a lot of the items in the 'negativeusage' subtest (and probably others).

Issue #572, opened 2016-02-29
Quick Fix: Default omit_empty=True in get_as_list()


cevich:
Also needs a quick double-check through existing tests to make sure this change won't cause any obvious breakage (I don't think it will).  Lining this up with the next version since it's an API change.

cevich:
oh fantastic, thanks!

PR #573, opened 2016-03-03
run_volumes: Add ability to execute optional pre-testing command

cevich:

### DIFF
--- a/config_defaults/subtests/docker_cli/run_volumes.ini
+++ b/config_defaults/subtests/docker_cli/run_volumes.ini
@@ -2,11 +2,15 @@
 subsubtests = volumes_rw, volumes_one_source
 #: These must be setup prior to testing
 __example__ = host_paths, cntr_paths
+#: Command-line string, required to succeed prior to testing (if any).
+#: Will only be run once, before any sub-subtests execute.
+#: e.g. setsebool virt_sandbox_use_nfs on
+pretest_cmd =
 #: CSV list of host paths to check, assumes filesystem is already mounted
 #: for testing NFS, SMB, etc.  **If using SELinux, don't forget to
 #: 'setsebool virt_sandbox_use_<type> on' where <type> is nfs, samba, etc.**
 host_paths =
-#: CSV list of destination paths in container cooresponding to host_paths above
+#: CSV list of destination paths in container corresponding to host_paths above
 #: **(must exactly match the number of host_paths)**
 cntr_paths =
 #: CSV options to docker command(s), values will be substituted for
--- a/subtests/docker_cli/run_volumes/run_volumes.py
+++ b/subtests/docker_cli/run_volumes/run_volumes.py
@@ -118,7 +118,7 @@ class volumes_base(SubSubtest):
     def make_dockercmd(subtest, dockercmd_class, fqin,
                        run_template, cmd_tmplate, test_dict):
         # safe_substutute ignores unknown tokens
-        subargs = get_as_list(str(run_template % test_dict))
+        subargs = get_as_list(str(run_template % test_dict), omit_empty=True)
         subargs.append(fqin)
         subargs.append(cmd_tmplate % test_dict)
         return dockercmd_class(subtest, 'run', subargs)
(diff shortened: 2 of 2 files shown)


config_defaults/subtests/docker_cli/run_volumes.ini
```diff
@@ -2,6 +2,9 @@
 subsubtests = volumes_rw, volumes_one_source
 #: These must be setup prior to testing
 __example__ = host_paths, cntr_paths
+#: Command(s) required to succeede prior to testing (if any)
```
edsantiago: Typo, no trailing 'e' in 'succeed'

subtests/docker_cli/run_volumes/run_volumes.py
```diff
@@ -42,7 +42,12 @@
 
 
 class run_volumes(SubSubtestCaller):
-    config_section = 'docker_cli/run_volumes'
+
+    def run_once(self):
+        ptc = self.config['pretest_cmd'].strip()
+        if ptc != '' and ptc is not None:
```
edsantiago: AIUI the preferred idiom for this is just:

    if ptc:

(because null string and None are non-truthy). References available on request.

subtests/docker_cli/run_volumes/run_volumes.py
```diff
@@ -42,7 +42,12 @@
...
+            utils.run(ptc, timeout=self.config['docker_timeout'])
+        super(run_volumes, self).run_once()
```
edsantiago: In every other instance I can find, super() is called before anything else. Is there a reason why ptc runs first here? If so, should that be documented?

config_defaults/subtests/docker_cli/run_volumes.ini
```diff
@@ -2,6 +2,9 @@
...
+#: e.g. setsebool virt_sandbox_use_nfs
```
edsantiago: Missing value (eg 1/true/on).

Given that the comment includes "(s)" for multiple commands, should the comment also specify "semicolon-separated"? It doesn't seem worth the effort to make this a list.

subtests/docker_cli/run_volumes/run_volumes.py
cevich: You're correct, however your reminded me there is a docker autotest "gotcha" WRT config values:  See ``dockertest/config.py:272  Though it's still wrong so I'll fix it.  Thanks.

subtests/docker_cli/run_volumes/run_volumes.py
cevich: run_once() in this case actually executes all the sub-subtests, so it has to come last.  Probably I should put the ptc thing into initialize() instead, okay I'll fix that, thx.

config_defaults/subtests/docker_cli/run_volumes.ini
cevich: I think it can be separated however you like, it's just passed through to ``utils.run()`` which IIRC sends it into a 'bash -c' as a string.  Unless I'm misunderstanding your comment.

cevich:
@edsantiago thanks for reviewing, thought this would be a quick fix :cry: I'm having lots of trouble getting it to pass in CI though, either the test breaks or the setup breaks or both.  I needa hack on it some more before it's really ready for another review.

config_defaults/subtests/docker_cli/run_volumes.ini
edsantiago: Sorry for not being clear. What I meant was, in your comment you specify "command(s)" which implies that one can specify multiple commands. But it isn't clear to the config-file editor (who might not be familiar with utils.run, and might not want to be) _how_ to specify multiple commands. Comma? ["command1", "command2"]? A hint ("use semicolon to separate multiple shell commands") could steer someone along the right path.

subtests/docker_cli/run_volumes/run_volumes.py
```diff
@@ -187,16 +200,8 @@ def initialize(self):
         super(volumes_rw, self).initialize()
         host_paths = get_as_list(self.config['host_paths'])
         cntr_paths = get_as_list(self.config['cntr_paths'])
```
edsantiago: Would it make sense to add `, omit_empty=True` above? That would give a more helpful message when config settings are empty.

cevich: Oh thanks, yep, should have that too until #572

cevich:
Okay, this is "working" in CI now.  Feel free to review, and if more changes needed, we can just pack them in on top.  @edsantiago thanks a bunch for your feedback!

PR #574, opened 2016-03-10
Update control.ini comments + BZ searching

cevich:
Matching bugs containing multiple subthing names should now be possible.


cevich:
Assuming I can get this working in CI, it will allow us to skip testing known-failing tests based on matching some values found in bugzilla searches.

PR #575, opened 2016-03-11
Monkeypatch and add no-execute control argument

cevich:
The Monkeypatch is used on sys.path to include the test directory in
module search path.  This is needed to support adding the
python-bugzilla module to the test-directory on systems where installing
elsewhere is impossible/impractical (i.e. Atomic).  The no-execute
argument is primarily intended for debugging control.ini and bugzilla
search options.

### DIFF
--- a/control
+++ b/control
@@ -80,11 +80,16 @@ def get_bzobj(bzopts):
         logging.debug("Bugzilla url empty, exclusion filter disabled")
         return None
     try:
+        # Allow bugzilla module to be embedded into test directory
+        sys.path.insert(0, os.path.dirname(job.control))
         import bugzilla  # Keep confined to this function
     except ImportError:
         logging.warning("Bugzilla status exclusion filter configured "
                         "but bugzilla python module unavailable.")
         return None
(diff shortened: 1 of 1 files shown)


control
```diff
@@ -80,11 +80,16 @@ def get_bzobj(bzopts):
...
         return None
     try:
+        # Allow bugzilla module to be embedded into test directory
+        sys.path.insert(0, os.path.dirname(job.control))
         import bugzilla  # Keep confined to this function
     except ImportError:
         logging.warning("Bugzilla status exclusion filter configured "
                         "but bugzilla python module unavailable.")
         return None
+    # We were never here, you didn't see or hear anything.
```
edsantiago: Should this be done within `finally` ?

cevich: Hmmmm, good point, that would be cleaner that the conditional.  I'm not too worried about it in the case of an exception, b/c anything breaking at this level will be cleaned up by autotest.  Okay, I'll sneak a fix in along with another PR :smile:   Thanks for checking!

PR #576, opened 2016-03-11
signal 48 (SIGRTMIN+14) is now valid

edsantiago:
Found while testing docker 1.9.1-el7: 'kill -s 48' in the kill_bad
subtest now succeeds because 48 is a valid, if unusual, signal.
Looks like change was introduced in August 2015 via
0e50d946a25beb134bce2aaf4a209b5cfcbacf8f

Submitting as a lone PR because this is 1.9-related but technically
not part of the new exit code work.

### DIFF
--- a/config_defaults/subtests/docker_cli/kill_bad.ini
+++ b/config_defaults/subtests/docker_cli/kill_bad.ini
@@ -9,4 +9,4 @@ exec_cmd = 'while :; do sleep 0.1; done'
 
 [docker_cli/kill_bad/bad]
 #: csv of bad signals which will be used in test
-bad_signals = 0,-1,-78,48,96,SIGBADSIGNAL,SIG,%%,!,\\,\,"''",'""', ,
+bad_signals = 0,-1,-78,96,SIGBADSIGNAL,SIG,%%,!,\\,\,"''",'""', ,


cevich:
Hmmmm, seems like we could use version-sensitive configuration options hmmm :confused:    Another idea could be to just remove this as an option, and stick it into the test, where the docker version can be examined.  What do you think?  The easiest thing to do is probably to just wait and merge this later on, when 1.9.x is common place.

edsantiago:
I'm not sure what your concern is. Is it that we'd lose the ability to check that docker-1.8 (incorrectly) marks sig48 as bad? ISTM that testing with 48 just-so-happened to work with docker-1.8, but that was a bug in docker (and hence in the test as well). What am I missing? Why would we want to keep 48 in the list of "bad" signals?

cevich:
Oh, my mistake then.  It works in 1.8.x with this change?  I thought you meant it only worked in 1.9+

edsantiago:
Sorry! Incomplete communication on my part. I had hoped that github would autolink the SHA to the appropriate docker commit, but obviously that only works within a repo. Let's try [this](https://github.com/calavera/docker/commit/0e50d946a25beb134bce2aaf4a209b5cfcbacf8f). Note the comment that (incorrectly) reads "The largest legal signal is 31". I don't actually know the full history of SIGRTMIN .. SIGRTMAX, but they are defined by POSIX and it seems only fair that docker should propagate them if so invoked. Basically, docker-1.9 removed this restriction.

cevich:
Ahhh gotcha, okay, so I can merge this and the test will still be okay on docker 1.8 and it'll be even better on 1.9 (just confirming).

edsantiago:
Well..... technically, to be better on 1.9 we'd extend `config_defaults/subtests/docker_cli/kill.ini:kill_signals` from `1 31` to `1 64`. But that's beyond the scope of this work.

cevich:
hehe, so it will be > wrong :smile:   okay, merging...

PR #577, opened 2016-03-11
Monkey fix and avoid empy subtest throwing error

cevich:
This gets around a special case where a subtest has only a single sub-subtest which is excluded (control file cannot inspect sub-subtests, only identify their names).


cevich:
@edsantiago is this better / ok?

edsantiago:
Yep, LGTM

PR #580, opened 2016-03-24
bz1097344 (docker-1.10 exit codes): prep work

edsantiago:
Background: 'docker run' exit codes will be changing in 1.10,
to differentiate between container exit status and errors from
docker itself[1]. This commit prepares for that, by
redefining the use of the docker_expected_status config
setting. Instead of 'PASS' or 'FAIL' (old values), it
is now expected to be an integer: 0 for success, X
for failure. As of this commit X is always 1, but at
some future point that will change. When it does,
the new failif_ne() helper will display the new exit
code, and it will be easy to update .ini files as
necessary.

Additionally, the mustfail() helper now requires the
expected exit status as an argument. These statuses can
actually be hardcoded in the code itself, because
mustfail() tests (once) if the installed docker binary
returns unique exit codes; if it doesn't, it changes
the expectation to 1 (what old-docker always returns).

Also includes two unrelated fixes for minor problems
discovered during this process: removed an unused
config setting, and fixed an incorrect test helper.

 [1] https://github.com/docker/docker/pull/14012/files

Note that the exit codes here are UNTESTED, and will
require tweaking once we actually have docker-1.10
available. That is expected to be minor effort.

### DIFF
--- a/dockertest/output.py
+++ b/dockertest/output.py
@@ -33,6 +33,7 @@ class DockerVersion(object):
     _server_lines = None
     _client_info = None
     _server_info = None
+    _has_distinct_exit_codes = None
 
     def __init__(self, version_string=None, docker_path=None):
         # If called without an explicit version string, run docker to find out
@@ -179,7 +180,7 @@ class DockerVersion(object):
         """
         if self._client is None:
             try:
-                return self._old_client()
+                self._client = self._old_client()
             except xceptions.DockerOutputError:
                 self._client = self.client_info('version')
         if self._client is None:
@@ -193,7 +194,7 @@ class DockerVersion(object):
         """
         if self._server is None:
             try:
-                return self._old_server()
+                self._server = self._old_server()
             except xceptions.DockerOutputError:
                 self._server = self.server_info('version')
         if self._server is None:
@@ -227,6 +228,29 @@ class DockerVersion(object):
         :raises DockerTestNAError: installed docker < wanted
         """
         return self._require(wanted, 'client', self.client)
+
+    @property
+    def has_distinct_exit_codes(self):
+        """
+        2016-03-23 **TEMPORARY** - for transition from docker-1.9 to 1.10
+
+        docker-1.10 will introduce distinct exit codes to allow differentiating
+        between container status and errors from docker run itself; see
+        bz1097344 and docker PR14012. If we see an exit code of 125 here,
+        assume we're using the new docker.
+        """
+        if self._has_distinct_exit_codes is None:
+            try:
+                # docker-1.10 *must* support distinct exit codes
+                self.require_client('1.10')
+                has = True
+            except xceptions.DockerTestNAError:
+                # some builds of 1.9 might support it. (FIXME: really?)
+                d_run = utils.run('docker run --invalid-opt invalid-image',
+                                  ignore_status=True)
+                has = (d_run.exit_status > 120)
+            DockerVersion._has_distinct_exit_codes = has
+        return DockerVersion._has_distinct_exit_codes
 
 
 class ColumnRanges(Mapping):
@@ -759,33 +783,52 @@ def mustpass(cmdresult, failmsg=None):
     return cmdresult
 
 
-def mustfail(cmdresult, failmsg=None):
+def mustfail(cmdresult, expected_status=None, failmsg=None):
     """
     Check docker cmd results for pass. Raise exception when command passed.
 
     :param cmdresult: results of cmd.
     :type cmdresult: object convertible to string with variable exit_status
+    :param expected_status: exit status we expect to see from cmd.
+                            NOTE: for backward compatibility with pre-20160331
+                            code, we try to gracefully handle if this param
+                            is missing (default to 1) or a string (set failmsg)
+    :type expected_status: integer, 1-255
     :param failmsg: Additional messages for describing problem when cmd fails.
     """
+    # FIXME: temporary: backward compatibility for pre-20160330 code
+    # FIXME: remove before the next API-changing release
+    if expected_status is None:
+        expected_status = 1
+    if isinstance(expected_status, basestring):
+        failmsg = expected_status
+        expected_status = 1
+
     if failmsg is None:
         details = "%s" % cmdresult
     else:
         details = "%s\n%s" % (failmsg, cmdresult)
     if cmdresult is not None:
         OutputNotBad(cmdresult)
-    if cmdresult.exit_status == 0:
-        raise xceptions.DockerExecError("Unexpected zero exit code,"
-                                        " details: %s" % details)
+    # FIXME: temporary; remove once we no longer run on pre-1.10 docker
+    if not DockerVersion().has_distinct_exit_codes:
+        expected_status = 1
+    if cmdresult.exit_status != expected_status:
+        raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
+                                        " Details: %s" % (
+                                            cmdresult.exit_status,
+                                            expected_status,
+                                            details))
     return cmdresult
 
 
 # This class inherits a LOT of public methods, but most of what
 # appears here is all 'behind the scenes' stuff for producing
-# imutable instances.
+# immutable instances.
 class DockerTime(datetime.datetime):  # pylint: disable=R0903
 
     """
-    Imutable Docker specialized zulu-time representation
+    Immutable Docker specialized zulu-time representation
 
     :note: value is undefined when instance == instance.tzinfo.EPOCH.
            For example, the FinishedAt time of a still-running container
--- a/dockertest/output_unittests.py
+++ b/dockertest/output_unittests.py
@@ -156,7 +156,7 @@ class DockerVersionTest(unittest.TestCase):
         self.assertEqual(docker_version.client, '1.8.2')
         self.assertEqual(docker_version.server, '1.8.2')
 
-    def test_client_new(self):
+    def test_client_info_new(self):
         version_string = (
             "Client:\n"
             " Go version:   go1.4.2\n"
--- a/subtests/docker_cli/build/build.py
+++ b/subtests/docker_cli/build/build.py
@@ -225,7 +225,8 @@ class postprocessing(object):
             if OutputGood(build_def.dockercmd.cmdresult,
                           ignore_error=True,
                           skip=['error_check']):
-                return build_def.dockercmd.cmdresult.exit_status != 0
+                # FIXME: this is likely to fail in docker-1.10; bz1097344
+                return build_def.dockercmd.cmdresult.exit_status == 1
         else:
             raise DockerTestError("Command error: %s" % command)
 
--- a/subtests/docker_cli/dockerimport/truncated.py
+++ b/subtests/docker_cli/dockerimport/truncated.py
@@ -26,7 +26,7 @@ class truncated(empty):
         image_name = self.sub_stuff['image_name']  # assume id lookup failed
         if self.parent_subtest.config['remove_after_test']:
             dkrcmd = DockerCmd(self, 'rmi', [image_name])
-            mustfail(dkrcmd.execute())
+            mustfail(dkrcmd.execute(), 1)
 
     def run_tar(self, tar_command, dkr_command):
         tarfile = os.path.join(self.parent_subtest.bindir, self.TARFILENAME)
--- a/config_defaults/subtests/docker_cli/commit.ini
+++ b/config_defaults/subtests/docker_cli/commit.ini
@@ -9,7 +9,7 @@ subsubtests = good,check_default_cmd
 
 [docker_cli/commit/good]
 #: expected result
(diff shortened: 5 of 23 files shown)


edsantiago:
This is frustrating: pylint/pep8 don't issue these warnings on my system. I'll fix the comma-whitespace but will postpone global because I suspect there's a better way to preserve that state, and hope you'll guide me.

cevich:
> This is frustrating

Likely due to differences between the Travis Ubuntu environment and yours. I lined up the version numbers best I could with RHEL 7, but it's not an exact match.  If you have any idea of how to fix this, please feel free.  I tried many times myself, but maybe there's something I missed.

Yeah, globals are generally not good to use, esp. in core code b/c you can't predict when/if test-code might be multi-process/thread and require locking.  No problem storing the data in something the caller passes in though, b/c then they can lock it if necessary.  I'll take a peek at the code and comment more...

cevich:
Oh also, you're commit comments are lovely and well written, but way-overkill from my POV.  You're welcome to spend less time/effort on them if you wish :stuck_out_tongue_winking_eye:

dockertest/output.py
```diff
@@ -723,7 +723,7 @@ def mustpass(cmdresult, failmsg=None):
     return cmdresult
 
 
-def mustfail(cmdresult, failmsg=None):
+def mustfail(cmdresult, expected_status, failmsg=None):
     """
```
cevich: Might do a quick full-repo grep (or cscope) to see if there are any ``mustfail()`` callers that pass two positional-parameters, I suspect there may be either none, or a billion.  I was planning on merging this PR into the new 0.8.5 release, so the API changes are segregated.  Otherwise, just FYI -  consider when adding parameters, usually sticking them at the end as a keyword parameter is safest, and have it default to previous behaviour.

dockertest/output.py
```diff
@@ -737,11 +737,32 @@ def mustfail(cmdresult, failmsg=None):
...
-        raise xceptions.DockerExecError("Unexpected zero exit code,"
-                                        " details: %s" % details)
+    # FIXME: temporary; remove once we no longer run on pre-1.10 docker
+    if not has_distinct_exit_codes():
+        expected_status = 1
+    if cmdresult.exit_status != expected_status:
+        raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
+                                        " Details: %s" % (cmdresult.exit_status,
+                                                          expected_status,
+                                                          details))
```
cevich: Much better, a welcome improvement.

dockertest/output.py
```diff
@@ -737,11 +737,32 @@ def mustfail(cmdresult, failmsg=None):
...
+    """
+    global HAS_DISTINCT_EXIT_CODES
+    if HAS_DISTINCT_EXIT_CODES is None:
+        d_run = utils.run('docker run --invalid-option nonexistent-container',
+                          ignore_status=True)
+        HAS_DISTINCT_EXIT_CODES = (d_run.exit_status > 120)
+    return HAS_DISTINCT_EXIT_CODES
 
 # This class inherits a LOT of public methods, but most of what
 # appears here is all 'behind the scenes' stuff for producing
```
cevich: A few ideas here:
    - Make it an immutable bool-subclass, maybe singleton-pattern.  Might be overcomplicating it though.
    - Have it accept a 'state' parameter that defaults to None, then is returned as something which evaluates true/false but is mutable (e.g. ``[True]`` or ``[False]``).
    - Have ``mustfail()`` accept a callable parameter, which defaults to ``has_distinct_exit_codes``, then it just calls it instead of looking up a global.  (You'll need to flip around the function definition order).

dockertest/output.py
```diff
@@ -737,11 +737,32 @@ def mustfail(cmdresult, failmsg=None):
...
+        raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
+                                        " Details: %s" % (cmdresult.exit_status,
+                                                          expected_status,
+                                                          details))
     return cmdresult
 
+HAS_DISTINCT_EXIT_CODES = None
+def has_distinct_exit_codes():
+    """
+    2016-03-23 **TEMPORARY** - for transition from docker-1.9 to 1.10
```
cevich: FYI- These types of notes normally stick around /way/ longer than anyone ever intended, just look around :grinning:   These doc-strings will wind up in docker-autotest.readthedocs.org so short-and-sweet and tastefully-terse is usually the best way to go.

subtests/docker_cli/build/build.py
```diff
@@ -225,7 +225,8 @@ def basic_postprocess(build_def, command, parameter):
             if OutputGood(build_def.dockercmd.cmdresult,
                           ignore_error=True,
                           skip=['error_check']):
-                return build_def.dockercmd.cmdresult.exit_status != 0
+                # FIXME: this is likely to fail in docker-1.10; bz1097344
+                return build_def.dockercmd.cmdresult.exit_status == 1
```
cevich: Nice, this one actually is helpful, even if it winds up being permanently-temporary (like the rest) :grinning:

subtests/docker_cli/dockerimport/truncated.py
```diff
@@ -26,7 +26,7 @@ def cleanup(self):
         image_name = self.sub_stuff['image_name']  # assume id lookup failed
         if self.parent_subtest.config['remove_after_test']:
             dkrcmd = DockerCmd(self, 'rmi', [image_name])
-            mustfail(dkrcmd.execute())
+            mustfail(dkrcmd.execute(), 1)
```
cevich: Ahh okay, you did grep/cscope them all out, nice!

cevich:
Okay, gave code a gander + comments.  This is really fantastic for a first-pass, should be really easy to sort out.  I'm going to flag it for 0.9.1 b/c API changes - there are tests which exist outside this repo. and may use some of the changed calls.  Though if you can get it to be backwards-compatible, then I can merge it into the 0.8.x series.

dockertest/output.py
edsantiago: My reasoning is that despite my best intentions, I may forget to remove this after docker-1.10. I want the next maintenance programmer to see this comment, read the datestamp, realize this code can be excised, and do so with minimal wasted time. Is there a better way to accomplish that?

edsantiago:
Thanks for your comments, and especially on a non-work day. I need to fix that global issue, and will take a look at it on Monday.

dockertest/output.py
```diff
@@ -171,7 +172,7 @@ def client(self):
         """
         if self._client is None:
             try:
-                return self._old_client()
+                self._client = self._old_client()
```
edsantiago: I'm pretty sure this was meant to be memoized.

dockertest/output.py
```diff
@@ -185,13 +186,29 @@ def server(self):
         """
         if self._server is None:
             try:
-                return self._old_server()
+                self._server = self._old_server()
```
edsantiago: I'm pretty sure this was intended to be memoized.

dockertest/output_unittests.py
```diff
@@ -156,7 +156,7 @@ def test_client_new(self):
         self.assertEqual(docker_version.client, '1.8.2')
         self.assertEqual(docker_version.server, '1.8.2')
 
-    def test_client_new(self):
+    def test_client_info_new(self):
```
edsantiago: Expand the context, so you can see that this used to have the same name as the previous function. Hence one of them was masked.

edsantiago:
> Though if you can get it to be backwards-compatible, then I can merge it into the 0.8.x series.

That's the whole idea (backward compatibility). This works with 1.8, and we have a window in which to discover what the actual docker-1.10 exit codes are and update the tests without 1.8 or 1.9 being affected.

dockertest/output.py
cevich: damnit, who wrote that!  :blush:

dockertest/output_unittests.py
cevich: Nice catch!

cevich:
> That's the whole idea (backward compatibility). This works with 1.8, and we have a window in which to discover what the actual docker-1.10 exit codes are and update the tests without 1.8 or 1.9 being affected.

Great, okay then be sure to double-check you're not changing parameter ordering then.  Callers don't always pass keywords and there are tests that exist outside of the repository.

edsantiago:
>there are tests that exist outside of the repository.

Ouch - really? If so, and if any of that code invokes `mustfail` with the failmsg option, it will break. Do you have a sense for how likely that is, and how much pain it would be to have the maintainers of said code update it?

cevich:
> Ouch - really?

Yes, I have one to be sure, and I know at least the fedora guys are running Docker Autotest in some capacity.  Since it's a design feature to allow external (out-of-repo) test code to be used, that's why I keep a careful eye on API changes.  I honestly can't tell you if/how many outside tests there are, there's no requirement to publish.  It could be zero, it could be hundreds :fish:  

TBH even if we put this into 0.9.1, I control the release tagging, so we could do that tomorrow if needed :smile:   So it's not a major problem, just makes CI testing esaier b/c they can version-lock and be guaranteed the core API won't change.

Anyway, we'll figure something out.

edsantiago:
Okay... new push includes two fixes of note: (1) backward-compatibility kludge for handling old invocations of mustfail(), and (2) use new `require_client` to test for 1.10. Also a FIXME asking if it's a good idea to do the manual exit-code check as well. Also, rebased.

dockertest/output.py
```diff
@@ -767,25 +791,37 @@ def mustfail(cmdresult, failmsg=None):
     :type cmdresult: object convertible to string with variable exit_status
     :param failmsg: Additional messages for describing problem when cmd fails.
     """
+    # FIXME: temporary: backward compatibility for pre-20160330 code
+    # FIXME: remove before the next API-changing release
```
cevich: oh gosh, I might need help remembering this one is here.

subtests/docker_cli/build/build.py
```diff
@@ -225,7 +225,8 @@ def basic_postprocess(build_def, command, parameter):
             if OutputGood(build_def.dockercmd.cmdresult,
                           ignore_error=True,
                           skip=['error_check']):
-                return build_def.dockercmd.cmdresult.exit_status != 0
+                # FIXME: this is likely to fail in docker-1.10; bz1097344
```
cevich: In fact it already is, wait until you see the results from running on fedora_rawhide  :sob:

cevich:
Wow, this is very nice work.  I'd suggest don't drive yourself completely mad by fixing up all the old tests, it's a big time-sink.  That work is much easier to do during the 0.9.1 testing when all the 'big picture' stuff comes into play.  Anyway, just a suggestion.

Is this ready to merge do you think, or was there more looking/tweaking you wanted to do?

edsantiago:
One more commit, to fix a typo introduced in failif_ne

edsantiago:
Remaining concerns: I haven't tested with docker-1.8, I haven't seen the rawhide results you mention, I don't have a solution to your "need help remembering" issue. Aside from those, I believe this could be merged. Subsequent problems can be addressed as needed.

cevich:
Okay, we can test it by fire: run it through CI and see what breaks.  I'll give it a quick double-check then and merge...

dockertest/output.py
```diff
@@ -767,25 +791,37 @@ def mustfail(cmdresult, failmsg=None):
...
+    if isinstance(expected_status, basestring):
+        failmsg = expected_status
+        expected_status = 1
+
```
cevich: hehe, cheap hack.  Okay, I'll buy that :grinning:

dockertest/output.py
```diff
@@ -767,25 +791,37 @@ def mustfail(cmdresult, failmsg=None):
     :type cmdresult: object convertible to string with variable exit_status
     :param failmsg: Additional messages for describing problem when cmd fails.
```
cevich: Ahh, we need a docstring update here for ``expected_status``.  Might be good to mention there that it will be used in place of  ``failmsg`` if it's a string instead [of something else].

dockertest/output.py
cevich: Hmm, maybe this should also test if ``failmsg == None``?  Or is that being too paranoid?

cevich:
Okay, just found one little docstring that needs some love, then this is good-to-go.  There's some possible paranoia you're welcome to fix as well, but it's probably just my ignorance.

dockertest/output.py
edsantiago: Good call. Testing for expected_status == None, though. This catches two old-style usages:

       mustfail(foo)           # missing exit status and failmsg
       mustfail(foo, "why")    # old-style call with message

dockertest/output.py
edsantiago: Good catch, thanks! New commit pushed.

cevich:
That's why I double-check, then blindly assume everything I didn't notice isn't a problem :grinning:   Okay, merging...

PR #581, opened 2016-03-28
new failif_ne test, for more helpful diagnostics

edsantiago:
failif(actual != expected) doesn't provide helpful diagnostics
on failure: the human reader might be able to figure out the
expected value by reading the code, but can't know actual.

This commit provides a new failif_ne(actual, expected) which,
on failure, displays both.


cevich:
Awesome, great work.  Thanks for patching up so many tests as well, that's well beyond "necessary".  Okay, where's my rubber-stamp, oh there it is :grinning:

PR #582, opened 2016-03-29
bz1320302 - docker --cgroup-parent=X

edsantiago:
Test that --cgroup-parent=X and X-Y behave as expected.
Test valid only on docker >= 1.9.1-25.el7 and >= 1.10.

### DIFF
--- a/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
+++ b/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
@@ -1,0 +1,187 @@
+r"""
+Summary
+----------
+
+Tests for bz1320302 - odd behavior of --cgroup-parent option.
+
+Operational Summary
+----------------------
+
+#. Start container with --cgroup-parent=X, for various values of X
+#. Run container command 'cat /proc/1/cgroup'
+#. Confirm that each line (controller) lists the correct cgroup path
+"""
+
+
+import os
+import os.path
+import re
+from autotest.client import utils
+from dockertest.dockercmd import DockerCmd
+from dockertest.images import DockerImage
+from dockertest.output import OutputGood, DockerVersion
+from dockertest.subtest import SubSubtest, SubSubtestCaller
+from dockertest.xceptions import DockerTestError, DockerTestFail
+
+
+class run_cgroup_parent(SubSubtestCaller):
+    pass
+
+
+class run_cgroup_parent_base(SubSubtest):
+    def _setup(self, cg_parent):
+        """
+        Set up test parameters: the value of the --cgroup--parent option.
+        String may include {rand1} and/or {rand2}; we replace those with
+        two pseudorandom strings.
+        """
+        DockerVersion().require_client("1.10")
+        self.sub_stuff['rand1'] = utils.generate_random_string(8)
+        self.sub_stuff['rand2'] = utils.generate_random_string(8)
+        self.sub_stuff['cgroup_parent'] = cg_parent.format(**self.sub_stuff)
+
+    def _expect(self, path=None, stderr=None):
+        """
+        Set up test parameters: what to expect on completion.
+        The format strings {rand1} and {rand2} will be replaced with the
+        same values as in _setup, and the string {cid} with the container ID
+        once we have it.
+        """
+        self.sub_stuff['expected_path'] = path if path else ''
+        self.sub_stuff['expected_stderr'] = stderr if stderr else ''
+
+    def run_once(self):
+        """
+        Run docker with the given --cgroup-parent; preserve cgroup info.
+        """
+        super(run_cgroup_parent_base, self).run_once()
+
+        cidfile = os.path.join(self.tmpdir, 'cidfile')
+        self.sub_stuff['cidfile'] = cidfile
+
+        subargs = ['--rm',
+                   '--cgroup-parent=%s' % self.sub_stuff['cgroup_parent'],
+                   '--cidfile', cidfile,
+                   DockerImage.full_name_from_defaults(self.config),
+                   '/bin/cat', '/proc/1/cgroup']
+        dc = DockerCmd(self, "run", subargs,
+                       timeout=self.config['docker_timeout'])
+        self.sub_stuff["cmdresult"] = dc.execute()
+
+    def postprocess(self):
+        super(run_cgroup_parent_base, self).postprocess()
+
+        cmdresult = self.sub_stuff["cmdresult"]
+        OutputGood(cmdresult, ignore_error=False)
+
+        self.sub_stuff["cid"] = self._read_cid()
+        path_exp = self.sub_stuff['expected_path'].format(**self.sub_stuff)
+        stderr_exp = self.sub_stuff['expected_stderr'].format(**self.sub_stuff)
+
+        # Check stderr first: that way if we're not expecting an error, but
+        # get one, our user has a better chance of seeing something useful.
+        # (Note: Running docker with -D (debug) produces lots of cruft:
+        #    time="...." level=debug msg="unwanted stuff"
+        # Strip them out.)
+        stderr = "\n".join([line
+                            for line in cmdresult.stderr.strip().split("\n")
+                            if not line.startswith('time="')])
+        self.failif_ne(stderr, stderr_exp, "stderr")
+
+        # If we're expecting stdout, it must contain multiple lines each
+        # of the form:
+        #    <num>:<cgroup controller>:<path>
+        # ...where <path> must exactly match the one in our test setup.
+        stdout = cmdresult.stdout.strip()
+        if path_exp:
+            re_cgroup = re.compile(r'^(\d+):([^:]+):(.*)')
+            found_match = False
+            for line in stdout.split("\n"):
+                m = re.match(re_cgroup, line)
+                if m is None:
+                    raise DockerTestFail("cgroup line does not conform to"
+                                         " '<n>:<controller>:<path>': '%s'"
+                                         % line)
+                self.failif_ne(m.group(3), path_exp, "cgroup path for %s:%s"
+                               % (m.group(1), m.group(2)))
+                found_match = True
+
+            # Must find at least one matching cgroup line
+            if not found_match:
+                raise DockerTestFail("No output from cgroups")
+        else:
+            self.failif_ne(stdout, '', "unexpected output on stdout")
+
+    def _read_cid(self):
+        """
+        Read container ID from --cidfile file, so we can replace {cid}
+        in format string and do an exact match on expected values.
+        """
+        cid = open(self.sub_stuff['cidfile'], 'rb').read().strip()
+        if len(cid) < 12:
+            raise DockerTestError("bad cid (length < 12) in --cidfile")
+        return cid
+
+    def cleanup(self):
+        """
+        Remove stray cgroups
+        """
+        cgroups_dir = '/sys/fs/cgroup'
+        try:
+            rand1 = self.sub_stuff['rand1']
+            rand2 = self.sub_stuff['rand2']
+        except KeyError:                # Test never ran
+            return
+
+        for ent in os.listdir(cgroups_dir):
+            # eg /sys/fs/cgroup/memory
+            controller_dir = os.path.join(cgroups_dir, ent)
+            if os.path.isdir(controller_dir):
+                # eg /sys/fs/cgroup/memory/foo.slice
+                rand1_dir = os.path.join(controller_dir, "%s.slice" % rand1)
+                if os.path.isdir(rand1_dir):
+                    # eg /sys/fs/cgroup/memory/foo.slice/foo-bar.slice
+                    rand2_dir = os.path.join(rand1_dir,
+                                             "%s-%s.slice" % (rand1, rand2))
+                    # Note that this is special cgroupfs magic! rmdir
+                    # succeeds even though ls lists "files" within.
+                    if os.path.isdir(rand2_dir):
+                        os.rmdir(rand2_dir)
+                    os.rmdir(rand1_dir)
+
+
+class run_cgroup_parent_invalid_name(run_cgroup_parent_base):
+    """
+    Invalid value for --cgroup-parent option
+    """
+
+    def initialize(self):
+        super(run_cgroup_parent_invalid_name, self).initialize()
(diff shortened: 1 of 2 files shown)


config_defaults/subtests/docker_cli/run_cgroups.ini
```diff
@@ -3,7 +3,7 @@
 cgroup_path = /sys/fs/cgroup/memory/system.slice/docker
 #: which key is used in this test
 cgroup_key_value = memory.limit_in_bytes
-subsubtests = cpu_positive,cpu_zero,memory_positive,memory_no_cgroup,cpu_none,memory_negative,cpu_overflow
+subsubtests = cpu_positive,cpu_zero,memory_positive,memory_no_cgroup,cpu_none,memory_negative,cpu_overflow,cgroup_parent_invalid_name,cgroup_parent_path,cgroup_parent_path_with_hyphens
```
edsantiago: I would recommend committing this addition, but disabling it in CI until the appropriate versions of docker wend their way into our testing. Is there an automated way to do that?

subtests/docker_cli/run_cgroups/run_cgroups.py
```diff
@@ -55,3 +56,42 @@ class memory_no_cgroup(memory_base):
...
+    """
+    Invalid value for --cgroup-parent option
+    """
+
+    def initialize(self):
+        super(cgroup_parent_invalid_name, self).initialize()
+        self._setup("/foobar",
+                    "",
+                    "Error response from daemon: Cannot start container {cid}:"
+                    " [8] System error: Invalid slice name /foobar")
```
edsantiago: I know the traditional way is to define these in a .ini file, and leave the class empty, but I just find that too painful to understand. How do you do it? Do you have tools or IDE or something that can superimpose the .ini settings with the code, making it possible to understand in one glance? What am I missing?

cevich: I don't believe I wrote this test, or at least haven't looked at it in a few decades or three.  The code is normally all in the parent class, so the empty sub-subtest class is really just there to create the name.  The name needs to be held so it can be found/loaded by the ``SubsubtestCaller()`` (or w/e it's name is) in the subtest API.  No fancy IDE's just two side-by-side terminal windows :grinning:

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
+"""
```
cevich: BTW, don't feel compelled to add new sub-subtests into existing subtest modules if they don't fit or it would be ugly.  Feel free to create new top-level subtests as needed.

edsantiago: Thanks for the reality check. I started under run_cgroups because I thought it fit in well with that subtest; now it's changed enough that it makes sense to split out. Would run_cgroup_parent be an acceptable name?

subtests/docker_cli/run_cgroups/run_cgroups.py
cevich: Reading more, no I think what you have here with the call to _setup() is fine.  There are many other tests that rely on shared initialization functions and overriding ``initialize()``.  No prob. doing it this way.

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
...
+from autotest.client import utils
+from dockertest.dockercmd import DockerCmd
+from dockertest.images import DockerImage
+from dockertest.output import OutputGood
+from dockertest.xceptions import DockerTestError, DockerTestFail
+from cgroups_base import cgroups_base
+
+
+class cgroup_parent_base(cgroups_base):
+    def _setup(self, cg_parent, expected_path, expected_stderr):
```
cevich: pylint might get mad there's no space between class/def.  I'll let it be the judge though :smile:

config_defaults/subtests/docker_cli/run_cgroups.ini
edsantiago: Ugh. No clean solution, then. OK, back to the drawing board.

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
...
+        stderr_exp = self.sub_stuff['expected_stderr'].format(**self.sub_stuff)
+
+        # Check stderr first: that way if we're not expecting an error, but
+        # get one, our user has a better chance of seeing something useful.
+        # (Note: Running docker with -D (debug) produces lots of cruft:
+        #    time="...." level=debug msg="unwanted stuff"
+        # Strip them out.)
+        stderr = "\n".join([line
+                            for line in cmdresult.stderr.strip().split("\n")
+                            if not line.startswith('time="')])
```
cevich: Just wait and see, they'll change the format in the next version, then three more times after that..  Feel free to just examine ``self.config['docker_options']`` and strip out any '-D' since it impacts the test so greatly.  TBH, I'm not sure I even like the '-D' in there by default.  It's not really ever been that useful, even when things do go south.  I'd also support nuking it from ``defaults.ini``, but I suppose you never know when it might get switched back on.  Uggg, it's never easy.

Maybe stderr only needs to be examined if the exit code is non-zero?

Hmmm.

Failing to think how this could be made simpler / avoided entirely.

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
...
+                                         % line)
+                self.failif_ne(m.group(3), path_exp, "cgroup path for %s:%s"
+                               % (m.group(1), m.group(2)))
+                found_match = True
+
+            # Must find at least one matching cgroup line
+            if not found_match:
+                raise DockerTestFail("No output from cgroups")
+        else:
+            self.failif_ne(stdout, '', "unexpected output on stdout")
```
cevich: Nice, well written, easy to read.

subtests/docker_cli/run_cgroups/run_cgroups.py
```diff
@@ -55,3 +56,41 @@ class memory_no_cgroup(memory_base):
...
+class cgroup_parent_path_with_hyphens(cgroup_parent_base):
+    """
+    Hyphens result in a more complicated path.
+    """
+
+    def initialize(self):
+        super(cgroup_parent_path_with_hyphens, self).initialize()
+        self._setup("{rand1}-{rand2}.slice",
+                    "/{rand1}.slice/{rand1}-{rand2}.slice/docker-{cid}.scope",
+                    "")
```
cevich: I think this is fine, I agree it's a little hard to follow and keep straight, but it's a rush-request and plenty of other tests do it this way.  If you want to try a different approach on a less-urgent test, please feel free, I'm happy to take a look.

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
...
+        OutputGood(cmdresult, ignore_error=False)
+
+        # Expected output may include the container ID; allow caller
+        # to use "{cid}", which we replace with actual one for this run.
+        cid = open(self.sub_stuff['cidfile'], 'rb').read().strip()
+        if len(cid) < 12:
+            raise DockerTestError("bad cid (length < 12) in --cidfile")
+        self.sub_stuff['cid'] = cid
+        path_exp = self.sub_stuff['expected_path'].format(**self.sub_stuff)
+        stderr_exp = self.sub_stuff['expected_stderr'].format(**self.sub_stuff)
```
cevich: This will make expanding/extending the test later (if necessary) a lot easier to do.

subtests/docker_cli/run_cgroups/cgroup_parent.py
```diff
@@ -0,0 +1,103 @@
...
+        for --cgroup-parent, and what to expect in stdout & stderr.
+        We replace the format strings {rand1} and {rand2} with two
+        pseudorandom strings, and {cid} with the container ID (once
+        we have it).
+        """
+        self.sub_stuff['rand1'] = utils.generate_random_string(8)
+        self.sub_stuff['rand2'] = utils.generate_random_string(8)
+        self.sub_stuff['cgroup_parent'] = cg_parent.format(**self.sub_stuff)
+        self.sub_stuff['expected_path'] = expected_path
+        self.sub_stuff['expected_stderr'] = expected_stderr
```
cevich: BTW (in case you were wondering), test-data is kept in ``stuff`` and ``sub_stuff`` (they could be merged to use same name) is to keep the runtime namespace separated from the class and API.  It's just a simple way to avoid smashed toes (which has happened, along days spent tracking down ``-/_`` mix-ups and speling mistakes).

subtests/docker_cli/run_cgroups/run_cgroups.py
cevich: Yet another idea: Just copy paste all the ``self.sub_stuff[blah] = blah`` lines into a separate ``initialize()`` call for each sub-subtest.  I realize it's duplication/repetition, however it's only a few lines, and it would allow one or more sub-subtests to diverge in the future w/o having to add all kinds of extra parameters/logic into the common ``_setup()`` method.  Anyway, something to consider because- if you're going to check/raise ``DockerTestNAError`` on too low a version, doing so inside ``initialize()`` may make more sense than inside ``_setup()`` (name wise).  But I'm not too particular on these things.  Just some ideas.

cevich:
Uggg, I'm seriously mad this dumb 'Type less with saved replies' box keeps popping up.

Changes/additions look great, well written, easy to read and understand.  I shared some ideas, really the only thing I'd like to see for sure is the raising of ``DockerTestNAError`` when docker version < 1.10  (I triple checked this is what martin wanted, even lokesh asked him about it).

Nice job, especially for a hot-potato.

config_defaults/subtests/docker_cli/run_cgroups.ini
edsantiago: Double ugh. 'docker version' returns 1.9.1, API Version 1.21, on both good and bad builds. Package version returns docker-1.9.1-25.el7.x86_64, but that won't help with Fedora/CentOS. Taking a step back now to let this simmer in my background brain.

subtests/docker_cli/run_cgroups/cgroup_parent.py
edsantiago: Thanks for the clarification on `stuff` vs `sub_stuff`! Any guideline on which one is preferred for future code?

subtests/docker_cli/run_cgroups/cgroup_parent.py
edsantiago: You might have noticed that the -D is also affecting CI results on Fedora. I don't know enough to have a strong opinion on keeping -D vs removing it, but my uninformed preference right now is to nuke it.

subtests/docker_cli/run_cgroups/run_cgroups.py
edsantiago: My first approach was to overload `super.initialize`, passing those three extra args. I didn't feel comfortable with that because it changed the `initialize` signature and expectation. But would it be OK?

subtests/docker_cli/run_cgroups/cgroup_parent.py
cevich: Yep that's fine, I'm not too particular on the names as long as they make sense (which I fail at often).

config_defaults/subtests/docker_cli/run_cgroups.ini
cevich: Just check and run test if >= '1.10' - those were our orders, and will help you manage your caffeine-quota :smile:

subtests/docker_cli/run_cgroups/cgroup_parent.py
cevich: My preference would be in the future (like for 1.9.1) we'd just edit ``dockertest.subtestbase`` and call them all 'stuff'.  There was a (now irrelevant) reason for the differentiation, now it's just a PITA.

subtests/docker_cli/run_cgroups/run_cgroups.py
cevich: Ya, signature for those methods is burried deep and in non-obvious ways inside Autotest.  TBH, it's only like 5-7 items, I'd just copy-paste them into an overloaded ``initialize()`` w/o arguments.  That allows future modification to whomever's heart's content w/o any affect on other code.

cevich: ...splitting this all off into your own subtest is fine too, then you're not held to any old duct-tape previously applied to this ancient code.

edsantiago: Already split. (And, sidenote, I added a `cleanup` to remove stray cgroups). I will push tomorrow, not today, because my brain is shutting down. I'll also look into refactoring all the DockerVersion checks, because with this addition we'll have three and that's too many to duplicate.

subtests/docker_cli/run_cgroups/cgroup_parent.py
cevich: okay, opened #583

edsantiago:
1.10 check now in place.

subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
```diff
@@ -0,0 +1,187 @@
...
+class run_cgroup_parent_path_with_hyphens(run_cgroup_parent_base):
+    """
+    Hyphens result in a more complicated path.
+    """
+
+    def initialize(self):
+        super(run_cgroup_parent_path_with_hyphens, self).initialize()
+        self._setup("{rand1}-{rand2}.slice")
+        self._expect(path="/{rand1}.slice/{rand1}-{rand2}.slice"
+                     "/docker-{cid}.scope")
```
cevich: hahaha, I had a feeling starting fresh would be a lot easier.  That other cgroups test is quite crusty.

cevich:
Okay, merging this, it'll show up in the internal 'autobrew' CI builds, those pull the latest/greatest RPMs from everything (which is why it's always broken) :grinning:

PR #583, opened 2016-03-29
Re-locate -D to it's more useful place

cevich:
Though in CI, generally daemon_options is overridden, having it here by
default is more useful than having it on the client.  The later is
infrequently useful and at worse problematic for output parsers.  Easier
to just drop it.

### DIFF
--- a/config_defaults/defaults.ini
+++ b/config_defaults/defaults.ini
@@ -17,10 +17,10 @@ autotest_version = @!NOVERSIONCHECK!@
 docker_path = /usr/bin/docker
 
 #: Global docker client command options to use (CSV)
-docker_options = -D
+docker_options =
 
 #: Global docker daemon options to use (CSV)
-daemon_options = -d,--selinux-enabled,--add-registry,registry.access.redhat.com
+daemon_options = daemon,-D,--selinux-enabled
 
 #: Max runtime in seconds for any docker command (auto-converts to float)
 docker_timeout = 300.0


edsantiago:
LGTM, but might not a future reader appreciate a commit-message blurb about the changes to daemon_options? Like maybe "docker daemon is more readable than docker -d, and the registry is better suited to CI config than a default"?

Also, where is daemon_options references anyway?

cevich:
Thanks for taking a look.  Docker has long since replaced '-d' with 'daemon' on the command-line for startup, I just forgot to update the config.  There are/were a handful of tests which restart the daemon with additional/replaced options.  When they run ``cleanup()`` they restart the daemon again using ``daemon_options`` from the config.  In CI, before starting any testing, we probe the running daemon process and transfer all it's options into the configuration.  So this default value is really just for runs outside of CI (which is why I forgot to update it).

edsantiago:
I'm happy with the changes, I'm just less so with the commit message. I wonder if this is a difference in our working styles, since you mentioned a few days ago that my commit messages are "overkill". Might I offer the reasoning behind my preference?

Sometimes it's necessary to run `blame` on a file, to see what changed when and why. (Today I use `tig` but in past it's been git, cvs, hg, rcs). When I view the changelog for a line, and there's no explanation for it, my stress levels rise. Was that intentional? Accidental? Overlooked? Not important enough to mention? Well, what if today (from the perspective of the future maintainer) it _is_ important, and breaking something? I want my future maintainer to know why something needed changing, and I can't assume that they'll have or remember full context of today. So for this my commit message would be something like:

    -D (debug) option: remove from client, add to server

    Reason: the -D option causes docker client to issue mostly-unuseful
    trace messages to stderr, interfering with tests that actually check stderr.
    Remove it. Add it to docker server, where it actually (FIXME: explain
    why it's useful here).

    Also: 'docker daemon' is the preferred way to start the server, replacing -d.
    And there's no reason to specify registry XYZ when starting.

Does that make sense? Have you ever had an instance where you're spelunking down a deep commit history, find the commit that caused a change, and are stymied by an unhelpful commit message?

cevich:
> unhelpful commit message?

Every now and again, so far never while the person-to-be-blamed is still around to ask, but I get your point, and for a public, open-source project, you're style is far superior.  Likely this is a matter of different experience levels, my impatience, and pressure to keep the entire house-of-cards from falling down.

Now, if some _other_ engineers would examine the trello backlog + the number of 'new test' issues here, maybe things would be easier :stuck_out_tongue_winking_eye: Until then, I'll try to make my commit messages better, but if you see any please feel free to call me on it, especially if that includes you writing a proper one so I can compare (then copy-paste and plagiarize yours without guilt) :grinning:

cevich:
There, fully plagiarized and expanded.  Thanks for the help!

PR #584, opened 2016-03-29
Refactor: new subtest.require_docker_version()

edsantiago:
Raises DockerTestNAError if currently-installed docker
version is less than a specified requirement.

Also: removed obsolete version checks and code from rmi

### DIFF
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -24,6 +24,11 @@ from autotest.client import test
 import version
 import config
 import subtestbase
+# There is a bug in Pylint + virtualenv that makes this fail in Travis CI
+# https://github.com/PyCQA/pylint/issues/73
+from distutils.version import LooseVersion  # pylint: disable=E0611,F0401
+from dockercmd import DockerCmd
+from output import DockerVersion
 from xceptions import DockerTestFail
 from xceptions import DockerTestNAError
 from xceptions import DockerTestError
@@ -290,6 +295,22 @@ class SubSubtest(subtestbase.SubBase):
         logging.warning("SubSubtest.make_subsubtest_config() is deprecated!")
         self.config = self.make_config(all_configs, parent_config,
                                        self.config_section)
+
+    def require_docker_version(self, wanted):
+        """
+        Run 'docker version', parse client version, compare to wanted.
+
+        :param wanted: required docker version
+        :raises DockerTestNAError: installed docker < wanted
+        """
+        stdout = DockerCmd(self, "version").execute().stdout
+        dockerversion = DockerVersion(stdout)
+        client_version = LooseVersion(dockerversion.client)
+        required_version = LooseVersion(wanted)
+        if client_version < required_version:
+            raise DockerTestNAError("Test requires docker version >= %s;"
+                                    " %s installed" % (required_version,
+                                                       client_version))
 
 
 class SubSubtestCaller(Subtest):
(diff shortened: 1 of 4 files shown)


dockertest/subtest.py
```diff
@@ -24,6 +24,11 @@
 import version
 import config
 import subtestbase
+# There is a bug in Pylint + virtualenv that makes this fail in Travis CI
+# https://github.com/PyCQA/pylint/issues/73
+from distutils.version import LooseVersion  # pylint: disable=E0611,F0401
+from dockercmd import DockerCmd
+from output import DockerVersion
```
cevich: Okay, this is probably fine.  I don't think we'll ever need to ``import subtest`` inside of ``output``, no?

dockertest/subtest.py
```diff
@@ -24,6 +24,11 @@
 import version
 import config
 import subtestbase
+# There is a bug in Pylint + virtualenv that makes this fail in Travis CI
+# https://github.com/PyCQA/pylint/issues/73
+from distutils.version import LooseVersion  # pylint: disable=E0611,F0401
```
cevich: Could save some future grief by just using the one from the already imported ``version`` module.  At some point (I hope) pylint will be fixed in travis (and match the versions we use in RHEL).  It's a minor issue though.

dockertest/subtest.py
```diff
@@ -291,6 +296,22 @@ def make_subsubtest_config(self, all_configs, parent_config,
         self.config = self.make_config(all_configs, parent_config,
                                        self.config_section)
 
+    def require_docker_version(self, wanted):
```
cevich: These classes are quite crowded, what do you think about putting this in as a method on the ``Dockerversion`` class itself?  There's no requirement to use ``DockerCmd`` for such a simple call, a ``utils.run()`` or a ``subprocess.check_call()`` would be perfectly fine as well - thus eliminating the dependencies on a ``subtest`` instance as well.  Thoughts?

subtests/docker_cli/create/create_signal.py
```diff
@@ -20,22 +14,9 @@ class create_signal(create_base):
...
-        else:
-            return True
-
     def initialize(self):
         super(create_signal, self).initialize()
-        if not self.is_newer_docker():
-            raise DockerTestNAError("Test does not support "
-                                    "docker version < %s",
-                                    self.non_zero_exit_version)
+        self.require_docker_version(self.non_zero_exit_version)
```
cevich: OMG this is a huge improvement, I like it.

cevich:
I'll do up a version with this moved into ``DockerVersion`` to see if it's any more than marginally better....

cevich:
...okay, hacked this together, totally untested.  See whatcha think: https://github.com/edsantiago/autotest-docker/pull/1

dockertest/subtest.py
edsantiago: YES! Thank you! That was my first approach, but I backed out after considering that `utils.run("docker")`  might not be the same as `config["docker_path"]`, and DockerVersion has no access to config settings. May I set aside that concern?

cevich: I think it's not so much an issue now, but only if/when we ever are testing multiple docker versions installed at the same time.  When that day comes, we can worry about this problem :smile: Until then, ``config["docker_path"]`` isn't all that critical IMHO.

cevich: Thought: since it's just that one config option, probably no harm in just adding a parameter for it.

edsantiago:
Back at ya: moved helper into DockerVersion initializer. Tested.

cevich:
Fantastic, I love it when a plan comes together :grinning:

edsantiago:
Rats! I was just writing unit tests for it :)

PR #585, opened 2016-03-30
Workaround NetworkSettings.Network data keys

cevich:
In docker 1.9.1 these key names come directly from the system, so no
style checking should be done (Bug 1322459).  However, the child items
for each network node do still need checking.  This workaround simply
substitutes the system network key-names with unlikely-to-clash
but style-conforming keys, retaining the original child values.  In the
case of lower versions of docker, this branch of the tree doesn't exist,
so the resulting KeyError is simply ignored.


cevich:
@edsantiago mind taking a quick peek?

cevich:
Hmm, having problems while testing this.  Investigating...

edsantiago:
What kind of problems? From reading, I can't find anything wrong. But, hint: this is the sort of thing that is perfectly suited for a unit test...

cevich:
> perfectly suited for a unit test...

Lol, exactly what I'm trying.  Not sure if problem is w/ the CI job or these changes.  Working on both at the same time FTW!

cevich:
Woops, I mean "kind of" what I'm trying.  I'd rather just run all the tests in the default environment.  That's the real "trial-by-live-fire" if I broke anything or not.  I also need to bang on the CI job so I can get it skoocum enough to put up for general (internal) use.

cevich:
Added unittest but it didn't help be spot the problem.  Somehow/someway the test is failing on the old docker 1.8 JSON when inspecting the ``Config.Labels`` keys which should have been filtered.  Unittest confirms the filter is working on the new (docker 1.9.1 keys).  Tomorrow I'll try adding another test for the older inspect JSON, but I think the bug is somehow in ``postprocess()``.

edsantiago:
Took the liberty of cleaning up unit tests and adding to PR.

subtests/docker_cli/dockerinspect/inspect_keys.py
```diff
@@ -43,13 +43,30 @@ def run_once(self):
...
         except KeyError:
-            pass
+            self.logdebug("Config.Labels branch not found?")
+        # These are data-keys, not structural, ignore names
+        try:
+            # New in docker 1.9.1
+            networks = json0['NetworkSettings']['Networks']
+            for idx, network in enumerate(networks.keys()):
+                # Child key names still need checking
+                dummy_key = 'NameIgnored%d' % idx
```
edsantiago: This loses the original name. Would you consider 'FilterBypass_%s' % network  ?

cevich: Oh! that's an even better idea, thanks!

edsantiago:
Waah! My nice test suite got clobbered! Was that intentional, or did you miss doing a git pull of my commit from last night?

cevich:
> Waah!

Uh oh, there was a pull? :blush:  Uhhh, okay, mmmm, my notifications might be messed up, did you send me a PR?

edsantiago:
No, I'm still figuring out the github flow. Stay tuned.

subtests/docker_cli/dockerinspect/inspect_keys.py
cevich: Oh whew, thought I lost something.  What I've found easiest, is to clone down my forked branch (the source of this PR), mangle^H^H^H^Hfix it however you like, then send it to me as a PR (against my personal fork/branch).  Hint: It's the drop-downs on the top-left in the 'compare' screen, not the ones on the left :grinning:

cevich:
OIC it's in  #586  okay, I'll stop pushing to this one.

cevich:
Closed in favor of #586

PR #586, opened 2016-03-31
Pr/585

edsantiago:
fixes #585


cevich:
oic here it is.

edsantiago:
Yeah... working on pylint issues now.

cevich:
pylint/pep8 are going to give you fits and rage with all the in-line json.  That's why I dumped it all into a file.

cevich:
FYI: Tou can't re-target a PR once it's created.  Instead, when creating it you change the repo/branch on the lhs to point at ``cevich/autotest-docker:fix_inspec_keys``, it will create a PR against my fork instead of the main repo.  That's how I created https://github.com/edsantiago/autotest-docker/pull/1

subtests/docker_cli/dockerinspect/test_inspect_keys.py
```diff
@@ -0,0 +1,270 @@
...
+    "Created": "2016-03-30T17:51:47.675978044Z",
+    "Path": "tail",
+    "Args": [
+        "-n1",
+        "/proc/1/cgroup"
+    ],
+    "State": {
+        "Status": "exited",
+        "Running": false,
+        "Paused": false,
```
cevich: like this, ``false`` is from java, python is more sophisticated, it needs an upper case `F` :grinning:

subtests/docker_cli/dockerinspect/test_inspect_keys.py
```diff
@@ -0,0 +1,270 @@
...
+    "Name": "/modest_colden",
+    "RestartCount": 0,
+    "Driver": "devicemapper",
+    "ExecDriver": "native-0.2",
+    "MountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c3,c28",
+    "ProcessLabel": "system_u:system_r:svirt_lxc_net_t:s0:c3,c28",
+        "AppArmorProfile": "",
+    "ExecIDs": null,
+    "HostConfig": {
+        "Binds": null,
```
cevich: ``null`` -> ``None``

I'm tellin' ya man, it's gunna drive ya nuts.  Best just dump it back into a test.json file :grinning:

cevich: or encode it all inside a string - but then you can't easily use ``json_verify`` and ``json_reformat``.

edsantiago:
It has to be a string. Look at the explanation.

subtests/docker_cli/dockerinspect/test_inspect_keys.py
```diff
@@ -0,0 +1,270 @@
...
+[
+{
+    "This":"Should"         | "Cause": "A Test Failure"
+}
+]
+""")
+        except AssertionError:
+            raised = True
+        if not raised:
+            self.fail("Known-bad test did not raise exception.")
```
cevich: Hehe, I think in the time you've been working on this, a few runs through my CI job would have been more time-effective.  I'll concentrate on getting that working and leave the real programming to you :grinning:

cevich:
> It has to be a string. Look at the explanation.

Right, not valid json.  Hmmm, ya best stick it in as a string-blob then instead of python objects.

edsantiago:
So, only if this test style makes sense. My goal in writing tests is writing code that is easy to maintain and even easier to extend when the situation changes, e.g. when the filter code needs new tweaks. I like the left-hand right-hand style because it's a simple way of highlighting diffs while keeping the rest clean. But I'd like to know if you find it readable/maintainable.

cevich:
I guess my main concern is the balance required prioritizing work-on-tests vs work-on-testing.  Almost always, management's preference is the later.  I tend to agree with management, the tests come and go and change fairly regularly.  I think more likely is the unittests just get dropped completely over time, possibly even along with the test.  So moving the test-testing into CI runs on docker-autotest PR's should be more time-effecient (for us) than writing/maintaining test-level unittests....but I reserve the right to be completely wrong :grinning:

cevich:
(that is to say, let's give test-unittests a whirl on a small scale, along with CI runs and then decide which is easier/better).

cevich:
Okay, this PR looks good-nuf, still unsure if it actually fixes the problem w/ the test failing, but time will tell, and then more PRs can follow :grinning:

cevich:
This was merged as 0d7eb8f298b010a8998a390542a3b0d0d0df26cb

PR #587, opened 2016-03-31
dockerinspect/inspect_keys : trivial bug fix

edsantiago:
Seems like a simple mistype: code was attempting to
reference the wrong dict, leading to test failure.


cevich:
@edsantiago for little trivial things that don't change any API's, please feel free to just merge them, you should have commit access to master.  I only set those labels / milestones out of habit, and because when I do the next release, it makes searching the PRs from prior one easier :grinning:

cevich:
oh, and thanks!  Sometimes looking at a problem for sooooo long makes it harder to spot typos like that :confounded:

PR #588, opened 2016-03-31
Workaround for bz1317096 : clean up stale images

edsantiago:
The rmi test is leaving stale images behind, causing
garbage_check to fail on it and all subsequent tests.
This harmless cleanup code deletes images even if
their name has been clobbered by bz1317096.

### DIFF
--- a/subtests/docker_cli/rmi/rmi.py
+++ b/subtests/docker_cli/rmi/rmi.py
@@ -239,3 +239,8 @@ class with_blocking_container_by_id(with_blocking_container_by_tag):
     def check_image_exists_by_id(self, image_id):
         di = DockerImages(self)
         return di.list_imgs_with_image_id(image_id)
+
+    def cleanup(self):
+        super(with_blocking_container_by_id, self).cleanup()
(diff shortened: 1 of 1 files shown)


subtests/docker_cli/rmi/rmi.py
```diff
@@ -239,3 +239,9 @@ def postprocess(self):
     def check_image_exists_by_id(self, image_id):
         di = DockerImages(self)
         return di.list_imgs_with_image_id(image_id)
+
+    def cleanup(self):
+        # FIXME: remove once bz1317096 is fixed
```
cevich: Actually no, I think this is a genuine test bug by omission.  The  ``cleanup()`` method must always restore the system to it's prior state (or as near as possible/practical).  This test was failing that, the underlying bug is irrelevant IMHO, ``cleanup()`` should still do it's job.  This is partly the reason why ``garbage_check`` was added, there can be bugs in both docker and the test at the same time :grinning:

cevich:
Yeah, cool!  Kill that ``fixme`` and let's merge this!

PR #589, opened 2016-03-31
exit codes: fix docker-daemon corner case

edsantiago:
docker <1.10 does *not* exit 1 on all errors: on daemon errors,
it exits 2. Recognize that in mustfail() by converting 125
(new 1.10 exit code for daemon errors) to 2.

Also: fix one test case. "bad search" was used as an argument
to the --dns-search option. This would work if we ran docker
via exec, or quoting args, but when run via shell it becomes:

    docker ... --dns-search bad search --privileged ...

...which unsurprisingly results in the following error:

   Unable to find image 'search:latest' locally
   Trying to pull repository [...]/search ... not found

...which of course has exit code 1, not 2. (TL;DR this test case
was not actually testing what it intended to).


cevich:
Wow man, knockin' em down 1ik3 @ 8055!

PR #590, opened 2016-04-05
Yet more exit code cleanup

edsantiago:
Two instances -- kill and start -- which really seem like docker
daemon exit codes (i.e. docker <1.10 should exit 2) are exiting 1.
Don't understand it and am giving up on trying. We'll have to sort
this out when docker-1.10 comes out.

### DIFF
--- a/subtests/docker_cli/kill_bad/kill_bad.py
+++ b/subtests/docker_cli/kill_bad/kill_bad.py
@@ -71,14 +71,14 @@ class kill_bad_base(subtest.SubSubtest):
             mustfail(DockerCmd(self, 'kill',
                                ['-s', signal,
                                 self.sub_stuff['container_name']],
-                               verbose=False).execute(), 125)
+                               verbose=False).execute(), 1)
             self.failif(self.sub_stuff['container_cmd'].done, "Testing "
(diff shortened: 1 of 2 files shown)


cevich:
Hmmm, that's a PITA. Unf. also typical for our test-subject,  along with taking a year to fix the dang exit codes to begin with :angry: 

One idea might be to make the expected code a configuration parameter, that way it can be altered before execution-time based on the environment (docker version).  Just a thought.

edsantiago:
PITA indeed. I considered adding a config setting but, as you can understand, this has already gotten unpleasantly complex. Since I can't seem to predict (from the docker commit) what the exit codes will be like in 1.10, I think the only thing we can do is wait, see how docker-1.10 behaves, and then figure out a sane approach. Right now it's just guesswork, and unproductive guesswork at that.

cevich:
Uggg, I'm sorry, yep, I remember being tossed around by the whale-boat as well.  Probably half the bugs we originally opened were found while writing tests, not executing them :confused:  There are lots of these 'doing it this way 'caus dunno WTF else to do' type solutions kicking around in the project.  It's normal unfortunately.

PR #591, opened 2016-04-05
commit subtest: fix race condition

edsantiago:
Running docker with --detach, just to get a CID, exposes us to
a race condition if we check the container contents before the
command has completed.

Solution: use --name=<rand name> instead. That seems to be the
standard convention on other tests.

Also: clean up a diagnostic message: failif_ne() now displays
the expected and actual values, so caller doesn't have to


cevich:
Nice catch!  TOCTOU?  There are a few other tests that rely on CID and probably have the same race condition (the psa test comes to mind, but I may be mistaken).  A fix with more removals than additions makes it even better :grinning: Thanks!

PR #592, opened 2016-04-08
Docker 1.10

edsantiago:
This is pretty huge. It's impossible to review in one chunk, that's why I've taken great pains to git-rebase into manageable chunks. I'm not sure how github handles this as a PR, though. If you can review individual commits, I suggest doing so. If you can't, let me know and I'll work on submitting smaller individual PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ script:
     - ./run_checkdocs.py
     - ./run_unittests.sh
     - ./run_pylint.sh --FF
-    - inspekt style --disable E501,E265,W601,E402,E731
+    - inspekt style --disable E501,E265,W601,E402,E731,E221
 
 # TODO: Do we need notifications?
 #notifications:
--- a/dockertest/images.py
+++ b/dockertest/images.py
@@ -58,7 +58,7 @@ class DockerImage(object):  # pylint: disable=R0902
                               r"(?P<repo_addr_port>:[\d]+?)?/)?"  # optional
                               r"(?P<user>[\w\-\.\+]+/)?"  # optional
                               r"(?P<repo>[\w\-\.]+)"  # required
-                              r"(?P<tag>:[\w\-\.]+)?")  # optional
+                              r"(?P<tag>:[\w\-\.]*)?")  # optional
 
     # Many arguments are simply required here
     # pylint: disable=R0913
--- a/dockertest/images.py
+++ b/dockertest/images.py
@@ -92,6 +92,10 @@ class DockerImage(object):  # pylint: disable=R0902
         self.user = user
 
         self.short_id = long_id[:12]
+        # docker-1.10 output includes a prefix describing the hash type
+        colon = long_id.find(":")
+        if colon >= 3:
+            self.short_id = long_id[colon + 1:colon + 13]
         self.full_name = self.full_name_from_component(repo, tag,
                                                        repo_addr, user)
 
@@ -360,7 +364,13 @@ class DockerImages(object):
         tag = row['TAG']
         long_id = row['IMAGE ID']
         created = row['CREATED']
-        size = row['VIRTUAL SIZE']
+        try:
+            size = row['VIRTUAL SIZE']
+        except KeyError:
+            try:
+                size = row['SIZE']
+            except KeyError:
+                raise KeyError("neither SIZE nor VIRTUAL SIZE found in header")
         return cls.DICLS(repo, tag, long_id, created, size)
 
     # private methods don't need docstrings
--- a/dockertest/images_unittests.py
+++ b/dockertest/images_unittests.py
@@ -235,7 +235,7 @@ class DockerImageTestBasic(ImageTestBase):
                          " DockerImage(full_name:fedora:32 LONG_ID:0d20aec6529d5d396b195182c0eaa82bfe014c3e82ab390203ed56a774d2c404 CREATED:5 weeks ago SIZE:387 MB),"
                          " DockerImage(full_name:fedora:rawhide LONG_ID:0d20aec6529d5d396b195182c0eaa82bfe014c3e82ab390203ed56a774d2c404 CREATED:5 weeks ago SIZE:387 MB)]",)
 
-    def test_name_comparsion_dockerimage(self):
+    def test_name_comparison_dockerimage(self):
         di_ref = self.images.DockerImage("fedora_repo", "last_tag",
                                          ("0d20aec6529d5d396b195182c0eaa82bfe0"
                                           "14c3e82ab390203ed56a774d2c404"),
--- a/dockertest/images_unittests.py
+++ b/dockertest/images_unittests.py
@@ -355,6 +355,19 @@ class DockerImageTestBasic(ImageTestBase):
         self.assertEqual(user, None)
         self.assertEqual(addr, "address:1234")
 
+    def test_empty_tag(self):
+        """
+        Empty tag names seen in docker-1.9 output of 'docker events'
+        in some tag actions. docker-1.10 seems to forbid trailing colon.
+        """
+        test = "myname:"   # note the trailing colon with no actual tag name
+        DI = self.images.DockerImage
+        repo, tag, addr, user = DI.split_to_component(test)
+        self.assertEqual(repo, 'myname')
+        self.assertEqual(tag, '')
+        self.assertEqual(user, None)
+        self.assertEqual(addr, None)
+
     def test_remove_cli(self):
         d = self.images.DockerImages(self.fake_subtest, 1.0, True)
         self.assertEqual(d.remove_image_by_id('123456789012').command,
--- a/dockertest/images_unittests.py
+++ b/dockertest/images_unittests.py
@@ -316,6 +316,20 @@ class DockerImageTestBasic(ImageTestBase):
         act = self.images.DockerImage.full_name_from_defaults(config)
         self.assertEqual(act, 'fedora')
 
+    def test_docker_110_id_parsing(self):
+        """
+        Docker 1.10 image and container IDs include a "sha256:" prefix;
+        make sure our parsing code strips off the prefix but still returns
+        a valid short_id
+        """
+        id_components = ["sha256:", "0123456789ab",    # <-- 12-char short id
+                         "9dcfe46a8f465344284a3538cd"  # <-- + 52 chars = 64
+                         "6b0beeb18b4f249619fec4a659"]
+        long_id = "".join(id_components)
+        di = self.images.DockerImage("myrepo", "a", long_id, "bb", "cc", "dd")
+        self.assertEqual(di.long_id, long_id)
+        self.assertEqual(di.short_id, id_components[1])
+
     def test_bastard_repo(self):
         test = "bAsTaRd-rEPo.lOcAl_hOsT:1073741824/.b+o-F_H./a.s-d_f:F.d-S_a"
         DI = self.images.DockerImage
--- a/dockertest/output.py
+++ b/dockertest/output.py
@@ -798,29 +798,32 @@ def mustfail(cmdresult, expected_status=None, failmsg=None):
     """
     # FIXME: temporary: backward compatibility for pre-20160330 code
     # FIXME: remove before the next API-changing release
-    if expected_status is None:
+    if expected_status is None:                    # old: mustfail(x)
         expected_status = 1
-    if isinstance(expected_status, basestring):
-        failmsg = expected_status
-        expected_status = 1
+    if isinstance(expected_status, basestring):    # old: mustfail(x, "msg")
+        if not expected_status.isdigit():     # pylint: disable=E1101
+            failmsg = expected_status
+            expected_status = 1
 
     if failmsg is None:
         details = "%s" % cmdresult
     else:
         details = "%s\n%s" % (failmsg, cmdresult)
-    if cmdresult is not None:
-        OutputNotBad(cmdresult)
+    OutputNotBad(cmdresult)
+    if cmdresult.exit_status == expected_status:
+        return cmdresult
+    # On pre-1.10 docker, accept any nonzero exit status: it's impossible
+    # to automatically map docker-1.10 codes to 1.9
     # FIXME: temporary; remove once we no longer run on pre-1.10 docker
     if not DockerVersion().has_distinct_exit_codes:
-        # Docker <1.10 exits 2 on daemon error (125 in 1.10), 1 on all else
-        expected_status = 1 + (expected_status == 125)
-    if cmdresult.exit_status != expected_status:
-        raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
-                                        " Details: %s" % (
-                                            cmdresult.exit_status,
-                                            expected_status,
-                                            details))
-    return cmdresult
+        if cmdresult.exit_status != 0:
+            return cmdresult
+
+    raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
+                                    " Details: %s" % (
+                                        cmdresult.exit_status,
+                                        expected_status,
+                                        details))
 
 
 # This class inherits a LOT of public methods, but most of what
--- a/dockertest/output.py
+++ b/dockertest/output.py
@@ -929,6 +929,10 @@ class DockerTime(datetime.datetime):  # pylint: disable=R0903
         dargs = dict(zip(tuple(keys), tuple(values)))
         return super(DockerTime, cls).__new__(cls, **dargs)
 
+    def __repr__(self):
+        return '{0}("{1:%Y-%m-%dT%H:%M:%S}.{2:06d}{1:%z}")'.format(
+            self.__class__.__name__, self, self.microsecond)
+
     @classmethod
     def __new_zulu__(cls, isostr, base, values, keys, tzn):
         # Killall letter Z and z's
--- a/dockertest/output_unittests.py
+++ b/dockertest/output_unittests.py
@@ -268,6 +268,29 @@ fedora                        latest              58394af373423902a1b97f209a31e3
         sr = tt.find('TAG', 'rawhide')
         self.assertEqual(sr['REPOSITORY'], 'fedora')
 
+    def test_images_d110(self):
+        """
+        New format for docker-1.10: VIRTUAL SIZE heading is now just SIZE,
+        and container IDs include the 'sha256:' prefix.
+        """
+        tt = self.TT("""REPOSITORY                    TAG                 IMAGE ID                                                                  CREATED             SIZE
+192.168.122.245:5000/fedora   32                  sha256:0d20aec6529d5d396b195182c0eaa82bfe014c3e82ab390203ed56a774d2c404   5 weeks ago         387 MB
+fedora                        32                  sha256:0d20aec6529d5d396b195182c0eaa82bfe014c3e82ab390203ed56a774d2c404   5 weeks ago         387 MB
+fedora                        rawhide             sha256:0d20aec6529d5d396b195182c0eaa82bfe014c3e82ab390203ed56a774d2c404   5 weeks ago         387 MB
+192.168.122.245:5000/fedora   latest              sha256:58394af373423902a1b97f209a31e3777932d9321ef10e64feaaa7b4df609cf9   5 weeks ago         385.5 MB
+fedora                        20                  sha256:58394af373423902a1b97f209a31e3777932d9321ef10e64feaaa7b4df609cf9   5 weeks ago         385.5 MB
+fedora                        heisenbug           sha256:58394af373423902a1b97f209a31e3777932d9321ef10e64feaaa7b4df609cf9   5 weeks ago         385.5 MB
+fedora                        latest              sha256:58394af373423902a1b97f209a31e3777932d9321ef10e64feaaa7b4df609cf9   5 weeks ago         385.5 MB
+""")
+        self.assertEqual(tt.columnranges.values(),
+                         ['REPOSITORY', 'TAG', 'IMAGE ID', 'CREATED',
+                          'SIZE'])
+        sr = tt.search('IMAGE ID', ('sha256:58394af373423902a1b97f209a31e3777932'
+                                    'd9321ef10e64feaaa7b4df609cf9'))
+        self.assertEqual(len(sr), 4)
+        sr = tt.find('TAG', 'rawhide')
+        self.assertEqual(sr['REPOSITORY'], 'fedora')
+
     def test_containers(self):
         tt = self.TT("""CONTAINER ID                                                       IMAGE               COMMAND                                                                                                                   CREATED              STATUS                          PORTS               NAMES               SIZE
 96e1db5c0fd559d7ad4c472425c8b5f1f5b8c3b5952cc1fdc3eee846f6b33fea   fedora:20           bash -c 'echo 'this is a really really really really super big really really really really long command' > /tmp/foobar'   About a minute ago   Exited (0) About a minute ago                       dreamy_brattain     166 B
--- a/run_pylint.sh
+++ b/run_pylint.sh
@@ -24,11 +24,12 @@ fi
 
 TMPFILENAME="/tmp/run_pylint_$RANDOM"
 PEP8=`which pep8`
-PEP8IGNORE='E731'
+# for readability, allow multiple spaces after commas, colons & before '='
+PEP8IGNORE='E731,E221,E241'
 MSGFMT='(pylint) {msg_id}:{line:3d},{column}: {obj}: {msg}'
 # Disable 'line too long' - will be picked up by pep8
 # Check "note" (W0511) separately
-DISABLEMSG="I0011,R0801,R0904,R0921,R0922,C0301,W0511${SPECIALONE}"
+DISABLEMSG="I0011,R0801,R0904,R0921,R0922,C0301,C0326,W0511${SPECIALONE}"
 INIT_HOOK="
 AP = os.environ.get('AUTOTEST_PATH', '/usr/local/autotest')
 sys.path.append(os.path.abspath(AP + '/..'))
@@ -36,7 +37,7 @@ sys.path.append(os.path.abspath('.'))
 import autotest
 import autotest.common
 "
-SUBTESTDISABLEMSG="I0011,R0801,R0904,E1101,E1002,R0903,F0401,C0103,C0111,W0232,C0301,W0511,${SPECIALONE}"
+SUBTESTDISABLEMSG="I0011,R0801,R0904,E1101,E1002,R0903,F0401,C0103,C0326,C0111,W0232,C0301,W0511,${SPECIALONE}"
 SUBTESTINIT_HOOK="
 AP = os.environ.get('AUTOTEST_PATH', '/usr/local/autotest')
 sys.path.append(os.path.abspath(AP + '/..'))
--- a/run_unittests.sh
+++ b/run_unittests.sh
@@ -13,6 +13,8 @@ do
     sleep "0.1s"
     wait %- &> /dev/null
     RET=$?
+
+    if [ $RET -ne 0 -a $RET -ne 127 ]; then exit $RET; fi
 done
 
 if [ -n "$AUTOTEST_PATH" ]
@@ -32,7 +34,8 @@ fi
 # FIXME: find a way to have nosetests recurse, or move the tests into
 #        a directory structure that nose can handle.
 for unittest in $(find subtests -name 'test_*.py'); do
-    nosetests -v $unittest
+    echo $unittest
+    nosetests -v $unittest || exit 1
 done
 
 echo ""
--- a/subtests/docker_cli/build/build.py
+++ b/subtests/docker_cli/build/build.py
@@ -125,7 +125,7 @@ class postprocessing(object):
                 result = method(build_def, command, parameter)
                 passed.append(result)
                 if not result:
-                    self.loginfo("Postprocess %s: Failed", command)
+                    self.logwarning("Postprocess %s: Failed", command)
         if not all(passed):
             raise DockerTestFail("One or more postprocess commands did not"
                                  " pass, see debug log for details")
--- a/subtests/docker_cli/build/build.py
+++ b/subtests/docker_cli/build/build.py
@@ -128,7 +128,7 @@ class postprocessing(object):
                     self.loginfo("Postprocess %s: Failed", command)
         if not all(passed):
             raise DockerTestFail("One or more postprocess commands did not"
-                                 "pass, see debug log for details")
+                                 " pass, see debug log for details")
         else:
             self.loginfo("All postprocessing passed")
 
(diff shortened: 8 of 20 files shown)


dockertest/output.py
```diff
@@ -798,29 +798,32 @@ def mustfail(cmdresult, expected_status=None, failmsg=None):
...
+    if isinstance(expected_status, basestring):    # old: mustfail(x, "msg")
+        if not expected_status.isdigit():     # pylint: disable=E1101
+            failmsg = expected_status
+            expected_status = 1
 
     if failmsg is None:
         details = "%s" % cmdresult
     else:
         details = "%s\n%s" % (failmsg, cmdresult)
-    if cmdresult is not None:
```
edsantiago: Unnecessary, and removed: it is nonsensical to invoke us with cmdresult=None. Also, the very next lines assume it's !None anyway.

dockertest/images.py
```diff
@@ -54,11 +54,11 @@ class DockerImage(object):  # pylint: disable=R0902
...
-    repo_split_p = re.compile(r"(?P<repo_addr>.+?"  # optional
-                              r"(?P<repo_addr_port>:[\d]+?)?/)?"  # optional
-                              r"(?P<user>[\w\-\.\+]+/)?"  # optional
-                              r"(?P<repo>[\w\-\.]+)"  # required
-                              r"(?P<tag>:[\w\-\.]+)?")  # optional
+    repo_split_p = (r"(?P<repo_addr>.+?"  # optional
+                    r"(?P<repo_addr_port>:[\d]+?)?/)?"  # optional
+                    r"(?P<user>[\w\-\.\+]+/)?"  # optional
+                    r"(?P<repo>[\w\-\.]+)"  # required
+                    r"(?P<tag>:[\w\-\.]*)?")  # optional
```
edsantiago: The re.compile made it so I couldn't reuse this regex in events.py. If you know of a way to re.UNcompile, or a better way to do that code in events.py, I'm open to suggestion.

edsantiago: Note also that the '+' in the last (tag) line is now '*', because "foo/bar:" (empty tag) seems to be valid syntax.

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
...
+        source = "registry.access.redhat.com/rhel7/rhel:latest"
+
+        event_log = ("{t[0]} {cid}: (from {source}) create\n"
+                     "{t[1]} {cid}: (from {source}) start\n"
+                     "{t[2]} {cid}: (from {source}) die\n".format(**locals()))
+        keys = ['datetime', 'identifier', 'operation', 'source']
+        expect = [
+            (cid, dict(zip(keys, [DockerTime(t[0]), cid, 'create', source]))),
+            (cid, dict(zip(keys, [DockerTime(t[1]), cid, 'start',  source]))),
+            (cid, dict(zip(keys, [DockerTime(t[2]), cid, 'die',    source])))]
```
edsantiago: This (and a few others below) is what I mean by wanting to align fields in columns, requiring the pylint/pep8 disabling of only-one-space-after-comma-colon-equals.

edsantiago:
This is still incomplete: I'm running into a possible race condition in run_cidfile, and unknown failure in run_sigproxy.

I'm submitting now because I'll be on PTO starting Monday midday, back Friday AM 15-Apr, and I didn't want to wait for the last minute.

edsantiago:
Forgot to mention: tests mostly pass[1] on RHEL-7.2 with docker-1.9.1,-9.el7, 1.9.1-26.el7, and 1.10.3-5.el7.

 [1] exceptions on 1.9 are known issues; I'm looking into the 1.10 ones.

cevich:
> If you can review individual commits

I can, but as soon as you rebase/re-push to the PR, all/any commit-comments become hard to find.  I'll start with looking at the comments you added.

edsantiago:
Any suggestions for a better way?

.travis.yml
```diff
@@ -25,7 +25,7 @@ script:
     - ./run_checkdocs.py
     - ./run_unittests.sh
     - ./run_pylint.sh --FF
-    - inspekt style --disable E501,E265,W601,E402,E731
+    - inspekt style --disable E501,E265,W601,E402,E731,E221
```
cevich: E221: "multiple spaces before operator"

Why disable this for the entire project, is there some inconsistency in the way it's applied (or not)?

edsantiago: (I think you're reviewing the whole shebang, not commit-by-commit. That will make it impossible to follow the thread).

Because pep8 is all-or-nothing: there's no way to disable a warning in just one place using comments.

dockertest/output.py
cevich: There was a reason for adding this, but my memory of it is long since gone.  I'm fine with killing it, and if the other thing is still a problem, it'll re-present itself :smiley:

dockertest/output.py
```diff
@@ -798,29 +798,32 @@ def mustfail(cmdresult, expected_status=None, failmsg=None):
...
+            return cmdresult
+
+    raise xceptions.DockerExecError("Unexpected exit code %d; expected %d."
+                                    " Details: %s" % (
+                                        cmdresult.exit_status,
+                                        expected_status,
+                                        details))
 
 
 # This class inherits a LOT of public methods, but most of what
```
cevich: LGTM

subtests/docker_cli/build/build.py
```diff
@@ -316,7 +316,7 @@ def intermediate_exist_postprocess(self, build_def, command, parameter):
             if created_ids.isdisjoint(img_ids):
                 self.logdebug("%s() All missing as expected", command)
                 return True
-        self.logdebug("%s() Unaccounted imtermediates: %s",
+        self.logdebug("%s() Unaccounted intermediates: %s",
                       command, created_ids)
         self.logdebug("w/ leftover image IDs: %s", img_ids)
         return False
```
cevich: LGTM

subtests/docker_cli/build/build.py
```diff
@@ -125,7 +125,7 @@ def postprocess(self):
                 result = method(build_def, command, parameter)
                 passed.append(result)
                 if not result:
-                    self.loginfo("Postprocess %s: Failed", command)
+                    self.logwarning("Postprocess %s: Failed", command)
         if not all(passed):
             raise DockerTestFail("One or more postprocess commands did not"
                                  " pass, see debug log for details")
```
cevich: LGTM

run_unittests.sh
```diff
@@ -32,7 +34,8 @@ fi
 # FIXME: find a way to have nosetests recurse, or move the tests into
 #        a directory structure that nose can handle.
 for unittest in $(find subtests -name 'test_*.py'); do
-    nosetests -v $unittest
+    echo $unittest
+    nosetests -v $unittest || exit 1
```
cevich: Thinking out loud:  I've since discovered some important things for us to consider.  (1) nosetests is unmaintained / forked to nosetests2.  (2) We're already depending on unittest2 which has it's own test discovery/runner script that does basically the same thing.  Consider replacing ``nosetests ...`` with ``unit2 ...`` (I think most of the options are even the same).  This would let us drop a dependency that will be disappearing in the future anyway.

dockertest/images.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, repo, tag, long_id, created, size,
         self.user = user
 
         self.short_id = long_id[:12]
+        # docker-1.10 output includes a prefix describing the hash type
+        # FIXME: find canonical list
+        if long_id.startswith("sha256:"):
+            self.short_id = long_id[7:19]
```
cevich: Why not extend the parsing regex with an ``(\w+:)?`` then check the length of ``len(mobj.groups())`` to see if there's a need to extract the hash method.  It could be useful to know what method is used, since it's changed before (at least once), I'll bet they change it again sometime.

dockertest/output_unittests.py
```diff
@@ -268,6 +268,29 @@ def test_images(self):
...
+                          'SIZE'])
+        sr = tt.search('IMAGE ID', ('sha256:58394af373423902a1b97f209a31e3777932'
+                                    'd9321ef10e64feaaa7b4df609cf9'))
+        self.assertEqual(len(sr), 4)
+        sr = tt.find('TAG', 'rawhide')
+        self.assertEqual(sr['REPOSITORY'], 'fedora')
+
     def test_containers(self):
         tt = self.TT("""CONTAINER ID                                                       IMAGE               COMMAND                                                                                                                   CREATED              STATUS                          PORTS               NAMES               SIZE
 96e1db5c0fd559d7ad4c472425c8b5f1f5b8c3b5952cc1fdc3eee846f6b33fea   fedora:20           bash -c 'echo 'this is a really really really really super big really really really really long command' > /tmp/foobar'   About a minute ago   Exited (0) About a minute ago                       dreamy_brattain     166 B
```
cevich: Okay, the rest LGTM

dockertest/images.py
cevich: I don't quite follow, why can't events just reference it directly?  You can access the pattern in a compiled object by it's ``pattern`` attribute.  However, if the pattern needs changing to be used in events, then it's probably easier to just have a separate pattern that's more purpose specific (i.e. fewer groups or simpler).  I think that the regex in images is used so much, having a less precise pattern in the test is perfectly fine.

dockertest/images.py
cevich: I suppose we can probably get at the hash-type another way, like in the JSON as well.  Okay, that's probably fine unless you think there's value to my comment above.

subtests/docker_cli/events/test_event_parsing.py
cevich: OIC.  Yeah, better to do local disables then, just where it needs to be ignored.  I like to avoid globally disabling checks unless it really is absolutely necessary.

cevich:
> This is still incomplete:

Okay, noted.  No problem submitting early.

.travis.yml
cevich: Nono, doing it commit by commit AND the whole shebang at the same time :dancer: 

Grrr, stooped pep8.  I have half-a-mind to just nuke the damn thing.

cevich:
> Any suggestions for a better way?

Nope, this is fine.  I'll let you know when I'm done taking a peek, then push away.  The comments will still be there as long as your branch exists.  I think it's good-nuf for a early-review.

cevich:
> > Any suggestions for a better way?

BTW, what I'm doing is- click on the ``commits`` tab, click the first commit, then use the "prev" | "next" buttons at the top to navigate the diffs.

run_unittests.sh
```diff
@@ -32,7 +34,8 @@ fi
...
 done
 
 echo ""
```
cevich: LGTM

subtests/docker_cli/load/load.py
```diff
@@ -26,7 +28,7 @@ class load(subtest.Subtest):
     def initialize(self):
         super(load, self).initialize()
         self.stuff['di'] = di = DockerImages(self)
-        img_id = self.config['test_id'].strip()
+        img_id = self._test_id
```
cevich: Shouldn't this be ``self.stuff['test_id']``?

subtests/docker_cli/load/load.py
```diff
@@ -49,7 +61,7 @@ def postprocess(self):
         super(load, self).postprocess()
         OutputGood(self.stuff['dkrcmd'].cmdresult)
         mustpass(self.stuff['dkrcmd'])
-        img_id = self.config['test_id']
+        img_id = self._test_id
```
cevich: Shouldn't this be self.stuff['test_id']?

dockertest/images_unittests.py
```diff
@@ -341,6 +341,15 @@ def test_addr_no_user(self):
         self.assertEqual(user, None)
         self.assertEqual(addr, "address:1234")
 
+    def test_empty_tag(self):
+        test = "myname:"   # note the trailing colon with no actual tag name
```
cevich: Oh man.  There are about a billion places that assume passing no tag name implies a ``latest`` will show up in the images list.  I predict the adockerclypse is nye.  It may be worthwhile to investigate and see if there are some bugs to file on this.  For one, it breaks backward compatibility with any tooling that relies on the old behavior.  I'm guessing it may also cause ambiguities WRT specifying an empty tag vs no tag.

dockertest/output.py
```diff
@@ -929,6 +929,10 @@ def __new__(cls, isostr, sep=None):
         dargs = dict(zip(tuple(keys), tuple(values)))
         return super(DockerTime, cls).__new__(cls, **dargs)
 
+    def __repr__(self):
+        return '{0}("{1:%Y-%m-%dT%H:%M:%S}.{2:06d}{1:%z}")'.format(
+            self.__class__.__name__, self, self.microsecond)
```
cevich: I should start getting use this //new// way, ugg "old-habits...".

subtests/docker_cli/events/events.py
```diff
@@ -1,3 +1,5 @@
+#
+# coding: utf-8
```
cevich: This might be better done as a mass-change all at once and tucked into a single commit by itself, just in case.

subtests/docker_cli/events/events.py
```diff
@@ -34,58 +36,13 @@
...
-
-
-def event_details(line):
-    return {'datetime': event_dt(line),
-            'source': event_source(line),
-            'operation': event_operation(line)}
+regexes = {
+    'timestamp': r'[\d-]+T[\d:]+\.\d+[+-][\d:]+',   # <iso8601>.<µs>[+/-]HH:MM
+    'cid':       r'(sha256:)?[0-9a-fA-F]{64}',      # 64-char hash
+    'fqin':      DockerImage.repo_split_p,          # eg some.repo/image:tag
```
cevich: Okay, I C, ya, just use ``DockerImage.repo_split_p.patern``  that will "un-compile" it for you :smile:

subtests/docker_cli/events/events.py
```diff
@@ -98,42 +55,88 @@ def is_dupe_event(needle, haystack):
...
+        'datetime':   DockerTime(mobj.group('timestamp')),
+        'identifier': mobj.group('identifier'),
+        'object':     mobj.group('object'),
+        'operation':  mobj.group('operation'),
+        'source':     None,
+    }
+    # TODO: (maybe): split out components of the parenthesized list.
+    # If so, keep in mind that you can't just split on commas (because
+    # of "Red Hat, Inc.") and that the fields are output in unpredictable
+    # order: even two consecutive event lines will have different ordering.
```
cevich: Sounds like a bug to me, is it too much to ask for an event-stream that's easily machine-readable?  It's much easier for docker to escape any commas than it is to pull them out on the parsing side.  Don't be afraid to open "design" bugs like this, worst case they get closed but in either case we have to deal with the problem.

subtests/docker_cli/events/events.py
```diff
@@ -98,42 +55,88 @@ def is_dupe_event(needle, haystack):
...
+    if mobj is not None:
+        return {
+            'datetime':   DockerTime(mobj.group('timestamp')),
+            'identifier': mobj.group('identifier'),
+            'source':     mobj.group('source'),
+            'operation':  mobj.group('operation'),
+        }
+    return None
+
+
```
cevich: At one point I was considering taking a lot of this code and putting it into a ``dockertest`` module for general use.  That's still a possibility if it makes any of this easier.  Only reason I never did is because I have yet to come across a second-test that would benefit from the events stream.  Though back in the day, the stream wasn't very useful, maybe it is more so today.

subtests/docker_cli/events/events.py
```diff
@@ -98,42 +55,88 @@ def is_dupe_event(needle, haystack):
...
 def parse_event(line):
     """
-    Return tuple(CID/FQIN, {DETAILS}) from parsing line
+    Return {DETAILS} from parsing line
 
     :param line: String-like containing a single event line
-    :returns: (CID/FQIN, {DETAILS}) from parsing line or None if unparseable
+    :returns: {DETAILS} from parsing line or None if unparseable
     """
```
cevich: Those curly braces won't mess with ``str.format()`` will they?  I suppose probably not.

subtests/docker_cli/events/events.py
```diff
@@ -98,42 +55,88 @@ def is_dupe_event(needle, haystack):
...
         n_lines += 1
-        if len(line) < 64:   # len of a cid
-            sloppy.append(line)
-            continue
         cid_details = parse_event(line)
         if cid_details is not None:
-            result.append(cid_details)
+            result.append((cid_details['identifier'], cid_details))
         else:
             sloppy.append(line)
```
cevich: If you like, it might be okay to get rid of the sloppy support.  There was a time ages ago when it was actually necessary to get a passing test :scream:

subtests/docker_cli/events/events.py
```diff
@@ -239,7 +242,8 @@ def postprocess(self):
         cid_events = events_by_id(all_events)
         cid = self.stuff['nfdc_cid']
         self.failif(cid not in cid_events,
-                    'Test container cid %s does not appear in events' % cid)
+                    'Test container cid %s does not appear in events %s'
+                    % (cid, cid_events))
```
cevich: ahh good catch

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
+# -*- python -*-
+#
```
cevich: Here too, might be good to do these as a project-wide mass change.

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
...
+        actions = ['pull', 'tag', 'tag']
+        event_log = ("{t[0]} {fqins[0]}: {actions[0]}\n"
+                     "{t[1]} {fqins[1]}: {actions[1]}\n"
+                     "{t[2]} {fqins[2]}: {actions[2]}\n".format(**locals()))
+        expect = []
+        for i in range(3):
+            expect.append((fqins[i], {'datetime':   DockerTime(t[i]),
+                                      'identifier': fqins[i],
+                                      'operation':  actions[i],
+                                      'source':     None}))
```
cevich: Yeah, you're right, it's much more readable this way.  Global disable for pep is fine then.

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
+# -*- python -*-
```
cevich: Wow man, new test-unittests too! You've outdone yourself.  Impressive.

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
...
+                     " Version=7.2, name=events_c36O, Architecture=x86_64,"
+                     " Authoritative_Registry=registry.access.redhat.com,"
+                     " Build_Host=somewhere.redhat.com,"
+                     " Name=rhel7/rhel)"
+                     "\n"
+                     "{t[4]} network disconnect {nid}"
+                     " (name=bridge, type=bridge, container={cid})\n"
+                     "{t[5]} container archive-path {cid}"
+                     " (ignore this)\n"
+                     "".format(**locals()))
```
cevich: Giant in-line text like this is sometimes easier to manage/maintain if you stick it into a separate file (and read it in).  Or if it should be in-line, stick it into a ``"""`` type block and separate the definition from the operation.  e.g.

```python
                ...
                foobar="""big long
...
ugly
...
text
...
block"""

                foobar = foobar.format(...)
                ...
```

subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -0,0 +1,167 @@
...
+        for i in range(4):
+            expect[i][1]['datetime']   = DockerTime(t[i])
+            expect[i][1]['identifier'] = expect[i][0]
+            expect[i][1]['operation']  = actions[i]
+            expect[i][1]['object']     = 'image'
+            expect[i][1]['source']     = None
+
+        actual = events.parse_events(event_log)
+        self.maxDiff = None
+        self.assertEqual(actual, expect)
```
cevich: LGTM

run_pylint.sh
```diff
@@ -24,19 +24,20 @@ fi
...
+DISABLEMSG="I0011,R0801,R0904,R0921,R0922,C0301,C0326,W0511${SPECIALONE}"
 INIT_HOOK="
 AP = os.environ.get('AUTOTEST_PATH', '/usr/local/autotest')
 sys.path.append(os.path.abspath(AP + '/..'))
 sys.path.append(os.path.abspath('.'))
 import autotest
 import autotest.common
 "
-SUBTESTDISABLEMSG="I0011,R0801,R0904,E1101,E1002,R0903,F0401,C0103,C0111,W0232,C0301,W0511,${SPECIALONE}"
+SUBTESTDISABLEMSG="I0011,R0801,R0904,E1101,E1002,R0903,F0401,C0103,C0326,C0111,W0232,C0301,W0511,${SPECIALONE}"
```
cevich: The C0326's are pylint messages, no?  Doing those inline while globally ignoring the one from pep8 seems nonsensical.  I have three-quarters of a mind to say let's just kill pep8 entirely.

subtests/docker_cli/negativeusage/negativeusage.py
```diff
@@ -180,8 +180,7 @@ def postprocess(self):
         NoPanic(self.sub_stuff['cmdresult'])
         expected_exit_status = self.config['extcmd']
         cmdresult = self.sub_stuff['cmdresult']
-        self.failif_ne(cmdresult.exit_status, expected_exit_status,
-                       "Exit status: %s" % cmdresult)
+        mustfail(cmdresult, expected_exit_status)
         # Same checks for both
         for outtype in ('stdout', 'stderr'):
             if self.sub_stuff[outtype] is not None:
```
cevich: ahh ok.  LGTM.

subtests/docker_cli/import_url/import_url.py
```diff
@@ -48,7 +48,7 @@ def run_image(self):
         subargs += [self.sub_stuff['import_repo']]
         subargs += ['some', 'dummy', 'command']
         self.logdebug("Next docker command should fail...")
-        mustfail(DockerCmd(self, 'run', subargs).execute(), 1)
+        mustfail(DockerCmd(self, 'run', subargs).execute(), 125)
```
cevich: Watch, they'll all change from 125 to 7 in six months :smile:

subtests/docker_cli/run_cidfile/run_cidfile.py
```diff
@@ -135,14 +135,14 @@ def run_once(self):
...
-                                               lambda x: mustfail(x, 1)))
+                                               lambda x: mustfail(x, 125)))
         self._check_failure_cidfile_present(containers[-1])
         # cidfile already exists (exited container)
         containers[0].stdin("exit\n")
         containers[0].wait(10)
         containers[0].close()
         containers.append(self._init_container(subargs, cidfile, 'true',
-                                               lambda x: mustfail(x, 1)))
+                                               lambda x: mustfail(x, 125)))
```
cevich: Are we there yet?

subtests/docker_cli/run_attach/run_attach.py
```diff
@@ -53,7 +53,7 @@ def _init_container(self, tty, cmd, cmd_input='\n'):
         fin = DockerImage.full_name_from_defaults(self.config)
         subargs.append(fin)
         subargs.append(cmd)
-        cmd = DockerCmd(self, 'run', subargs, timeout=3, verbose=False)
+        cmd = DockerCmd(self, 'run', subargs, timeout=30, verbose=False)
```
cevich: I agree, never liked those short timeouts.  They lead to quite a few "discussions" in the past.

subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
```diff
@@ -117,7 +117,10 @@ def _read_cid(self):
...
         in format string and do an exact match on expected values.
         """
-        cid = open(self.sub_stuff['cidfile'], 'rb').read().strip()
+        try:
+            cid = open(self.sub_stuff['cidfile'], 'rb').read().strip()
+        except IOError:
+            return "<n/a>"         # no cidfile, e.g. because of bad command
         if len(cid) < 12:
             raise DockerTestError("bad cid (length < 12) in --cidfile")
         return cid
```
cevich: LGTM.  Worth adding a config./test to negativeusage for the invalid ``--cgroup-parent``?

subtests/docker_cli/tag/tag.py
```diff
@@ -20,6 +20,7 @@
 from dockertest.images import DockerImage
```
cevich: They always could have multiple tags, not sure I understand the commit message, but TBH I haven't played with 1.10 at all yet, so it's probably fine.

subtests/docker_cli/tag/tag.py
```diff
@@ -186,8 +187,12 @@ def initialize(self):
...
-        self.expect_pass(False)
+        # On docker 1.10, the second tag should pass. On < 1.10, fail.
+        try:
+            DockerVersion().require_server("1.10")
+            self.expect_pass(True)
+        except xceptions.DockerTestNAError:
+            self.expect_pass(False)
 
 
 class double_tag_force(double_tag):
```
cevich: Okay, LGTM (just ignore my misunderstanding, that's not a code problem)

cevich:
Whew! I made it!  This is quite a magnum-opus It's very much appreciated and you deserve a vacation or three :grinning:   I can't wait to merge it!

subtests/docker_cli/events/events.py
edsantiago: Filed: bz1328242

subtests/docker_cli/events/events.py
edsantiago: That could be useful for a general-purpose docker management library; I'm not sure how much need there is for it within docker-autotest. I'm inclined to defer that refactoring for an if-and-when-its-needed day...

subtests/docker_cli/events/events.py
edsantiago: Fixed, thanks! (not pushed yet)

subtests/docker_cli/events/test_event_parsing.py
edsantiago: A separate file wouldn't work well with the {cid} {etc} refactoring. Triple-quotes don't work because the lines are too long, triggering E501. But I've made a commit (unpushed) that cleans this up somewhat.

subtests/docker_cli/load/load.py
edsantiago: No, see the new _test_id() below. This is a temporary coexistence hack for 1.9 and 1.10. I've tried to come up with a better way to do this, but can't think of anything.

dockertest/images.py
edsantiago: >Why not extend the parsing regex with an (\w+:)? 

That puts the burden of understanding container IDs on the caller. I think it makes more sense for the caller to deal with cid as a black box. But I hate the hardcoded sha256 so I've checked in (unpushed) a different way to do it.

dockertest/images_unittests.py
edsantiago: The context here is support for docker-1.9 events output. In particular, the 1.9 output format is:

    <timestamp> <identifier>: <action>

...where <identifier> can be a SHA or a symbolic name. The symbolic name is usually of the form foo/bar:tag or even bar:tag but in some cases tag is null:

    2016-04-11T12:17:25.671704251-04:00 change_repository_bqtg:: tag
    2016-04-11T12:17:33.963279935-04:00 double_tag_ndal:: tag

(note the double colons: one of those is part of the docker-events output format, the other one is an empty tag). And, FWIW, docker-1.9 allows specifying an image name with a trailing colon (docker-1.10 does not).

Should I add a FIXME comment to the + -> * change, indicating that it should be reverted back to + once docker-1.10 is the standard?

subtests/docker_cli/events/events.py
cevich: Ya, that was my thinking as well, and why it's all stayed in that test for over a year. :smile:

subtests/docker_cli/load/load.py
cevich: oh it's a property, missed that.  Still don't quite follow, but I trust you, it's needed

dockertest/images.py
cevich: Ugg, what a PITA.  Knowing Docker, it'll probably change again in 6-mo.  It's almost as if we need a separate class just to handle the stupid ID.

dockertest/images_unittests.py
cevich: I get the impression that the bigger picture is to keep the lifespan of 1.9 as short as possible.  It's a franken-patch-beast IMHO.

CI Runs that generally pass, are more useful than CI runs with 100% coverage on 100% platforms/versions.  I think the best long-term strategy here is to spend as little time/effort as possible fussing with 1.9.  Any bugs uncovered would have fixes stuck into 1.10+ anyway.

It's okay by me to just duct-tape workaround / TestNA's all the 1.9 specialities, and concentrate on making it right for 1.10.

Does that make sense?

Remember, disabling/enabling tests and altering test configuration for platform-specific requirements is really easy to do in CI.  I have all the test configs in jinja2 templates, so adding conditionals there is almost a no-brainer.

edsantiago: I'm torn. Ditching 1.9 support is quite likely the right thing to do. But I'm not seeing much big push for 1.10, and am starting to get a bad feeling about this. The kind where it's easier to keep dual support in for a while than to graft it on three months from now when someone says "we need to support 1.9 for a little longer". BTDT.

cevich:
> I'm torn. Ditching 1.9 support

This reminds me of the time Docker changed the requirements for using the ``run -t`` option.  It was used in probably better than 75% of the tests here.  I spent a month changing and debugging everything and getting it all working again.  The very next version, they partially reverted the change - so all my work was basically pointless.

I have had many a night where I've dreamed of some kind of magic inbetween/decoupling module that could abstract away most of the version-specific behaviour.  Unf. when I consider also having to write the unittests for it, the prospect work quickly becomes much larger than just piling the duct-tape :confounded: 

Since it looks likely we'll always be in a multi-version world, across all the platforms we test on...well...I've ordered a case of duct-tape and some hand lotion (that adhesive is terrible on your skin, some people use it to remove warts!) :smile:

There are some utility classes/functions buried in autotest that may be of use (version-dependent classes comes to mind).  So do keep an open-mind for any sparkle of inspiration, on how this stuff could be made easier to maintain (writing it is always going to be hard).

edsantiago:
I keep thinking about it (if-then hell vs multibranch vs fork-copies). I think multibranch makes for cleanest code but is just too hard to maintain over time. I think if-then spaghetti is the least awful compromise.

If you agree, may I merge this? I'd like to do the import_url cleanup as a separate PR.

edsantiago:
@cevich ping, any objection to merging this? I have more fixes upcoming but I'd rather submit them separately

cevich:
Pong.  Sorry, you're earlier updated fell in a crack among many other updates, my bad.  Yes, if there's nothing else to do here, I think it's probably as good as it's going to get, and I'm sure you're sick of tinkering by now.  Any problems can be fixed as found.  Yep, go ahead.

PR #593, opened 2016-04-22
Update quickstart documentation and Prerequisites

cevich:
The quickstart formerly recommended resetting back to the tag
of the current version.  However, this means any between-version
fixes are skipped.  Esp. between 0.8.4 and HEAD, in combination with
changes in docker's output will cause the version test to fail.
Because the latest master HEAD is more likely to be bug-free, it's a
much better starting-off point for anyone following the quickstart.

Also updated the prerequisites section to include the 'Atomic' platforms
as well as make the runtime vs devel. software dependencies more distinct.
Docker Autotest is more than happy to run as-is on Atomic, but extra
bits are helpful for documentation and/or development work.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -72,36 +72,40 @@ Prerequisites
     *  **Clean** environment, only test-related images, containers
        and dependent services (besides ``docker -d``) at the start
        of **every** Autotest run.
-    *  Docker installation (with ``docker -d`` running at startup)
+    *  Docker installation (with docker daemon running)
     *  Default settings for Docker on your platform/OS,
-       unless otherwise noted.
+       including storage (LVM thin-pool + XFS).
     *  Sufficient available temporary storage for multiple copies of the
        test-related images/container content.
 
 *  `Supported Docker OS platform`_
 
-    *  Red Hat Enterprise Linux 7 Server
-    *  Fedora 20 or later recommended
-    *  Linux kernel 3.12.9 or later
+    *  Red Hat Enterprise Linux 7 Server (preferred for development)
+    *  Red Hat Enterprise Linux Atomic Host (preferred for testing)
+    *  Fedora 22 or later including Atomic (not all tests may function)
 
 *  Platform Applications/tools
 
+    *  Autotest 0.16.0 or later (preferred).
     *  Coreutils or equivalent (i.e. ``cat``, ``mkdir``, ``tee``, etc.)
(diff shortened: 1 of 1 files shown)


index.rst
```diff
@@ -145,7 +146,7 @@ Quickstart
 ::
 
     [root@docker tests]# cd docker
-    [root@docker docker]# git checkout $(git tag --list | tail -1)
+    [root@docker tests]# git reset origin/master
```
edsantiago: This is right on the heels of a git clone. Why not just remove this line entirely?

index.rst
```diff
@@ -72,30 +72,31 @@ Prerequisites
...
     *  Sufficient available temporary storage for multiple copies of the
        test-related images/container content.
 
 *  `Supported Docker OS platform`_
 
     *  Red Hat Enterprise Linux 7 Server
-    *  Fedora 20 or later recommended
-    *  Linux kernel 3.12.9 or later
+    *  Red Hat Enterprise Linux Atomic Host
+    *  Fedora 22 or later including Atomic
```
edsantiago: Last time I tried (February?), docker-autotest just didn't work on Fedora. Has that changed?

index.rst
```diff
@@ -72,30 +72,31 @@ Prerequisites
     *  **Clean** environment, only test-related images, containers
        and dependent services (besides ``docker -d``) at the start
        of **every** Autotest run.
-    *  Docker installation (with ``docker -d`` running at startup)
+    *  Docker installation (with docker daemon running)
     *  Default settings for Docker on your platform/OS,
-       unless otherwise noted.
+       including storage (LVM thin-pool + XFS).
     *  Sufficient available temporary storage for multiple copies of the
        test-related images/container content.
```
edsantiago: Sigh, might be worth adding "run as root" as prerequisite, including an explanation of why (I can't remember the full reasons, but I do remember spending a lot of time trying to get permissions right but still not having any luck).

index.rst
cevich: I half considered that...okay, it's fully considered now :smile:

index.rst
cevich: Works...doesn't mean all the tests pass, but ya, it should at least run.

index.rst
cevich: Oh good catch, ya, I agree.

PR #594, opened 2016-04-25
run_sigproxy: better diagnostics for timeout

edsantiago:
Every so often, run_sigproxy/default fails with:

    Signal was not raised in container even though sig-proxy was enabled

Actual reason for this seems to be that the container just
isn't ready. Not sure why, or why only sometimes.

This commit tries to make that a little clearer by waiting
for a READY indicator. If we time out waiting for it, the
error message is a little more clear.


subtests/docker_cli/run_sigproxy/run_sigproxy.py
```diff
@@ -81,6 +81,20 @@ def init_test_specific_variables(self):
         raise NotImplementedError("Test specific variables has to be "
                                   "overridden by child.")
 
+    def _wait_for_ready(self):
```
cevich: Could this method be replaced by a call to ``output.wait_for_output()``?

(I know sometimes it doesn't always fit well)

subtests/docker_cli/run_sigproxy/run_sigproxy.py
```diff
@@ -98,7 +112,7 @@ def initialize(self):
             self._init_container_attached(name)
         else:
             self._init_container_normal(name)
-        time.sleep(self.config['wait_start'])
+        self._wait_for_ready()
```
cevich:
```python
    container_cmd = self.sub_stuff['container_cmd']
    self.failif(not output.wait_for_output(lambda: getattr(container_cmd, 'stdout'),
                                           r'READY', self.config['wait_start'], timestep=0.5),
                                           'some message')
```

cevich:
Excellent concept / workaround for a legit race-condition.  Trap is not setup w/in container by the time signal is sent/received - IIRC, there is no guarantee of timing or delivery order (but my kernel-foo is much weakened post weekend involving beer and welding).

cevich:
LGTM w/ or w/o my proposed use of ``wait_for_output()``.  Working is more important than "pretty".

subtests/docker_cli/run_sigproxy/run_sigproxy.py
edsantiago: I like that (I didn't know about wait_for_output), thanks! But can we compromise? Keep _wait_for_ready() because it keeps initialize() smaller and more readable? Even though it's a single-use method, I'm a fan of small functions for compartmentalizing head-scratchy stuff.

edsantiago: Sorry, I wasn't clear: I meant, rewrite _wait_for_ready() so it uses the much nicer wait_for_output()?

cevich: Hmmm, actually in that case, what about adding it to the ``output`` module?  No unittest required - ``wait_for_output()`` is already tested, ``wait_for_ready()``  would just be wrapping it.  There are more than a handful of tests that could use this method.

Even if not, yep, your counter-proposal is perfectly fine as well.

edsantiago: >what about adding it to the output module?

Good idea, so done, tested, and then:

   (pylint) C0302:  1,0: : Too many lines in module (1020/1000)

*whimper*

cevich:
Let's move this along as-is, can always improve things w/ other commits.  No reason to hold up a fix for module/package juggling.

edsantiago:
Aw, *man*. I was actually having fun with this (in between bouts of firefighting on docker-latest)

cevich:
Sorry :(  It occurred to me the business of testing > pretty and perfect code.  I'm fine with improving the API, let's just not tie those changes to a test / bugfix PR.  I also understand sometimes that's not possible.  For this test especially, it's a sensitive subject with many non-obvious side-effects.

PR #595, opened 2016-04-25
docker run --cgroup-parent: error message change

edsantiago:
The error message for --cgroup-parent=/xxxxx (invalid name)
changed between docker-1.9 and 1.10

### DIFF
--- a/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
+++ b/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
@@ -161,9 +161,10 @@ class run_cgroup_parent_invalid_name(run_cgroup_parent_base):
     def initialize(self):
         super(run_cgroup_parent_invalid_name, self).initialize()
         self._setup("/{rand1}")
-        self._expect(stderr="Error response from daemon:"
-                     " Cannot start container {cid}: [8] System error:"
-                     " Invalid slice name /{rand1}")
+        self._expect(stderr="docker: Error response from daemon:"
+                     " cgroup-parent for systemd cgroup should be a"
+                     " valid slice named as \"xxx.slice\".\n"
+                     "See '/usr/bin/docker run --help'.")
 
 
 class run_cgroup_parent_path(run_cgroup_parent_base):
--- a/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
+++ b/subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
@@ -40,7 +40,7 @@ class run_cgroup_parent_base(SubSubtest):
         self.sub_stuff['rand2'] = utils.generate_random_string(8)
         self.sub_stuff['cgroup_parent'] = cg_parent.format(**self.sub_stuff)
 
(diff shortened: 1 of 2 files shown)


subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
```diff
@@ -161,9 +161,10 @@ class run_cgroup_parent_invalid_name(run_cgroup_parent_base):
     def initialize(self):
         super(run_cgroup_parent_invalid_name, self).initialize()
         self._setup("/{rand1}")
-        self._expect(stderr="Error response from daemon:"
-                     " Cannot start container {cid}: [8] System error:"
-                     " Invalid slice name /{rand1}")
+        self._expect(stderr="docker: Error response from daemon:"
+                     " cgroup-parent for systemd cgroup should be a"
+                     " valid slice named as \"xxx.slice\".\n"
+                     "See '/usr/bin/docker run --help'.")
```
cevich: Unf. this happens way more than would make our lives easy.  Consider making the message token into a configuration option value.  Also consider making it a less precise regex,  e.g. ``r'rror response from daemon.+slice name'``.  

Reason:  Configuration changes are **_easy_** to do in CI to suit the situation at hand.  Anything dealing with code/commit/version changes are much more difficult/problematic in CI (possibly affecting other tests as well).  Also, slightly Less precise because it reduces the maintenance burden to a degree disproportional to the risk of false-positives (i.e. not matching when it should have).

cevich:
Fix LGTM, one suggestion, take/leave it, I'm fine w/ the code either way (don't think about it too long) :smile:

subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
edsantiago: Yeah, I hate this too. It's fragile. Let me look into having self._expect() handle regex instead of exact string.

edsantiago: Sigh..... speak of the devil. While firefighting more docker-latest issues: [docker da38ac6c](https://github.com/docker/docker/commit/da38ac6c79fe902ed0687afc73d731c95c6d491a#diff-29081af54e3d2dea9c82ab2e757e61bf) changes "Uid" to "uid" and "Unable" to "unable", leading to failure in run_user test. This is frustrating.

cevich: Lol, oh just you wait and see how frequent this is :scream: Sometimes they change the order of words too.  I've also had to open bugs because they dropped the word 'error' from error messages :confounded:

edsantiago:
On an unrelated thread, is this the kind of separation you were looking for?

cevich:
I think that'll do quite nicely.  Now, if "they" try and stick-it-to-us by changing "daemon" to "Daemon", we'll be ready wa-ha-ha-ha-ha-ha :D  Nice work, mine would had hammer marks and miss-drilled holes all over it :laughing:

PR #596, opened 2016-04-25
import_url: inconsistent exit codes in docker-1.10

edsantiago:
The import_url test runs a docker command that's expected to fail:

    docker run Not-A-Valid_Image some dummy command

docker-1.9 consistently fails with "some" not in $PATH

docker-1.10 fails in three (AFAICT) different ways:

    exec: "some": executable file not found in $PATH

Then one of these three errors:

    docker: Error response from daemon: Cannot start container <cid>: [9] System error: invalid character 'e' looking for beginning of value.
    docker: Error response from daemon: Cannot start container <cid>: [9] System error: json: cannot unmarshal string into Go value of type libcontainer.generic Error.
    docker: Error response from daemon: Container command not found or does not exist..

...and an exit code of 125 (first two) or 127 (last). This makes it
impossible to test with mustfail().

Workaround: do our own test. Allow exit code 125 or 127.

### DIFF
--- a/subtests/docker_cli/import_url/import_url.py
+++ b/subtests/docker_cli/import_url/import_url.py
@@ -26,7 +26,6 @@ from dockertest.containers import DockerContainers
 from dockertest.images import DockerImages
 from dockertest.dockercmd import DockerCmd
 from dockertest.output import mustpass
-from dockertest.output import mustfail
 
 
 class import_url(SubSubtestCaller):
@@ -48,7 +47,10 @@ class base(SubSubtest):
         subargs += [self.sub_stuff['import_repo']]
         subargs += ['some', 'dummy', 'command']
         self.logdebug("Next docker command should fail...")
-        mustfail(DockerCmd(self, 'run', subargs).execute(), 125)
+        cmdresult = DockerCmd(self, 'run', subargs).execute()
+        if cmdresult.exit_status not in [125, 127]:
+            self.failif(True, "Unexpected exit code %d; expected 125 or 127."
+                        % cmdresult.exit_status)
 
     def initialize(self):
         super(base, self).initialize()


cevich:
In the past, we (docker autotest schmucks) have opened a LOT of bugs with both 'docker' and 'inconsistent' in the subject.  There's no reason (I know of) why docker can't / shouldn't behave consistently in error/exit codes in this situation.  It not only makes our lives harder, but any other tooling hooked into the same code-paths will also cause engineer-grief.  It would be nice to get a 'fix' for this in 1.10. I think we still need your changes here even if as a workaround, until bugfix.  Changes LGTM.

edsantiago:
Getting back to this... the imported "container" isn't actually valid. This is the sort of thing I'm happy to leave as undefined behavior. Is this really worth filing an issue?

cevich:
Err, woops, does it always say ``exec: "some": executable file not found in $PATH``?  If so, then it's fine.  I read your note  above thinking it was returning three completely different errors.  Yes, that's fine, as long as there is at least one thing that's consistent and we can check for, just ignore the other crap.

edsantiago:
Yes, it always says `executable file not found in $PATH` but then follows that with another 1D3 choice and inconsistent exit code. And, um, the code doesn't actually check stderr but that seems to be covered in negativeusage.py

cevich:
The exit-code consistency is a must (unless it's otherwise meaningful in some way we don't know).  That damn PR too a year to merge, it better be frickin' done right.  If you can't discover a reason, bug that (and let the test fail).

edsantiago:
OK, I'll file a bz. I fully expect it to be closed as WONTFIX-GIGO.

edsantiago:
rhbz#1331575

cevich:
Could very well be, but at least we're trying to make the world a better place :smile:

edsantiago:
So... merge this now, or wait for ack on bz? If the latter, how long before we assume it's in limbo?

cevich:
Hmmm, there are basically three paths I've taken in the past:

1) Setup the test for a single error-code and let it fail, but tag the test in the whiteboard so docker autotest knows to skip it.  The magic string for the BZ "whiteboard" in this case would be: ``docker-autotest:0.8.4:docker_cli/import_url/md5sum``.

2) Setup the test to only check for non-zero exits, then forget about the problem and pretend it never existed.

3) Obfuscate the code with spelling mistakes to the degree that it always passes no matter what :grinning: 

Or I suppose talking to the team and getting their input is a valid course as well.  I'm leaning toward option 1, but have no problem with option 2 - the test can always be "improved" later, maybe add a TODO or FIXME to it.

cevich:
This is a tricky one.  I'm conflicted WRT testing multiple versions.  I think since it's mostly 'behaving' on 1.9 (frankenpatch but "stable") merging is the right thing.  1.10 is much more bleeding edge, and this test failing can simply be expected - there's an open bug.  I don't think we should mark the bug's whiteboard though, because we want the test to run in 1.9

PR #598, opened 2016-04-26
run_pylint: predictable execution order

edsantiago:
Apropos of nothing, sort file list before running tests. It
just helps greatly with readability and especially when
looking for results on a recently-touched file.


cevich:
Yep, good, LGTM.

PR #599, opened 2016-04-28
DockerCmd : always run with verbose

edsantiago:
When looking at a debug log of a failure, the first and
most important thing I want to know is: what docker commands
led to this? Too often it's not in the log, and I see this
in the source:

    DockerCmd(...., verbose=False)

...requiring an edit, and rerunning the test, then (when not
trivially reproducible) rerunning and rerunning some more.

Enough. Let's just always log docker commands. Since verbose
is the default, just remove most instances of verbose=False.
(And, while we're at it, also remove verbose=True since
it's a possibly-misleading NOP).

Also: fixed minor misspellings.

### DIFF
--- a/subtests/docker_cli/kill_stopped/kill_utils.py
+++ b/subtests/docker_cli/kill_stopped/kill_utils.py
@@ -80,7 +80,7 @@ class kill_base(subtest.SubSubtest):
         subargs.append("bash")
         subargs.append("-c")
         subargs.append(self.config['exec_cmd'])
-        container = AsyncDockerCmd(self, 'run', subargs, verbose=False)
+        container = AsyncDockerCmd(self, 'run', subargs)
         self.sub_stuff['container_cmd'] = container
         container.execute()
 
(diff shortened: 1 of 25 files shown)


subtests/docker_cli/kill_stopped/kill_utils.py
```diff
@@ -80,7 +80,7 @@ def _init_container_normal(self, name):
         subargs.append("bash")
```
cevich: Just FYI: There's an open issue about the duplicated ``kill_utils.py``.  I haven't looked too closely, but I believe all the separate subtests which use it can/should be combined.  This will make for a massive test, but it's okay b/c it's signalling related and a part of docker that needs through testing.

edsantiago: Yay! I've been wondering but hadn't looked close enough yet (to see if they could be combined). Thanks! Glad to know it's not forgotten.

cevich:
Great , LGTM.

PR #601, opened 2016-04-30
Split dockertest/output.py into a package

edsantiago:
Split dockertest/output.py into a package

Extract classes into four separate files. Move AllGoodBase out of environment.py and into one of the new modules; remove unused EnvCheck. Normalize many import statements.


dockertest/output/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# ARGH. Without this, 'import xceptions' in a submodule fails because
+# dockertest isn't in the search path.
+from os.path import dirname
+__path__ += [dirname(__path__[0])]
+
+# pylint: disable=W0401
+from . dockertime import *
+from . dockerversion import *
+from . texttable import *
+from . validate import *
```
cevich: This is bad-juju, can't ``import *`` b/c it pollutes the namespace.  e.g. ``validate`` may define something with a name that collides with ``dockertime``, however the collision is "out of it's jurisdiction" as it were.

At least until the next minor version,  you'll have to list out the names specificity that you want to import from each module, into the package.

dockertest/output/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""
+API compatibility: on 2016-04-29 output.py got split up into a
+package with four smaller, more maintainable submodules. This init
+allows existing clients to continue importing as a single unit.
+"""
+
+# ARGH. Without this, 'import xceptions' in a submodule fails because
+# dockertest isn't in the search path.
+from os.path import dirname
+__path__ += [dirname(__path__[0])]
```
cevich: IIRC, it's better to use ``dirname(sys.modules[__name__].__file__)``.

dockertest/output/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""
+API compatibility: on 2016-04-29 output.py got split up into a
+package with four smaller, more maintainable submodules. This init
+allows existing clients to continue importing as a single unit.
+"""
+
+# ARGH. Without this, 'import xceptions' in a submodule fails because
```
cevich: I would very much like to just kill the xceptions module, there's really no need for it. All of the exceptions are either standard ones or come from autotest.

run_pylint.sh
```diff
@@ -34,6 +34,7 @@ INIT_HOOK="
 AP = os.environ.get('AUTOTEST_PATH', '/usr/local/autotest')
 sys.path.append(os.path.abspath(AP + '/..'))
 sys.path.append(os.path.abspath('.'))
+sys.path.append(os.path.abspath('dockertest'))
```
cevich: Hrmmmm, I don't think this should be necessary.  It's assumed that ``run_pylint.sh`` is always run with CWD of repo. root.  Since the ``dockertest`` directory has a ``__init__.py`` python should recognise it as a package automaticly.  Perhaps you need to ``import output`` from w/in ``dockertest/__init__.py``?

dockertest/output/__init__.py
edsantiago: Uh-oh. I'm starting to think I've bitten off more than I can chew...

run_pylint.sh
edsantiago: I think this problem must be related to the one I'm seeing with different namespaces in exceptions. But again, I'm starting to wonder if this is a worthwhile effort right now.

dockertest/dockercmd_unittests.py
```diff
@@ -284,13 +282,13 @@ def test_no_fail_docker_cmd(self):
 
         docker_command = self.dockercmd.DockerCmd(self.fake_subtest,
                                                   'unittest_fail')
-        self.assertRaises(self.xceptions.DockerExecError,
+        self.assertRaises(self.output.xceptions.DockerExecError,
```
cevich: This seems wrong.

cevich:
What about adding this to ``dockertest/__init__.py``:

```python
import output
import xceptions

output.xceptions = xceptions
```

You could add a placeholder into ``dockertest/output/__init__.py`` like ``xceptions = None`` to reserve the name.  I think this would be okay, as long as ``xceptions`` never ever imports ``output``, there will be no circular deps.

dockertest/output/__init__.py
cevich: heh, ya, this stuff will drive you crazy.  it's all about namespaces, forget about the directory structures.  Think namespaces, what names need to be anchored where, and then make the directory structure and ``__init__``'s to realize it.

run_pylint.sh
cevich: It will be _much_ easier to do w/in a minor release where upsetting the API (namespace) is more kosher.  Less duct-tape will be used (and we can kill the damn ``xceptions`` module - it's nothing but trouble).

cevich:
One thing that may help is to play with a toy instead.  e.g. make a directory ``package_a`` and ``package_a/package_b``, stick some empty modules in them, and experiment.  You get a ``__doc__`` name in each package and module for free, so just try accessing those.  ``package_a.package_b.module_c.__doc__``

cevich:
Oh, one more, one more thing.  Don't forget about the special, magic ``__all__`` list/tuple inside ``__init__.py``.

edsantiago:
Yeah, I read up on it last week (__all__) but it didn't seem like it was applicable here.

I'm not sure what to do: continue with this, postpone until some minor release, or abandon entirely?

cevich:
> like it was applicable here.

For sure it is, that's the list of names you get if you import the package w/o any qualifiers.  Think of it this way:  You're basically renaming ``output.py`` to ``output/__init__.py``.  You set '__all__' to contain the public-API names that were formerly in the ``output.py`` module.  Then when any caller does a ``from dockertest import output`` it knows which names to load into ``output``.  Otherwise, if a caller does ``from dockertest.output import foo`` python already knows it needs to resolve the name 'foo' (but nothing else).

Anyway, that's my understanding (which could smell a bit like poo, so don't get your nose near by).

dockertest/dockercmd_unittests.py
edsantiago: Absolutely... but if you look at `dockertest/output_unittests.py` there are two already-existing instances of it (`self.output.xceptions.DockerOutputError`) dating to 2014-03-25 (57e6e60). That's where I got the idea.

cevich: Ya, that's just wrong (and I probably simply didn't care enough at the time).   Both the test and the code should be using the same exception, without exception :D  Otherwise, what good is checking?  Might as well compare it to an apple-pie.

cevich: ...clarification:  What I mean is, a subtest should be assume a ``xceptions.DockerOutputError`` would be thrown, and the output module should throw a ``xceptions.DockerOutputError``.  So the unittest should also be checking for a ``xceptions.DockerOutputError``.

edsantiago: \*sob\*. That's exactly how I think it should work, and that's exactly what I've been struggling to make happen, and why my forehead hurts so much from banging on my desk. I am flummoxed!

cevich:
> *sob*. That's exactly how I think it should work

Great, so we're in agreement.  I've shared these headaches before as well.  Maybe less than idea, but you could nuke the unittests, with a ``TODO: write me`` in their place.  Fussing with namespaces is right up there with spaghetti-brain induced Mock()ing.  Bottom-line is test-doing always take priority, so if getting there hinges on a better ``output`` module, we gotta do what we gotta do.

I'm fine too if you want to just let this issue stew, while you work on something else.

edsantiago:
Whew. I'd like to do one more walkthrough on this tomorrow, just as a triple-check, but I think this is as good as it'll get. Glad I didn't let it drop: I think this change may save us some headaches in the future.

edsantiago:
Final review complete. I've addressed almost all of your review comments, the two exceptions being nuking `xceptions.py` (beyond the scope of this) and setting `__all__` (I don't see any point to it, and it would introduce duplication).

dockertest/dockercmd_unittests.py
```diff
@@ -284,13 +285,13 @@ def test_no_fail_docker_cmd(self):
...
                           self.output.mustpass, docker_command.execute())
 
     def test_must_fail_docker_cmd(self):
         docker_command = self.dockercmd.DockerCmd(self.fake_subtest,
                                                   'fake_subcommand')
-        self.assertRaises(self.xceptions.DockerExecError,
+        self.assertRaises(self.DockerExecError,
                           self.output.mustfail, docker_command.execute(), 1)
 
         docker_command = self.dockercmd.DockerCmd(self.fake_subtest,
```
cevich: Nice!

dockertest/environment.py
```diff
@@ -7,176 +7,9 @@
...
-        fullpath = os.path.join(self.envcheckdir, name)
-        # Arguments to subprocess.Popen for script "name"
-        return {'args': fullpath, 'bufsize': 1, 'stdout': subprocess.PIPE,
-                'stderr': subprocess.PIPE, 'close_fds': True, 'shell': True,
-                'env': self.config}
-
-
 def set_selinux_context(pwd, context=None, recursive=True):
     """
     When selinux is enabled it sets the context by chcon -t ...
```
cevich: hehe, okay.

run_unittests.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export PYTHONPATH=.
+
```
cevich: hrmmm, sure this is needed,  I thought CWD was always implied.  Might be safer to do:

``sh
export PYTHONPATH=$(dirname $0):$PYTHONPATH``

or something to that effect.

dockertest/output/dockerversion.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+
+# Pylint runs from a different directory, it's fine to import this way
+# pylint: disable=W0403
+
+import re
+from autotest.client import utils
+import subprocess
+from dockertest.xceptions import DockerOutputError, DockerTestNAError
+from dockertest.version import LooseVersion
```
cevich: oh wow, travis isn't breaking on this anymore? cool!

cevich:
Awesome, ya, I agree this will greatly help maintain a modicum of future sanity, nice job.  The only (minor) concern I have is with the ``PYTHONPATH=.`` but it really isn't a big deal.  I'm wondering if it could cause import problems if you're CWD happens to have a dir / file that would collide with a module name, and you run ``/path/to/docker/autotest/run_unittests.sh``.  Anyway, I think it's very minor.  LGTM.  Nice work!

cevich:
Any reason this should go into 0.8.4 over 0.8.5?  (I'm leaning toward 0.8.4, then bump version).

edsantiago:
PYTHONPATH: fixed, see commit #3. 0.8.[45]: no preference, mostly because I don't have a sense for versioning policicies here.

cevich:
heh, me neither.  doesn't really matter that much for the third-number.  I just like to do it when the number of commits grows large and things are at a fairly "stable" point.  Let's do that, I'll roll the version then merge this.

PR #604, opened 2016-05-10
Adding run test with tmpfs option

afomm:

### DIFF
--- a/config_defaults/subtests/docker_cli/run.ini
+++ b/config_defaults/subtests/docker_cli/run.ini
@@ -1,5 +1,5 @@
 [docker_cli/run]
-subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_passwd,run_ping
+subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_tmpfs,run_passwd,run_ping
 #: The most basic sub-subtests can be generated from a generic
 #: class at runtime, set 'yes' to enable this feature.
 generate_generic = no
@@ -52,6 +52,13 @@ run_append_name = no
 last_name_sticks = yes
 #: The number of ``--name`` options to generate for the command line.
 names_count = 1000
+exit_status = 0
+
+[docker_cli/run/run_tmpfs]
+#: --tmpfs parameter
+tmpfs_path = /run
+run_options_csv = --tmpfs,%(tmpfs_path)s
+cmd = findmnt -n -t tmpfs %(tmpfs_path)s
 exit_status = 0
 
 [docker_cli/run/run_interactive]
--- a/subtests/docker_cli/run/run_tmpfs.py
+++ b/subtests/docker_cli/run/run_tmpfs.py
@@ -1,0 +1,30 @@
+from run import run_base
+from dockertest.output import DockerVersion
+
+
+class run_tmpfs(run_base):
+    #: Run container with --tmpfs option and verifies that it is mounted
+
+    def initialize(self):
+        super(run_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(run_tmpfs, self).run_once()
+        try:
+            self.sub_stuff["cont"].wait_by_name(self.sub_stuff['name'])
+        except ValueError:
+            pass  # container already finished and exited
+
+    def postprocess(self):
+        super(run_tmpfs, self).postprocess()
+        actual_exit_status = self.sub_stuff['dkrcmd'].exit_status
+        expected_exit_status = self.config['exit_status']
+        cmd = self.sub_stuff['dkrcmd'].command
+        self.failif_ne(actual_exit_status, expected_exit_status,
+                       "exit status from %s" % cmd)
+        tmpfs_mnt = self.config['tmpfs_path']
+        tmpfs_mnted = self.sub_stuff['dkrcmd'].stdout.find(tmpfs_mnt) != -1
+        self.failif(not tmpfs_mnted,
+                    "Given --tmpfs not mounted: --tmpfs %s" % tmpfs_mnt)


config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,19 @@ last_name_sticks = yes
 names_count = 1000
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
```
cevich: Don't forget to add ``run_tmpfs`` to the ``subsubtests`` CSV list here: https://github.com/afomm/autotest-docker/blob/afomdev/config_defaults/subtests/docker_cli/run.ini#L2

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class run_tmpfs(run_base):
+    # Verify behavior when multiple --name options passed
+
+    def init_subargs(self):
+        cont = self.sub_stuff["cont"]
+        name_base = cont.get_unique_name()
+        names = []
+        for number in xrange(self.config['names_count']):
```
cevich: this loop is unrelated to testing --tmpfs option, and related config option as well...

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,19 @@ last_name_sticks = yes
...
+cmd = sleep 2s
+# Test will find container by CID instead of name, CID must appear
+# on stdout.
+run_options_csv = --detach --tmpfs /tmp --tmpfs /run
+run_append_name = no
+#: Set to 'yes' if the last generated ``--name`` is used by docker, or
+#: 'no' if the first.  Affects test pass/fail.
+last_name_sticks = yes
+#: The number of ``--name`` options to generate for the command line.
+names_count = 1
```
cevich: ...this one.  (not needed)

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+            name = ('%s_%d' % (name_base, number))
+            names.append(name)
+            self.sub_stuff['containers'].append(name)  # just in case
```
cevich: this line you probably do need, otherwise test-container won't get cleaned up after test.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        self.sub_stuff['subargs'] += ["--name %s" % name for name in names]
```
cevich: this line too (keep it)

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        if self.config['last_name_sticks']:
+            self.sub_stuff['expected_name'] = names[-1]
+        else:
+            self.sub_stuff['expected_name'] = names[0]
+        super(run_tmpfs, self).init_subargs()
```
cevich: ahh, okay, nvm the above comments.  Take a look at ``run_base.init_subargs()``, it will take care of creating the test-container name and appending it.  IIUC, this whole overriden ``init_subargs()`` method (here) is probably unnecessary (dunno 4sure, so double-check me)

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def run_once(self):
+        super(run_tmpfs, self).run_once()
+        cid = self.sub_stuff['cid'] = self.sub_stuff['dkrcmd'].stdout.strip()
+        self.sub_stuff['containers'].append(cid)
+        try:
+            self.sub_stuff["cont"].wait_by_long_id(cid)
```
cevich: Consider ``self.sub_stuff["cont"].wait_by_name(self.sub_stuff['name'])`` might be easier and there's no reason to use the CID and it's caused trouble elsewhere IIRC.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        except ValueError:
+            pass  # container already finished and exited
+
+    def postprocess(self):
+        super(run_tmpfs, self).postprocess()
```
cevich: base class just does basic checks here for exit codes, panic messages, etc.  So you'll want to nuke most of the lines below here, and add your customized checks to verify the ``--tmpfs`` option functioned as expected.  This is where the real _meat_ of your test will go.

if you find the docker command executed (prior to getting here) is a PITA to verify (for --tmpfs), then step back and work on ``initialize()`` and/or ``run_once()`` further.   Bend the behaviour as needed, to make verification easier.  Lather...rinse...and repeat :D

edsantiago:
Forgive me if I'm missing something obvious, but how does this test the --tmpfs option? Aside from failing on docker-1.9?

cevich:
okay, I'm done.  This is really good for never having done a PR before.  Make your changes/fixes, either do a --fixup or --amend commit on top, then push back to your branch.  Then we'll review again :D

cevich:
@edsantiago I suggested in an e-mail about docker 1.9, test should check docker version and raise DockerTestNAError for < 1.10

cevich:
@edsantiago this is his first-ever PR, so be kind :grinning:

edsantiago:
Blush. Sorry about that! I'm just feeling dense from days of poking at Ansible and ADEPT...

cevich:
Hmm, I dunno why I didn't get any notice of your push. Oh, it might be because Travis CI failed (click ``details`` link) - https://travis-ci.org/autotest/autotest-docker/builds/129833134 

You must not be using ``run_unittests.sh`` and ``run_pylint.sh`` before you commit & push.  Those are designed to help find code and style problems.  Otherwise travis runs them automaticly on push.

Fixup whatever it's complaining about, commit, and re-push.  I'll take a look at the code once Travis is happy :grinning:

cevich:
Huh, still no notification on update.  How strange.  Okay, taking a look...

subtests/docker_cli/run/run.py
```diff
@@ -43,6 +43,7 @@ def init_subargs(self):
             self.sub_stuff['containers'].append(name)
             subargs.append('--name')
             subargs.append(name)
+            self.sub_stuff['expected_name'] = name
```
cevich: Probably don't need this as long as you're only working with a single container, you can reference it by ``self.sub_stuff['containers'][-1]``

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,15 @@ last_name_sticks = yes
 names_count = 1000
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+cmd = sleep 2s
+# Test will find container by CID instead of name, CID must appear
+# on stdout.
+run_options_csv = --detach --tmpfs /run
```
cevich: nit-pick, this is suppose to be a comma-separated list of arguments, but for now it won't cause any problems.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+        cont = self.sub_stuff["cont"]
+        json = cont.json_by_long_id(self.sub_stuff['cid'])
+        self.failif(len(json) == 0)
+        actual_option = str(json[0]['HostConfig']['Tmpfs']).split("'")[1]
+        self.failif_ne(actual_option,
+                       self.sub_stuff['expected_option'], "run_tmpfs")
```
cevich: Okay, this looks like a good start for verifying the command-line option made it into the container metadata.  Next, see if you can figure out a way to confirm there really is a tmpfs-type filesystem mounted on ``self.sub_stuff['expected_option']`` (inside the container).

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,28 @@
+from run import run_base
+
+
+class run_tmpfs(run_base):
+    # Verify behavior when multiple --name options passed
+
+    def init_subargs(self):
+        super(run_tmpfs, self).init_subargs()
+        self.sub_stuff['expected_option'] = \
+            self.config['run_options_csv'].split('--tmpfs ')[1]
```
edsantiago: Python .ini files support string interpolation. The code above is fragile: it will fail, e.g., if run_options_csv is converted to actual CSV (which it should be). You can actually get rid of this entire method by changing run.ini into something like:

    tmpfs_path = /run    # or, even better, /SomEthIngUnlikeLY2appearInNature
    run_options_csv = --detach,--tmpfs,%(tmpfs_path)s

...and using `self.config['tmpfs_path']` in `postprocess()`.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+            self.sub_stuff["cont"].wait_by_name(self.sub_stuff['name'])
+        except ValueError:
+            pass  # container already finished and exited
+
+    def postprocess(self):
+        super(run_tmpfs, self).postprocess()
+        cont = self.sub_stuff["cont"]
+        json = cont.json_by_long_id(self.sub_stuff['cid'])
+        self.failif(len(json) == 0)
+        actual_option = str(json[0]['HostConfig']['Tmpfs']).split("'")[1]
```
edsantiago: This is fragile: it depends on an exact string representation. One fairly easy way to fix it might be to compare dicts directly:

    expected_dict = { self.config['tmpfs_path'] : '' }
    actual_dict = json[0]['HostConfig']['Tmpfs']
    self.failif_ne(actual_dict, expected_dict, 'Container .HostConfig.Tmpfs info setting')

subtests/docker_cli/run/run_tmpfs.py
edsantiago: What @cevich said! Nice start, but can you think of a way to test that the given filesystem is actually accessible within the container as a tmpfs?

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+        self.sub_stuff['containers'].append(cid)
+        try:
+            self.sub_stuff["cont"].wait_by_name(self.sub_stuff['name'])
+        except ValueError:
+            pass  # container already finished and exited
+
+    def postprocess(self):
+        super(run_tmpfs, self).postprocess()
+        cont = self.sub_stuff["cont"]
+        json = cont.json_by_long_id(self.sub_stuff['cid'])
```
edsantiago: "json" as a variable name doesn't indicate what the variable actually holds. Can you put yourself in the shoes of a future maintenance programmer -- possibly you :wink: -- reading this code a year from now? What variable name do you think would be most helpful in understanding what this code is trying to accomplish?

cevich:
@edsantiago You're tips are spot-on (as per usual).  As an FYI- Afom's in the "baby-steps" phase, so it's okay (IMHO) to leave out some clever/advanced suggestions for now.  i.e. Learning the pull-request workflow and getting a "working" test committed is a **huge** step for him.  I'm okay with accepting a bit  lower (relative) quality initially.  Especially if doing so provides additional learning opportunities (like submitting improvement PRs / bugfixes).  How many painful lessons did you have to learn when getting started coding?

@afomm Ed's comments are very good suggestions, they will save you TONs of time later.  Don't hesitate to ask either of us for clarification if you don't understand how/why.

subtests/docker_cli/run/run_tmpfs.py
afomm: OK. That & version checking (it runs with 1.10 or later) is next. Of course, I'll clean it with your suggestions first. Thanks for all the feedback.

subtests/docker_cli/run/run_tmpfs.py
afomm: I agree. More descriptive variables will be used from now on.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class run_tmpfs(run_base):
+    # Verify behavior when multiple --name options passed
+
+    def initialize(self):
+        super(run_tmpfs, self).initialize()
+        # On docker 1.10, the second tag should pass. On < 1.10, fail.
+        expected_version = self.config['min_docker_ver']
+        try:
```
edsantiago: You might be overthinking this :-). All you need is:

    DockerVersion().require_server(self.config['min_docker_ver'])

No need for the try/except/failif_ne

afomm: Thanks - reduced it to just a single line. But hard coded the required version.

By the way, my previous pushes weren't passing Travis because I didn't document the new variables.

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,20 @@ last_name_sticks = yes
...
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+cmd = 'sleep 5s'
+# command to execute in running container
+# passed as argument to exec
+fndmnt = findmnt -n %(tmpfs_path)s 
+# Test will find container by CID instead of name, CID must appear
+# on stdout.
+run_options_csv = --detach=true,--tmpfs %(tmpfs_path)s
```
edsantiago: Nit: trailing whitespace. I don't know why github isn't griping about it; @cevich any ideas if there's a setting for that in the repo config? Anyway, @afomm, if you use emacs search for Noah Friedman's nuke-trailing-whitespace.el; if vi, any search for vim+whitespace will offer advice on detecting/removing it.

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,20 @@ last_name_sticks = yes
...
+tmpfs_path = /run
```
edsantiago: Nit: although Python interprets %(foo) strings at runtime, humans tend to interpret in a top-down manner. To make your reader's life easier, would you consider moving this declaration to earlier in the sequence?

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,20 @@ last_name_sticks = yes
 names_count = 1000
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+cmd = 'sleep 5s'
```
edsantiago: Suggestion for a bit of an overhaul: instead of this `sleep` and a separate docker command, why not remove the `--detach` from `run_options_csv` and use `cmd = findmnt ...` ? First, that simplifies the code enormously. Second, it's robust: with the sleep, you have a race condition because on a loaded system the `sleep` container may finish before the `findmnt` one starts. I hope this makes sense; please ask for clarification as needed.

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,20 @@ last_name_sticks = yes
...
+# command to execute in running container
+# passed as argument to exec
+fndmnt = findmnt -n %(tmpfs_path)s
```
edsantiago: AFAICT it should be safe to add `-o FSTYPE`; this makes it easy to test that it's actually the expected `tmpfs` type. (Your test currently verifies only that /run is mounted).

edsantiago:
This is markedly improved from last week. Nice work!

cevich:
> if you use emacs...if vi...

Eep! Those are the only two choices now :open_mouth: 

There's no github repo setting I'm aware of to handle whitespace stuff, though Git can be made to complain about it.  Here's a bit from my .vimrc that does the trick (you may need to change the colors):

```vim
....
set bg=dark
hi ExtraWhitespace ctermbg=darkgreen
match ExtraWhitespace /\s\+\%#\@<!$/
match ExtraWhitespace /\s\+$\| \+\ze\t/
match ExtraWhitespace /[^\t]\zs\t\+/
autocmd InsertLeave * redraw!
let c_space_errors = 1
let python_space_errors = 1
...
```

(I have no idea _how_ this works, it's some magic incantation I found by google)

config_defaults/subtests/docker_cli/run.ini
cevich: Actually, IIRC python's ``ConfigParser`` operates only in line-forward mode, for interpreting those.  So unless I'm wrong, you _have_ to define it earlier in the file before you use it.

edsantiago: $ pydoc ConfigParser | grep -A1 resolve  
        would resolve the "%(dir)s" to the value of dir.  All reference
        expansions are done late, on demand.

I had to look it up before I commented :-). And the fact that I had to look it up is itself an argument for declaring it early.

config_defaults/subtests/docker_cli/run.ini
cevich: Ed's right, using sleep sucks (I forgot that was in there).  Better to block on a much more deterministic condition (like a docker command exiting).  There are way too many tests that do the sleep thing, but most were written in haste and desperation - terrible examples to follow.  Sorry about that.

config_defaults/subtests/docker_cli/run.ini
cevich: Ahh it's a lazy-parser, good, much better design.

``pydoc``?

Cool! 

I didn't know about that, will have to check it out.

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,16 @@ last_name_sticks = yes
...
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+#: --tmpfs parameter
+tmpfs_path = /run
+run_options_csv = --tmpfs %(tmpfs_path)s
+#: command to run in container to check if passed --tmpfs is mounted as tmpfs type
+fndmnt = findmnt -n -t tmpfs %(tmpfs_path)s
+cmd = 'sleep 2s && %(fndmnt)s'
+run_append_name = True
```
edsantiago: This is already the default, and it isn't especially germane to this test. Why not just remove it?

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,16 @@ last_name_sticks = yes
 names_count = 1000
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+#: --tmpfs parameter
+tmpfs_path = /run
+run_options_csv = --tmpfs %(tmpfs_path)s
+#: command to run in container to check if passed --tmpfs is mounted as tmpfs type
+fndmnt = findmnt -n -t tmpfs %(tmpfs_path)s
+cmd = 'sleep 2s && %(fndmnt)s'
```
edsantiago: Why the sleep?

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,29 @@
...
+            self.sub_stuff["cont"].wait_by_name(self.sub_stuff['name'])
+        except ValueError:
+            pass  # container already finished and exited
+
+    def postprocess(self):
+        super(run_tmpfs, self).postprocess()
+        actual_exit_status = self.sub_stuff['dkrcmd'].exit_status
+        expected_exit_status = self.config['exit_status']
+        self.failif_ne(actual_exit_status, expected_exit_status,
+                       'actual_exit_status != expected_exit_status')
```
edsantiago: Put yourself in the shoes of someone seeing this message. Will this description help them understand the error? How about this one instead?

    'exit status from %s' % self.config['cmd']

edsantiago:
Just keeps getting better and better! Thank you for all your hard work.

config_defaults/subtests/docker_cli/run.ini
cevich: Maybe he found a docker bug :grinning:

cevich:
Thanks for help reviewing this Ed, I still don't understand why I'm not seeing push-notifications on this.  I'm subscribed to it, and I certainly see notifications from your PRs.  Maybe it's because I'm assigned to it?  So strange.

edsantiago:
I don't understand it either. github says I'm subscribed, but the only notifications I'm getting are via the pale blue dot on the upper-right on github. I received your "So strange" comment as an email, but I'm not getting notified of @afomm's pushes.

cevich:
OMG really, so it's not just me.  Huh.  Oh, I wonder if it might have something to do with the github group memberships, I'll check that out.

@afomm  until it's fixed, when you push and want us to take a look, just mention us in a comment with ``@cevich`` and ``@edsantiago`` in the comment text.  That seems to force github to send out e-mail notifications.

I'll check into your group/team membership also, maybe that's the problem.

afomm:
Made the suggested changes - @cevich and @edsantiago please review again.

config_defaults/subtests/docker_cli/run.ini
afomm: Should I also remove exit_status?

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -54,6 +54,15 @@ last_name_sticks = yes
 names_count = 1000
 exit_status = 0
 
+[docker_cli/run/run_tmpfs]
+#: --tmpfs parameter
+tmpfs_path = /run
+run_options_csv = --tmpfs,%(tmpfs_path)s
+#: command to run in container to check if passed --tmpfs is mounted as tmpfs type
+fndmnt = findmnt -n -t tmpfs %(tmpfs_path)s
+cmd = '%(fndmnt)s'
```
edsantiago: My 'why sleep' was an earnest question: I didn't know if it was necessary or just for luck.

If it isn't necessary, though, neither is this two-step process: there's no need to define an intermediate single-purpose variable, it's much cleaner to just say:

    cmd = findmnt ...

config_defaults/subtests/docker_cli/run.ini
edsantiago: >Should I also remove exit_status?
￼
That's a more interesting question, because exit_status is vital to understanding the test. So even though it's duplication to list it explicitly, it _might_ make it slightly quicker for a future reader to understand. My (slight) preference is to remove it, but the existing convention in `run.ini` is clearly to reiterate it in each subsubtest -- so I would recommend leaving it, in deference to convention.

But I don't have a strong preference either way, and will accept whatever you choose.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -10,39 +7,24 @@ class run_tmpfs(run_base):
...
-        OutputGood(self.sub_stuff['tmpfs_cmdresult'])
-        mustpass(self.sub_stuff['tmpfs_cmdresult'])
+        actual_exit_status = self.sub_stuff['dkrcmd'].exit_status
+        expected_exit_status = self.config['exit_status']
+        cmd = self.sub_stuff['dkrcmd'].command
+        self.failif_ne(actual_exit_status, expected_exit_status,
+                       "exit status from %s" % cmd)
         tmpfs_mnt = self.config['tmpfs_path']
-        tmpfs_mnted = tmpfs_cmdresult.stdout.find(tmpfs_mnt) != -1
+        tmpfs_mnted = self.sub_stuff['dkrcmd'].stdout.find(tmpfs_mnt) != -1
```
edsantiago: @cevich look what I found, a perfect case to reuse my new `failif_not_in()`!

config_defaults/subtests/docker_cli/run.ini
afomm: @edsantiago and @cevich - no particular reason for sleep & it is now removed. cmd is now changed to suggestion above.

edsantiago:
LGTM. Thanks again!

afomm:
@edsantiago & @cevich thanks!

config_defaults/subtests/docker_cli/run.ini
cevich: Don't let the quotes bite you either, they are preserved right down through execution.  I'm not sure off-hand if they're needed here or not, but I've certainly been tripped up myself by adding them into ``ini`` file when there were not needed.

config_defaults/subtests/docker_cli/run.ini
cevich: When in doubt, choose the path which requires the least amount of work.  Pressing the delete-key several times is constituted as work in my book :grinning: 

Seriously though, clarity is important, and IMHO removing it really doesn't save a whole lot of typing.  Unless you remove it from all the other subtests here.  But that's even more work.  So I say just leave it.

subtests/docker_cli/run/run_tmpfs.py
cevich: Ok man, you got me.  You've won the internet.

subtests/docker_cli/run/run_tmpfs.py
```diff
@@ -0,0 +1,30 @@
+from run import run_base
+from dockertest.output import DockerVersion
+
+
+class run_tmpfs(run_base):
+    # Verify behavior when multiple --name options passed
```
cevich: Tip: If you're going to put a comment here, it might as well be a docstring :grinning: 

(no need to fix)

cevich:
@afomm okay this LGTM now, one last thing before we merge.  Could you combine all the commits into one?  ``git rebase -i upstream/master`` then change all the ``pick`` into ``s`` (for squash).  That will jam all the commit messages into one and let you edit them.  If you need help doing this, let me know.

PR #605, opened 2016-05-27
Test docker login

edsantiago:
Various forms of docker login/logout, including push with/without valid credentials.

### DIFF
--- a/config_defaults/subtests/docker_cli/login.ini
+++ b/config_defaults/subtests/docker_cli/login.ini
@@ -1,0 +1,14 @@
+[docker_cli/login]
+subsubtests = login_ok,login_fail,logout_ok,push_ok,push_fail
+#: docker image containing a functional registry, including auth checks.
+#: Make sure to include this in config_custom/defaults.ini:preserve_fqins !
+registry_fqin = registry-1.docker.io/distribution/registry:2.1
+#: docker image to use for the push test. Make it something small, so test
+#: can pass/fail quickly and without consuming too many system resources.
+#: As of this writing (2016-05-26) busybox is just over 1MB.
+push_image = busybox
+#: config file into which docker login writes credentials. This is not
+#: likely to change.
+docker_auth_file = /root/.docker/config.json
+#: maximum time, in seconds, to wait for registry container to become ready.
+wait_for_registry_ready = 10
--- a/dockertest/subtestbase.py
+++ b/dockertest/subtestbase.py
@@ -139,6 +139,24 @@ class SubBase(object):
         spec = "{}: expected " + arg + "; got " + arg
         raise DockerTestFail(spec.format(reason, expected, actual))
 
+    @staticmethod
+    def failif_not_in(needle, haystack, description=None):
+        """
+        Convenience method for subtests to test for an expected substring
+        being contained in a larger string, e.g. to look for XYZ in a
+        command's stdout/stderr.
+
+        :param needle: the string you're looking for
+        :param haystack: the actual string, e.g stdout results from a command
+        :param description: description of haystack, e.g. 'stdout from foo'
+        """
+        if description is None:
+            description = 'string'
+        if needle in haystack:
+            return
+        raise DockerTestFail("Expected string '%s' not in %s '%s'"
+                             % (needle, description, haystack))
+
     @classmethod
     def log_x(cls, lvl, msg, *args):
         """
--- a/subtests/docker_cli/login/login.py
+++ b/subtests/docker_cli/login/login.py
@@ -1,0 +1,368 @@
+r"""
+Summary
+-------
+
+Test variations of 'docker login'
+
+Operational Summary
+-------------------
+
+#. Create a self-signed SSL certificate
+#. Install cert locally for use by docker daemon (FIXME:seems unnecessary)
+#. Generate a pseudorandom password
+#. Generate htpasswd file with a test user and this password
+#. Run a registry container using new certificate & account
+#. Run individual subtests against this standard setup
+"""
+
+import os
+from base64 import b64encode
+import errno
+import json
+import urllib2
+import time
+from shutil import copyfile
+from autotest.client import utils
+from dockertest.containers import DockerContainers
+from dockertest.images import DockerImages
+from dockertest.subtest import SubSubtest
+from dockertest.output import OutputGood
+from dockertest.output import mustpass, mustfail
+from dockertest.dockercmd import DockerCmd
+from dockertest.xceptions import DockerTestFail
+from dockertest import subtest
+
+
+class login(subtest.SubSubtestCaller):
+
+    """ SubSubtest caller """
+
+
+class login_base(SubSubtest):
+
+    """ login base class; performs common setup. """
+
+    def initialize(self):
+        super(login_base, self).initialize()
+
+        # We need to create cert and htpasswd files for bind-mounting in
+        # the registry container; do so in a safe temporary workdir.
+        os.chdir(self.tmpdir)
+
+        # Creates domain.crt and domain.key files
+        self._create_certs()
+
+        # Credentials: fixed username with pseudorandom password
+        self._create_htpasswd()
+
+        self._preserve_docker_auth_file()
+
+        # Now run the registry itself, leave it running for the login test
+        c_name = DockerContainers(self).get_unique_name()
+        subargs = ['-d', '-p', '{port}:{port}'.format(**self.sub_stuff),
+                   '--name', c_name,
+                   '-v', '{}:/auth:Z'.format(self.tmpdir),
+                   '-e', 'REGISTRY_HTTP_TLS_CERTIFICATE=/auth/domain.crt',
+                   '-e', 'REGISTRY_HTTP_TLS_KEY=/auth/domain.key',
+                   '-e', 'REGISTRY_AUTH=htpasswd',
+                   '-e', 'REGISTRY_AUTH_HTPASSWD_REALM="Registry Realm"',
+                   '-e', 'REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd',
+                   self.config['registry_fqin']]
+        self.sub_stuff['my_containers'] = [c_name]
+        self.sub_stuff['my_images'] = []
+        mustpass(DockerCmd(self, 'run', subargs).execute())
+        self._wait_for_registry_ready()
+
+    def _create_certs(self):
+        """
+        Create a self-signed SSL certificate for use by the registry
+        container (registry auth only works with TLS).
+        """
+        localhost = 'localhost.localdomain'
+        port = '5000'
+        subj = '/C=US/ST=Foo/L=Bar/O=Red Hat, Inc./CN=%s' % localhost
+        make_cert = ('openssl req -newkey rsa:4096 -nodes -sha256'
+                     ' -keyout domain.key -x509 -days 2 -out domain.crt'
+                     ' -subj "%s"' % subj)
+        utils.run(make_cert, verbose=True)
+
+        self.sub_stuff['localhost'] = localhost
+        self.sub_stuff['port'] = port
+        self.sub_stuff['server'] = localhost + ':' + port
+
+        # Make certificate available to our docker daemon
+        # FIXME: remove this. It doesn't seem like this is necessary.
+        #       docker login and push work exactly as expected even if I
+        #       comment out the copyfile. Not surprising, since I wouldn't
+        #       expect the copied cert to take effect without a daemon
+        #       restart (which we can't do, because of docker-latest
+        #       confusion). So I'm proposing leaving this in for my
+        #       first commit in the PR, then immediately removing it.
+        #       That at least leaves it for posterity.
+        docker_cert_dir = '/etc/docker/certs.d/{server}'.format(
+            **self.sub_stuff)
+        self._mkdir_p(docker_cert_dir)
+        docker_cert_file = os.path.join(docker_cert_dir, 'domain.crt')
+        copyfile('domain.crt', docker_cert_file)
+        # FIXME: now we presumably need to restart docker daemon
+
+        self.sub_stuff['docker_cert_dir'] = docker_cert_dir
+        self.sub_stuff['docker_cert_file'] = docker_cert_file
+
+    @staticmethod
+    def _mkdir_p(path):
+        """
+        mkdir, but stay cool if directory already exists.
+        """
+        try:
+            os.mkdir(path)
+        except OSError as e:
+            if e.errno == errno.EEXIST and os.path.isdir(path):
+                pass
+            else:
+                raise
+
+    def _create_htpasswd(self):
+        """
+        Write an htpasswd file in our current directory, for use by
+        the registry container (via bind mount). Use the registry image
+        itself to run htpasswd, with -n so output goes to stdout, then
+        capture that stdout and write it to a file in our current directory.
+        """
+        self.sub_stuff['user'] = 'testuser'
+        self.sub_stuff['passwd'] = utils.generate_random_string(12)
+
+        registry_fqin = self.config['registry_fqin']
+        subargs = ['--rm', '--entrypoint', 'htpasswd', registry_fqin,
+                   '-Bbn', self.sub_stuff['user'], self.sub_stuff['passwd']]
+        htpasswd_cmdresult = DockerCmd(self, 'run', subargs).execute()
+        with open('htpasswd', 'wb') as htpasswd_fh:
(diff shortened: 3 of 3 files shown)


config_defaults/subtests/docker_cli/login.ini
```diff
@@ -0,0 +1,12 @@
+[docker_cli/login]
+subsubtests = login_ok,login_fail,logout_ok,push_ok,push_fail
+#: docker image containing a functional registry, including auth checks.
+#: Make sure to include this in config_custom/defaults.ini:preserve_fqins !
```
cevich: You could just add a ``preserve_fqins`` here and override the default...errr...wait, nope that won't work.  Never-mind.

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+        subargs = ['--rm', '--entrypoint', 'htpasswd', registry_fqin,
+                   '-Bbn', self.sub_stuff['user'], self.sub_stuff['passwd']]
+        htpasswd_cmdresult = DockerCmd(self, 'run', subargs).execute()
+        with open('htpasswd', 'wb') as htpasswd_fh:
+            htpasswd_fh.write(htpasswd_cmdresult.stdout)
+
+    def _preserve_docker_auth_file(self):
+        """
+        docker login creates/overwrites a standard file; preserve it
+        if it already exists.
```
cevich: Awww so nice, thanks!  (and good catch)

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+        a well-defined v2 URL in the registry. It will fail with SSLEOF
+        at the beginning, then fail with HTTP 401 (Unauthorized). Once
+        we get that, trust that it's ready for login.
+        """
+        if timeout is None:
+            timeout = 10
+
+        url = "https://{server}/v2/".format(**self.sub_stuff)
+        end_time = time.time() + timeout
+        while time.time() <= end_time:
```
cevich: Don't remember if this timeout is configurable, but if so, this could be a problem if ``timeout == 0`` or is a very small number.  There's a (small) race-condition here because time passes between setting ``end_time`` and evaluating the ``while`` condition.  You can avoid that (to a greater extent) with:

``while time.time() <= time.time() + timeout:``

or

```python
while time.time() <= end_time:
    ...
else:
    raise DockerTestError("Timeout %s is too damn small" % timeout)
```

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+            try:
+                urllib2.urlopen(url)
+            except urllib2.HTTPError as e:
+                if e.reason == 'Unauthorized':
+                    return
+            except urllib2.URLError as e:
+                # FIXME: assume that it's SSLEOFError; how can we make sure?
+                pass
+            time.sleep(0.5)
+        raise DockerTestFail("timed out waiting for registry")
```
cevich: I really like your wait-strategy here and the very nice comment as well.  Nicely done.

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+                        ('"%s" should not be in %s after docker logout'
+                         % (server, auth_file)))
+        else:
+            # Stored credentials
+            cred_hash = creds['auths'][server]['auth']
+            # ...simply a base64-encoded('username:password'), see:
+            #  https://coreos.com/os/docs/latest/registry-authentication.html
+            expected = b64encode('{user}:{passwd}'.format(**self.sub_stuff))
+            self.failif_ne(cred_hash, expected,
+                           'saved credentials in %s' % auth_file)
```
cevich: Wow, uhhh, this hashing business was actually documented somewhere ya?

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+    def docker_push(self, img_name, tag_name):
+        """
+        Run docker tag, then docker push against our local registry.
+        Tag should always succeed but make no assumptions about push.
+        """
+        push_image = self.config['push_image']
+        dc = DockerCmd(self, 'pull', [push_image])
+        mustpass(dc.execute())
+        self.sub_stuff['my_images'] = [push_image]
+        # FIXME: add push_image to self.containers_needing_cleanup
```
cevich: Did you forget this?  Or is this PR just a first-pass (it appears way-too high-quality for a first-pass to me) :grinning:

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def docker_push(self, img_name, tag_name):
+        """
+        Run docker tag, then docker push against our local registry.
+        Tag should always succeed but make no assumptions about push.
+        """
+        push_image = self.config['push_image']
+        dc = DockerCmd(self, 'pull', [push_image])
+        mustpass(dc.execute())
+        self.sub_stuff['my_images'] = [push_image]
```
cevich: Did you mean ``self.sub_stuff['my_images'] += [push_image]`` or was ``my_images`` not initialized to a list yet?

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+    def run_once(self):
+        super(login_ok, self).run_once()
+        self.sub_stuff["cmdresult"] = self.docker_login()
+
+    def postprocess(self):
+        super(login_ok, self).postprocess()
+        cmdresult = self.sub_stuff['cmdresult']
+        OutputGood(cmdresult)
+        mustpass(cmdresult)
+        self.failif_not_in("Login Succeeded", cmdresult.stdout,
```
cevich: You are aware this changes to ``Successful login`` in Docker 1.12 and then it changes to ``Login successful`` in 1.13 and then...

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+        wrong_passwd = '!' + self.sub_stuff['passwd']
+        self.sub_stuff["cmdresult"] = self.docker_login(wrong_passwd)
+
+    def postprocess(self):
+        super(login_fail, self).postprocess()
+        cmdresult = self.sub_stuff['cmdresult']
+        OutputGood(cmdresult, ignore_error=True)
+        mustfail(cmdresult, 1)
+        self.failif_not_in("no successful auth challenge", cmdresult.stderr,
+                           "stderr from failed docker login")
```
cevich: Curiosity: Does it give you the same failure-pattern if the login is wrong?  If not, it's a credentials-leak :grinning:

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def run_once(self):
+        super(push_ok, self).run_once()
+        self.docker_login()
+        self.sub_stuff['pushresult'] = self.docker_push('push_ok', 'tag_ok')
+
+    def postprocess(self):
+        super(push_ok, self).postprocess()
+        cmdresult = self.sub_stuff['pushresult']
+        print cmdresult.stdout
```
cevich: debugging straggler?

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def run_once(self):
+        super(push_fail, self).run_once()
+        self.docker_login('!' + self.sub_stuff['passwd'])
+        self.sub_stuff['pushresult'] = self.docker_push('push_fail', 'tag_x')
+
+    def postprocess(self):
+        super(push_fail, self).postprocess()
+        cmdresult = self.sub_stuff['pushresult']
+        print cmdresult.stderr
```
cevich: :scream: another one! Quick, stomp on it!

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        self.sub_stuff['user'] = 'testuser'
+        self.sub_stuff['passwd'] = utils.generate_random_string(12)
+
+        registry_fqin = self.config['registry_fqin']
+        subargs = ['--rm', '--entrypoint', 'htpasswd', registry_fqin,
+                   '-Bbn', self.sub_stuff['user'], self.sub_stuff['passwd']]
+        htpasswd_cmdresult = DockerCmd(self, 'run', subargs).execute()
+        with open('htpasswd', 'wb') as htpasswd_fh:
+            htpasswd_fh.write(htpasswd_cmdresult.stdout)
```
cevich: Wow, nicely done setup, so compact!  I'm glad you discovered an 'easy' way.

cevich:
Beautiful work Ed, I can't tell if this is a first pass or almost final-draft :grinning: 

I don't see it as a problem for getting this merged, but maybe for a future enhancement-

Is it "safe" to just ``docker login`` over and over again?  What I mean is, if there's a chance a prior subtest or sub-subtest failed to login/logout, could it affect other tests?  If so, that could cause some really difficult to debug failures.  Hopefully the client is mostly stateless (across separate runs). If that ever changes though, ugg, what a mess it could make.  e.g. if a later test tries to docker-pull but the creds are wrong.   I didn't look closely enough to know if this is a problem or not.  Just a thought I had.

subtests/docker_cli/login/login.py
edsantiago: timeout wasn't configurable, but I've made it so. Do I really have to worry about someone setting it to <= 1? Can we leave that as you-deserve-what-you-get?

subtests/docker_cli/login/login.py
edsantiago: The FIXME comment is because the exception (`e`) is a dict containing a tuple containing, possibly, SSLEOFError. Diving into that made my head hurt. Is it safe enough to just handle URLError? (I lean that way)

subtests/docker_cli/login/login.py
edsantiago: Sigh. Comment removed: the line right above it is where I added the code, without actually removing the FIXME. Thank you.

subtests/docker_cli/login/login.py
edsantiago: It wasn't initialized in `initialize()`, but on reflection it really should be. Fixed. And `=` is now `+=`. Thank you.

subtests/docker_cli/login/login.py
edsantiago: Yep, I fully considered that. In the interests of sanity, I chose to defer configuration until it's actually necessary. Would you prefer that I make this (and logout and push messages) configurable right now?

subtests/docker_cli/login/login.py
edsantiago: Good point. Yes, it's the same message (or at least that same string within the message). Is this  worth a subsubtest?

subtests/docker_cli/login/login.py
edsantiago: Sigh. Yes. Fixed. (Not yet pushed).

subtests/docker_cli/login/login.py
edsantiago: Sigh.... thanks again.

edsantiago:
If tests will be running in parallel, the preserve/restore of `/root/.docker/config.json` might screw up other tests running at the same instance in time. Otherwise, given my understanding of how login works, no, there's no risk of affecting/being affected by later/earlier tests. Keep in mind that everything this test does is on the `localhost.localdomain:5000` registry.

edsantiago:
New force push. Useless comment simply to confirm that @cevich received a notification other than this one.

subtests/docker_cli/login/login.py
cevich: Yes this is fine, it's a very minor / unlikely issue IMHO.

subtests/docker_cli/login/login.py
cevich: Totally, just kill the fixme.  There are way too many things that can go wrong w/ most things involving networking.  Simply checking for success / !success is fine, and logging as many details as is practical.  Most of the time when it breaks, it's because someone fudged a name/ip or the server is just down.

subtests/docker_cli/login/login.py
cevich: NP, I have the opposite problem - useless/WTF code and no ``FIXMES`` :grinning:

subtests/docker_cli/login/login.py
cevich: Oh, okay, as long as you thought about it.  No, it doesn't bother me too much now, but it will when it breaks :grinning:  Since you're already checking for a clean exit, it might even be okay to "increase by trimming, the intersection angles of the three adjoining surfaces", by writing ``self.failif_not_in("ucce", ...``.  IMHO, that four-letter combination is unlikely to appear in any failure message :grinning:

subtests/docker_cli/login/login.py
cevich: Hmmm, I'm not sure.  It's a fairly well-known problem, so I think liklihood of docker breaking vs the test breaking is not in our hair-length's favor.  I'm fine w/o the extra work, it can always be a TODO.

subtests/docker_cli/login/login.py
edsantiago: > self.failif_not_in("ucce", .... IMHO, that four-letter combination is unlikely to appear in any failure message

"Unsuccessful login attempt" ?

subtests/docker_cli/login/login.py
cevich: I wonder...because I make this same mistake as well as leaving being ``pdb.set_trace()`` sometimes.  Both would be good things to add to our checkers.  I'll make a note to take a stab at that.

cevich:
> If tests will be running in parallel

This almost never happens, and would not be very kosher testing since there's a single user and daemon involved.

edsantiago:
New force-push; comment-only changes on the second commit. I can't figure out a way to ask github to show only those diffs; since it was a force, perhaps there isn't any way.

cevich:
> confirm that @cevich received a notification other than this one.

Not on any of the pushes, only on the comments :frowning:

subtests/docker_cli/login/login.py
cevich: I hate you

cevich:
> perhaps there isn't any way.

I think you can stick commit IDs into the compare tool (as if you were going to make a new PR).  But probably more trouble than it's worth.  Taking a final look...

dockertest/subtestbase.py
```diff
@@ -139,6 +139,24 @@ def failif_ne(actual, expected, reason=None):
...
+    @staticmethod
+    def failif_not_in(needle, haystack, description=None):
+        """
+        Convenience method for subtests to test for an expected substring
+        being contained in a larger string, e.g. to look for XYZ in a
+        command's stdout/stderr.
+
+        :param needle: the string you're looking for
+        :param haystack: the actual string, e.g stdout results from a command
+        :param description: description of haystack, e.g. 'stdout from foo'
```
cevich: Tip:  This one is also supported:
```
:raise DockerTestFail: When needle not found in haystack.

File: dockertest/subtestbase.py [hunk 14]
```diff
@@ -139,6 +139,24 @@ def failif_ne(actual, expected, reason=None):
...
+        """
+        if description is None:
+            description = 'string'
+        if needle in haystack:
```
cevich: I don't remember exactly, but seem to recall that using ``needle.find(haystack) > -1`` is a lot faster when haystack is large.  Not a problem, just something that I seem to remember being called on once a long time ago.

subtests/docker_cli/login/login.py
```diff
@@ -0,0 +1,329 @@
...
+        """
+        push_image = self.config['push_image']
+        dc = DockerCmd(self, 'pull', [push_image])
+        mustpass(dc.execute())
+        self.sub_stuff['my_images'] += [push_image]
+
+        self.sub_stuff['img_name'] = img_name
+        self.sub_stuff['tag_name'] = tag_name
+        pushed_name = ('{server}/{img_name}:{tag_name}'
+                       .format(**self.sub_stuff))
```
cevich: I really like this using the "new-way" (string formatting), much easier to read.

cevich:
Okay, LGTM, really nice work, super-easy to read, well done.

dockertest/subtestbase.py
edsantiago: Thanks for the link, but why did I not get it in the github email? Did you edit/revise your comment? Anyway, the link answers my initial question re: whether it's `:raise` or `:raises` (singular vs plural); both are in the code base. Answer: both are valid. 

TODO in a separate checkin: fix `dockertest/containers.py` which has a `:rais` (no 'e'), and several files with `:raise: ExceptionName` (misplaced colon).

PR #607, opened 2016-06-01
Extra run_pylint checking

cevich:
It's been an unwritten rule that the pyhong 'print'
statement shouldn't appear anywhere in module or test code.
Additionally, while used for debugging, the pdb module also
shouldn't appear in any checked-in code.  Update run_pylint
to with a function to check for both of these.

There is no easy way to add these checks to pylint/pep8.  In
the case of pylint and pdb, the workaround is to mark the pdb
module itself as deprecated instead.  This ensures all the possible
ways of importing it are checked.

In the case of print, the simplest solution is to just egrep
for it in every checked file.  If found, the file and line
references are printed.

Finally, I added a check for AUTOTEST_PATH or successful autotest
import, similar to the one that appears in run_unittests.sh

### DIFF
--- a/run_pylint.sh
+++ b/run_pylint.sh
@@ -22,11 +22,24 @@ then
     fi
 fi
 
-TMPFILENAME="/tmp/run_pylint_$RANDOM"
+# Tests need to include autotest modules; make sure we can access them.
+if [ -z "$AUTOTEST_PATH" ]
+then
+    python -c 'import autotest'
+    if [ "$?" -ne "0" ]
+    then
+        echo "ERROR: autotest module won't load or " \
+             "AUTOTEST_PATH env. var. is not set"
+        exit 1
+    fi
+fi
+
+TMPFILENAME="$(mktemp --suffix=run_pylint)"
 PEP8=`which pep8`
 # for readability, allow multiple spaces after commas, colons & before '='
 PEP8IGNORE='E731,E221,E241'
 MSGFMT='(pylint) {msg_id}:{line:3d},{column}: {obj}: {msg}'
+# General 'bad' stuff that should never appear
 # Disable 'line too long' - will be picked up by pep8
 # Check "note" (W0511) separately
 DISABLEMSG="I0011,R0801,R0904,R0921,R0922,C0301,C0326,W0511${SPECIALONE}"
@@ -73,6 +86,17 @@ record_return() {
     fi
 }
 
+check_print() {
+    PRINTS=$(egrep --with-filename --line-number '^\s*print\s+' "$1")
+    if [ -n "$PRINTS" ]
+    then
+        echo "Python print statement(s) found:"
+        echo "$PRINTS"
+        return 1
+    fi
+    return 0
+}
+
 # Run a command, checking exit status.
 # If command fails: if called with --FF, exit immediately. Otherwise,
 # continue but remember to exit with failure at end of tests.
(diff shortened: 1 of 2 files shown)


run_pylint.sh
```diff
@@ -22,11 +22,23 @@ then
     fi
 fi
 
-TMPFILENAME="/tmp/run_pylint_$RANDOM"
+# Tests need to include autotest modules; make sure we can access them.
+if [ -z "$AUTOTEST_PATH" ]
```
edsantiago: I assume you caught the missing 'then'. But while you're at it...

run_pylint.sh
```diff
@@ -22,11 +22,23 @@ then
...
+    python -c 'import autotest'
+    if [ "$?" -ne "0" ]
+    then
+        echo "ERROR: autotest module won't load or"\
+             "AUTOTEST_PATH env. var. is not set"
```
edsantiago: ...also missingspacebetween 'or' and 'AUTOTEST_PATH'.

edsantiago:
Are you seeing the failures I'm seeing from `event/event.py` and `tag/change_registry.py` ?

run_pylint.sh
cevich: ahh, good catches, WTF didn't neither vim or bash flag those, hmmm.

cevich:
Unf. I ran ``vim`` in my development container, and **POW** bash segfaulted and the docker happily honoured the --rm I had given it :scream: 

So I need to recover my nerves for a bit, then I'll do some more test runs with it.

cevich:
> (pylint) W0402: 27,0: : Uses of a deprecated module 'string'

Uggg, I forgot the defaults are not the defaults (for every OS).

cevich:
@edsantiago Okay, good now?

run_pylint.sh
```diff
@@ -73,6 +86,17 @@ record_return() {
     fi
 }
 
+check_print() {
+    PRINTS=$(egrep --count '^\s*print\s+' "$1")
```
edsantiago: The duplication here bothers me. Why not this?

    PRINTS=$(egrep --with-filename --line-number '^\s*print\s+' "$1")
        if [ -n "$PRINTS" ]
            echo "Python print ..."
            echo "$PRINTS"

I've tested it, with multiple prints, and it works.

run_pylint.sh
```diff
@@ -22,11 +22,24 @@ then
...
+    python -c 'import autotest'
+    if [ "$?" -ne "0" ]
+    then
+        echo "ERROR: autotest module won't load or " \
+             "AUTOTEST_PATH env. var. is not set"
+        exit 1
+    fi
+fi
+
+TMPFILENAME="$(mktemp --suffix=run_pylint)"
```
edsantiago: (Very minor): how about `mktemp --tmpdir $(basename $0).XXXXXXX` ? That (1) removes hardcoded run_pylint, (2) puts the script name at the beginning instead of at the end so if there are /tmp droppings -- which there shouldn't be, thanks to `cleanup`, but Murphy's Law -- they are easier to identify. Or if not, how about prepending a dot so the suffix doesn't run-on with the XXXXX?

edsantiago:
Two nits, neither of them a blocker. 
FWIW I never received a notification of your push. Just your comment.

edsantiago:
BTW in your commit message, is "pyhong" deliberate?

cevich:
> is "pyhong" deliberate?

Yes, this is the code name for the latest version which comes with a complimentary keyboard with really really really small keys :grinning:

cevich:
Fixed and fixed and re-pushed.  I'll merge if Travis is happy also.

PR #608, opened 2016-06-03
Adding negativeusage parameters to --tmpfs

afomm:
Added parameters to run negativeusage test with wrong parameters to --tmpfs

### DIFF
--- a/config_defaults/subtests/docker_cli/negativeusage.ini
+++ b/config_defaults/subtests/docker_cli/negativeusage.ini
@@ -1,5 +1,5 @@
 [docker_cli/negativeusage]
-subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, ip1
+subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, iv5, iv6, iv7, ip1
 #: Optional docker subcommand (string). Automatically generated substitution keys:{n}
 #: {n}
 #: {t} * ``%%(FQIN)s`` **-** valid, default test image name{n}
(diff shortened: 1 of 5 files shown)


afomm:
@cevich & @edsantiago, in case you don't get any notification, please review.

edsantiago:
Looking now. (I got the notification; I think it's only git push --force that skips the email).

edsantiago:
Interesting test cases. I suspect those error messages and exit codes might not stay consistent, because at least one of them looks like a golang crash, but I can't currently think of any better tests and won't be looking at this for the next two weeks. FWIW, LGTM at first pass.

afomm:
Thanks Ed. And yes, these errors messages might change but I couldn't think of better ones. Chris & I talked about having one SELinux related but my tests with context="..." is returned as wrong --tmpfs parameter. Let me know if you can think of a good one to add.

cevich:
> one of them looks like a golang crash

That's almost always is a bug (whoo hoo!), is it reproducable?

afomm:
Yes, it is reproducible. Some of the options that show the same System error are: . (dot), ./ (dot backslash), /dev/null.

cevich:
excellent! Please pen a -medium- BZ against RHEL 7.2 with the version numbers and the manual steps that reproduce the problem.  Then go get yourself a brownie :D

cevich:
@afomm Okay, checked out the changes, all LGTM.  Though I can't tell from the config., it seems like maybe there are more bugs here. 

I don't have a setup ready to test these.  Are the errors consistent and sensible given the incorrect values you're passing in?

e.g. the "--tmpfs ." error strikes me as a really strange, but I can't see the complete context.

Happy to go over the debug output with you to make sure.

config_defaults/subtests/docker_cli/negativeusage.ini
```diff
@@ -70,6 +70,24 @@ subarg = -e,PATH=/tmp,%%(FQIN)s,true
...
+subarg = --tmpfs,:rw,%%(FQIN)s
+stderr =  open /proc/self/task/1/attr/exec: no such file or directory
+extcmd = 125
+
+[docker_cli/negativeusage/iv7]
+subcmd = run
+subarg = --tmpfs,.,%%(FQIN)s
+stderr =  Message: Failed to open /dev/null - open /dev/null: permission denied
+extcmd = 125
+
```
cevich: This would also be interesting:
```
...
subcmd = run
subarg = --tmpfs,,%%(FQIN)s
...
```
(empty / Null value passed to --tmpfs)

afomm: @cevich with subarg = --tmpfs,,%%(FQIN)s, "/usr/bin/docker-latest -D run --tmpfs  registry.access.redhat.com/rhel7/rhel:latest" is executed and gives 'docker run --invalid-opt invalid-image'

cevich: good good.  And the exit code is proper and expected?

afomm: Yes, the exit code are expected.

cevich:
New additions LGTM.  Is this PR ready to merge then?

It's okay if the new ``create --tmpfs`` stuff goes into a new PR.

afomm:
Yes, it is ready to merge. And OK, I will make a new PR for create --tmpfs addition.

PR #610, opened 2016-06-15
Adding create subtest with --tmpfs option

afomm:
Adding a create container subtest with --tmpfs option and inspecting the
created container for the --tmpfs parameter

### DIFF
--- a/subtests/docker_cli/create/create_tmpfs.py
+++ b/subtests/docker_cli/create/create_tmpfs.py
@@ -1,0 +1,28 @@
+"""
+Create container with --tmpfs option and inspect it
+"""
+
+from create import create_base
+from dockertest.output import DockerVersion
+
+
+class create_tmpfs(create_base):
+
+    def initialize(self):
+        super(create_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(create_tmpfs, self).run_once()
+        cid = self.get_cid()
+        self.sub_stuff['metadata'] = (
+            self.sub_stuff['cont'].json_by_long_id(cid))
+
+    def postprocess(self):
+        super(create_tmpfs, self).postprocess()
+        tmpfs_arg = self.config['tmpfs_arg']
+        actual_tmpfs = self.sub_stuff['metadata'][0]['HostConfig']['Tmpfs']
+        self.failif(tmpfs_arg not in actual_tmpfs,
+                    "Container was not created with: --tmpfs %s"
+                    % tmpfs_arg)
--- a/config_defaults/subtests/docker_cli/create.ini
+++ b/config_defaults/subtests/docker_cli/create.ini
@@ -1,6 +1,6 @@
 [docker_cli/create]
 subsubtests = create_true,create_false,create_signal,
-              create_remote_tag,create_names
+              create_remote_tag,create_names,create_tmpfs
 #: Extremely short timeout, all these commands should return
 #: almost immediatly.
 docker_timeout = 10
@@ -18,6 +18,11 @@ cmd = /bin/true
 
 [docker_cli/create/create_false]
 cmd = /bin/false
+
+[docker_cli/create/create_tmpfs]
+#: --tmpfs parameter to create container
+tmpfs_arg = /run
+run_options_csv = --tmpfs,%(tmpfs_arg)s
 
 [docker_cli/create/create_names]
 cmd = sleep 2s


cevich:
awesome.  taking a look.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+
+class create_tmpfs(create_base):
+
+    def initialize(self):
+        super(create_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(create_tmpfs, self).run_once()
```
cevich: (unless you were planning to add more) There's no need to define a method that only calls the super-class method.  That happens automatic by **not** defining a method in the sub-class.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def postprocess(self):
+        # Inspect container for --tmpfs
+        subargs = ["--format", "'{{.HostConfig.Tmpfs}}'",
```
cevich: Not sure you need the double and single-quotes here, though python may complain about the curly-braces (escape them with ``\`` if so).  Argument-passing in a list works differently in this respect than a shell.  In act, the list-format is the same in C/C++ and other languages as well.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+    def initialize(self):
+        super(create_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(create_tmpfs, self).run_once()
+
+    def postprocess(self):
+        # Inspect container for --tmpfs
```
cevich: A good comment explains "why" you are doing what follows, not "what".  We can tell "what" just by reading the code :smile:

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        subargs = ["--format", "'{{.HostConfig.Tmpfs}}'",
+                   self.sub_stuff['name']]
+        dkrcmd = DockerCmd(self, "inspect", subargs)
+        self.sub_stuff['dkrcmd'] = dkrcmd
+        self.sub_stuff['cmdresult'] = mustpass(dkrcmd.execute())
```
cevich: Looks good.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        tmpfs_arg = self.config['tmpfs_arg']
+        with_tmpfs = self.sub_stuff['dkrcmd'].stdout.find(tmpfs_arg) != -1
+        self.failif(not with_tmpfs,
+                    "Container was not created with: --tmpfs %s"
+                    % tmpfs_arg)
+        # Change inspect paramerter to get CID
+        subargs = ["--format", "'{{.Id}}'", self.sub_stuff['name']]
+        dkrcmd = DockerCmd(self, "inspect", subargs)
+        self.sub_stuff['dkrcmd'] = dkrcmd
+        self.sub_stuff['cmdresult'] = mustpass(dkrcmd.execute())
```
cevich: I think would be better to not re-use both the ``dkrcmd`` instance _and_ the execution result, pick one or neither.  In fact, there's a ``cmdresult`` built-in to the ``dkrcmd`` instance already, so you don't even really need to store it separately - similar to how ``create_base.postprocess()`` does it.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,38 @@
...
+        super(create_tmpfs, self).postprocess()
```
cevich: Nearly always (in a destructor is virtually the only exception) methods call the super-class method first.  This would be a great place to add a comment, explaining why it's done last (I can guess, but better to have it in a comment).

cevich: IIUC, this super call will fail in it's call to ``cid = self.get_cid()``.  It's fine to either omit the super call, or override ``get_cid()`` in this class (if you do want to call the super-class method).

cevich:
@afomm this is really great, nice easy to read, clean and compact code.  I can tell right away that your skill is improving.  I shared a few comments, only one major problem, the rest are mostly just minor style/commenting issues.  Looking forward to your second version, this one is really close!

afomm:
@cevich & @edsantiago, in case you didn't get a notification, please take a look.

cevich:
I did not, thanks for the mention, I'll take another look.

edsantiago:
You're clearly getting the hang of things, and this is much better work, but some of it still made me uncomfortable: the unexplained overriding of `get_cid()`, the wrong thing within `try`, the loss of the original `dkrcmd`. The length. The confusing double-single quotes in the docker inspect. This is the sort of situation that calls for taking a big step back to evaluate. I did so, and think it would be much cleaner to do something like:

    def run_once(...):
        super(...)
        cid = self.get_cid()  # original get_cid; you can/should delete the override version
        self.sub_stuff['metadata'] = self.sub_stuff['cont'].json_by_long_id(cid)

    def postprocess(...):
        super(...)
        tmpfs_arg = ...
        actual_tmpfs = self.sub_stuff['metadata'][0]['HostConfig']['Tmpfs']
        self.failif(tmpfs_arg not in actual_tmpfs, ...)

That's a lot fewer lines, and doesn't require the uncomfortable delayed-super in `postprocess`. I've tested it and it works as expected.

afomm:
OK. Thanks Ed.

afomm:
@cevich & @edsantiago, applied Ed's feedback and here is an update.

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def initialize(self):
+        super(create_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(create_tmpfs, self).run_once()
+        cid = self.get_cid()
+        cont_long_id = self.sub_stuff['cont'].json_by_long_id(cid)
```
edsantiago: Much better! But I don't understand this variable name. `json_by_long_id` returns a dict, parsed from the json returned by `docker inspect`. I don't think the `long_id` part is meaningful to the contents of the data structure or the variable. I think a better name would reflect the fact that this is docker inspect contents. But I may have missed something; please let me know if you have a good reason for this var name.

afomm: If I understand your question correctly, I was trying to avoid using this long line "self.sub_stuff['metadata'] = self.sub_stuff['cont'].json_by_long_id(cid)". Would it be better to do:
self.sub_stuff['metadata'] = (
            self.sub_stuff['cont'].json_by_long_id(cid))

edsantiago: Using a temporary variable to avoid line wrap is fine. My question was about the variable's name: `cont_long_id` is a weird name. To me it doesn't indicate the variable's purpose or content. I actually find it more confusing than `x` or even `foo`: I see the name `cont_long_id` and my first thought is that it contains an id, or a container, or at least _something_ relating to `cont` or `long` or `id`. But AFAICT it doesn't, which means it will confuse the next maintenance programmer too.

afomm: Got it & thanks! Would something like cont_metadata or cont_info have been a better? For now, I will avoid the variable & just wrap the line.

edsantiago: Yes, either of those would be better.

I realize this is a fine point but it's an important one: our job as programmers is to write code that someone, some day, will find easy to understand. Good variable and function names will help that future person.

edsantiago:
LGTM!

cevich:
Jeepers, @edsantiago throws a ton of comments and LGTM before I can respond to any of them.  Nice teamwork guys.  Taking a quick peek...

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+    def initialize(self):
+        super(create_tmpfs, self).initialize()
+        # Minimum docker version 1.10 is required for --tmpfs
+        DockerVersion().require_server("1.10")
+
+    def run_once(self):
+        super(create_tmpfs, self).run_once()
+        cid = self.get_cid()
+        self.sub_stuff['metadata'] = (
+            self.sub_stuff['cont'].json_by_long_id(cid))
```
cevich: This is fine here, though for more complex/long cases, keep Ed's suggestion in mind:  

> Using a temporary variable to avoid line wrap is fine

Names and references in python (generally) tend to be on the long-side (compared to other languages).  It's totally cool to use other temporaries to hold values that are immediately re-assigned.  Python will most likely just optimize them away, but they do add a bit to readability.  Really long call references can be shortened in a similar way, e.g.:

```python
_blah = self.other.some['really'].long.reference['to']['a'].method_or_function
result = _blah(parameters)
```

subtests/docker_cli/create/create_tmpfs.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def postprocess(self):
+        super(create_tmpfs, self).postprocess()
+        tmpfs_arg = self.config['tmpfs_arg']
+        actual_tmpfs = self.sub_stuff['metadata'][0]['HostConfig']['Tmpfs']
+        self.failif(tmpfs_arg not in actual_tmpfs,
```
cevich: __Gotcha__: This will match if ``actual_tmpfs`` **contains** ``tmpfs_arg`` anywhere.  e.g. ``"/run" not in "/run_for_the_hills" == False``.  Though I think extremely unlikely to every happen.  Though ``tmpfs_arg`` is configuration, and the test can't ever predict what it will contain.  I'm being overly pedantic though, just for example's sake.  What you have is fine.

cevich:
I always write things way more complicated / verbose the first few times, before condensing them down.  Perfectly normal to work that way.  Really nice job here, merging...

subtests/docker_cli/create/create_tmpfs.py
edsantiago: Very good observation, but:

"Then why are you smiling?"

"Because I know something you do not know: _I_ am not left-handed". Whoops, no wait, I mean: `actual_tmpfs` is not a string, it's a dict:

    {u'/run': u''}

cevich: lolz.   I have spent a lifetime developing an immunity to...damnit, your right.  Okay, nothing to see here, move along, move along.

PR #611, opened 2016-06-30
(minor) login test: cleanup: fix exception

edsantiago:
Not all subsubtests create a login auth file. In cleanup,
check before removing a nonexistent file.


cevich:
Ahh okay, so key is only set in sub_stuff if file is preserved, so it's okay to skip restoring it if it's not there.  Great, LGTM.

PR #615, opened 2016-07-14
github issue #572: default omit_empty=True in get_as_list()

edsantiago:
https://github.com/autotest/autotest-docker/issues/572

The omit_empty=False default caused another problem in
ADEPT tests this week. I can think of no sensible reason
for passing back empty values in the usual case.

### DIFF
--- a/dockertest/config.py
+++ b/dockertest/config.py
@@ -482,7 +482,7 @@ class Config(dict):
         return the_copy
 
 
-def get_as_list(value, sep=",", omit_empty=False):
+def get_as_list(value, sep=",", omit_empty=True):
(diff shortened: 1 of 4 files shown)


edsantiago:
Full run in progress now; I'll check back in tomorrow.

cevich:
Heyhey look at that, even CI is happy.  Okay let me see what I can do to break it :wink:

edsantiago:
Oh dear. I don't think this was ready for merge: there's a problem with the run_volumes test, and I'm trying to find a clean solution. My fault for not following up earlier this morning, sorry!

PR #616, opened 2016-07-15
Treat DockerTestNAError as PASS

edsantiago:
When summarizing subtest results, treat DockerTestNAError
as a passing subsubtest.

FIXME: this could potentially mask real problems, such as
run_volumes skipping because of a wrong config setting or
one of the expecting-docker-1.10 tests skipping even though
the test system should have 1.10. I don't have any good ideas
on how to address this.

### DIFF
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -18,7 +18,7 @@ import imp
 import sys
 import copy
 from ConfigParser import Error
-from autotest.client.shared.error import TestError
+from autotest.client.shared.error import TestError, TestNAError
 from autotest.client.shared.version import get_version
 from autotest.client import test
 import version
@@ -410,7 +410,11 @@ class SubSubtestCaller(Subtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
-            # cleanup() will still be called before this propigates
+            # Treat N/A as passed, even though this may hide obscure failures
+            if isinstance(detail, TestNAError):
+                self.logwarning("Treating TestNAError as PASS")
+                self.final_subsubtests.add(name)
+            # cleanup() will still be called before this propagates
             raise exc_info[0], exc_info[1], exc_info[2]
 
     def run_all_stages(self, name, subsubtest):


cevich:
>  I don't have any good ideas on how to address this.

This was my idea about just having a config. variable, like 'enable_paranoia_mode = True'.  Then at least it's an option to switch it on/off.  In fact, maybe when paranoia mode is enabled, __all__ TestNAError's are raised as TestFail's instead?  Anyway, just an idea.

dockertest/subtest.py
```diff
@@ -410,7 +410,11 @@ def try_all_stages(self, name, subsubtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
-            # cleanup() will still be called before this propigates
+            # Treat N/A as passed, even though this may hide obscure failures
+            if isinstance(detail, DockerTestNAError):
```
cevich: Maybe ``isinstance(detail, TestNAError)`` would be slightly more general?  It's a small thing.

dockertest/subtest.py
```diff
@@ -410,7 +410,11 @@ def try_all_stages(self, name, subsubtest):
...
+                self.logdebug("Treating DockerTestNAError as PASS")
+                self.final_subsubtests.add(name)
```
cevich: Ahh ok, good, ya nice to have that message, it also saves needing a comment :smile: (for the ``final_subsubtests.add()``)

edsantiago:
New force push; addresses your TestNAError comments; and I thought I updated the commit message, and it shows as updated under the Commits tab, but it's not updated on the Conversation tab.

dockertest/subtest.py
```diff
@@ -410,7 +410,11 @@ def try_all_stages(self, name, subsubtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
-            # cleanup() will still be called before this propigates
+            # Treat N/A as passed, even though this may hide obscure failures
+            if isinstance(detail, TestNAError):
+                self.logdebug("Treating TestNAError as PASS")
```
cevich: Do you think a ``self.logwarning()`` would be safer / less obscure?

cevich:
okay, LGTM.

dockertest/subtest.py
edsantiago: Good call. So done.

PR #617, opened 2016-07-19
run_sigproxy test: race condition fix

edsantiago:
...at least, it's acting like a race condition (unpredictable
failures), even though on examination it looks like a pretty
clear bug that should fail every time.

The problem seems to be that some containers are run with --detach,
hence we don't see READY in stdout. Solution: use 'docker logs' in
addition to checking stdout. But why do the tests pass sometimes??


cevich:
LGTM.

PR #619, opened 2016-08-03
failif_not_in() : allow multiple needles

edsantiago:
failif_not_in(needle, haystack) now allows needle to be
a pipe-separated list of possibilities. The idea is to
allow support for multiple docker versions:

   foo.ini:
   expected_stderr = this is what docker-1.10 says
                   | docker-1.12 says something completely different

Since it's getting complicated, add new unit tests.

Also: run_unittests wrapper: bug fix: we were ignoring error
exit status from the first test to exit.

### DIFF
--- a/dockertest/subtestbase.py
+++ b/dockertest/subtestbase.py
@@ -146,7 +146,9 @@ class SubBase(object):
         being contained in a larger string, e.g. to look for XYZ in a
         command's stdout/stderr.
 
-        :param needle: the string you're looking for
+        :param needle: the string you're looking for. May be a list of
+                       multiple strings separated by '|' and optional
+                       whitespace, e.g. 'needle | fiddle | doo wah diddy'
         :param haystack: the actual string, e.g stdout results from a command
         :param description: description of haystack, e.g. 'stdout from foo'
         :raise DockerTestFail: if needle is not found in haystack
@@ -155,8 +157,15 @@ class SubBase(object):
             description = 'string'
         if needle in haystack:
             return
-        raise DockerTestFail("Expected string '%s' not in %s '%s'"
-                             % (needle, description, haystack))
+        expect_s = "Expected string '%s'" % needle
+        if '|' in needle:
+            needles = [n.strip() for n in needle.split('|')]
+            expect_s = "Expected strings %s" % needles
+            for subneedle in needles:
+                if subneedle.strip() in haystack:
+                    return
+        raise DockerTestFail("%s not in %s '%s'" %
+                             (expect_s, description, haystack))
 
     @classmethod
     def log_x(cls, lvl, msg, *args):
--- a/dockertest/subtestbase_unittests.py
+++ b/dockertest/subtestbase_unittests.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+
+# Pylint runs from a different directory, it's fine to import this way
+# pylint: disable=W0403
+
+import os
+import sys
+import types
+from unittest2 import TestCase, main
+
+
+###############################################################################
+# BEGIN boilerplate crap needed for subtests, which should be refactored
+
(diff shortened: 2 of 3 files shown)


dockertest/subtestbase.py
```diff
@@ -155,8 +157,15 @@ def failif_not_in(needle, haystack, description=None):
...
-                             % (needle, description, haystack))
+        expect_s = "Expected string '%s'" % needle
+        if '|' in needle:
+            needles = [n.strip() for n in needle.split('|')]
+            expect_s = "Expected strings %s" % needles
+            for subneedle in needles:
+                if subneedle.strip() in haystack:
+                    return
+        raise DockerTestFail("%s not in %s '%s'" %
+                             (expect_s, description, haystack))
```
cevich: Can we say with confidence that the '|' character will never be needed as part of the search string?  If not, then an escaping-mechanism will be needed, along with escapes for the escape character as well.  An easier implementation maybe to add a ``delimiter="|"`` parameter, which will work with existing code, yet allow the multiple-strings feature as well.  It would also permit disabling the feature with e.g. ``delimiter = None``.

dockertest/subtestbase.py
```diff
@@ -155,8 +157,15 @@ def failif_not_in(needle, haystack, description=None):
...
         if needle in haystack:
             return
-        raise DockerTestFail("Expected string '%s' not in %s '%s'"
-                             % (needle, description, haystack))
+        expect_s = "Expected string '%s'" % needle
+        if '|' in needle:
+            needles = [n.strip() for n in needle.split('|')]
+            expect_s = "Expected strings %s" % needles
+            for subneedle in needles:
+                if subneedle.strip() in haystack:
```
cevich: Possible nit:  There are some rare occasions where searching for a leading or trailing space is handy.  Not a big deal though, and it could be worked around if the feature can be disabled (see comment for line 168)

cevich:
One comment worth considering, otherwise LGTM.

dockertest/subtestbase.py
edsantiago: >Can we say with confidence that the '|' character will never be needed as part of the search string?

No, but I'm reluctant to add complexity for what seems like a low-probability risk. If this code becomes wildly popular, with millions of users worldwide, someone will inevitably find a case where they need '|', and it will fail, and they will curse me. For our small limited-case purposes, a handful of developers on a tiny scale, I can't justify the code complexity.

cevich: How about just adding a ``delimiter="|"`` parameter then?  That seems like a very small complexity increase that at least gives an interface to workaround related issues.  Though in general I agree with your statement, there isn't much justification for lots of extra complexity.  So I'm okay if you just want to merge this and be done with it.

edsantiago: >How about just adding a delimiter="|" parameter then? 

The problem is, that has to be done in the code. But the place for defining a delimiter is in the .ini file. The code shouldn't (and, realistically, cannot) know what a suitable delimiter is for a given test. So we end up with something like:

    expected_stderr_delimiter = '*'
    expected_stderr = a|b * b|c * lmnop

    expected_stderr = self.config['expected_stderr']
    expected_delimiter = '|'
    if 'expected_stderr_delimiter' in self.config
         expected_delimiter = self.config['expected_stderr_delimiter']
    self.failif_etc(...delimiter=expected_delimiter)

...and having to do that in Every Single Call to self-failif_not_in()

cevich: > ...and having to do that in Every Single Call to self-failif_not_in()

It wouldn't __have__ to be a configuration option, just documented for those cases should be good enough.  Either way, this can easily be added in the future if it's really necessary, so let's just call it "good'nuf" then.

PR #620, opened 2016-08-03
rhel_push_plugin test

edsantiago:
rhel_push_plugin tries to prevent docker push of a RHEL-based
image, where "RHEL-based" seems to be defined (via hardcoding
in the plugin) as:

    for each layer:
       run docker inspect
           True if Vendor is "Red Hat, Inc." and
                   Name starts with "rhel"
       otherwise False

Test by building images with variations of the above, and
making sure docker push is blocked (with the expected message)
when RHEL-based and allowed otherwise.

### DIFF
--- a/config_defaults/subtests/redhat/rhel_push_plugin.ini
+++ b/config_defaults/subtests/redhat/rhel_push_plugin.ini
@@ -1,0 +1,37 @@
+[redhat/rhel_push_plugin]
+subsubtests = push_blocked, push_vendor_ok, push_name_ok, push_registry_ok
+#: base image from which to build the image we try to push. We can't
+#: default to full_name_from_defaults, because that may be RHEL itself.
+#: We could use CentOS, but that would mean pulling it if not already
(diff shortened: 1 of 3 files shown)


cevich:
Code looks good.  Since this feature is plugin/implementation specific, and (IMHO) there are likely to be more of these going forward.  What do you think about separating it from the more generic subtest-crowd?  e.g. ``subtests/rhel/push_plugin``  (or something to that effect).

edsantiago:
>What do you think about separating it from the more generic subtest-crowd? e.g. `subtests/rhel/push_plugin`

I didn't know I could do that. I'll look into it, thanks.

cevich:
#621 could also fall into this category.  Maybe instead of ``rhel`` call it ``subtests/red_hat/`` and then it can include CentOS and Fedora as well.  It would help with shortening names, reducing number of items in ``docker_cli``, and make it clear in testing results that the tests are very environment-specific.

edsantiago:
Thanks for the suggestion! Done and tested.

PR #621, opened 2016-08-05
New meta-test: confirm debug symbols

edsantiago:
bz1280068 - make sure docker binary ships with debuginfo

### DIFF
--- a/config_defaults/subtests/redhat/packaging.ini
+++ b/config_defaults/subtests/redhat/packaging.ini
@@ -1,0 +1,2 @@
+[redhat/packaging]
+subsubtests = debuginfo_present
--- a/subtests/redhat/packaging/packaging.py
+++ b/subtests/redhat/packaging/packaging.py
@@ -1,0 +1,58 @@
+r"""
+Summary
+-------
+
+Meta-tests: docker packaging, installation, setup. Nothing here
+actually runs docker.
+
+Operational Summary
+-------------------
+
+#. Run 'file' against docker binaries, make sure output includes 'not stripped'
+"""
+
+import os
+import magic
+from dockertest import subtest
+from dockertest.subtest import SubSubtest
+
+
+class packaging(subtest.SubSubtestCaller):
+
+    """ SubSubtest caller """
+
+
+class packaging_base(SubSubtest):
+    pass
+
+
+class debuginfo_present(packaging_base):
+    """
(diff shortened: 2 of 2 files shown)


subtests/docker_cli/meta/meta.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def run_once(self):
+        super(debuginfo_present, self).run_once()
+        found_binary = False
+        # Another way to search for DWARF might be:
+        #    readelf -S /usr/bin/docker-current |grep -i debug_info
+        for binfile in ['docker-current', 'docker-latest']:
+            path = os.path.join('/usr/bin', binfile)
+            if os.path.exists(path):
+                found_binary = True
```
cevich: Second time through the loop (``binfile == "docker-latest"``), ``found_binary`` will overwrite the prior iteration value.  i.e. ``docker-current`` is not found but ``docker-latest`` is found.  Probably okay to just get rid of the ``os.path.exists()`` check and allow ``self.sub_stuff['magic'].file(path)`` to fail.  Disadvantage is that failure on the first item would prevent testing of the second.

An alternative idea (to always test both) could be to use test-iterations as list indexes into the binfile names.  IIRC, all you need to do is set a class attribute ``iterations = 2``, and another for ``binfiles = ['docker-current', 'docker-latest']``.  Then, in your ``run_once()``, set ``binfile = self.binfiles[self.iteration - 1]`` (or something like that).

cevich:
@edsantiago found one "gotcha", otherwise test looks fine.  Maybe this could be another candidate for a less-general subtest namespce, since it's RPM/build specific?

subtests/docker_cli/meta/meta.py
edsantiago: The purpose was to make sure that we found at least one docker binary. Maybe I should've named the variable `found_any_binary`. But on reflection I'm not sure it actually serves any useful purpose; let me consider removing it.

cevich: >  found at least one docker binary

Oh okay, that matches the code better.  If it's an important consideration, then maybe just add a ``break`` at the end of the ``os.path.exists(path)`` branch.

edsantiago:
Done - moved into subtests/redhat along with rhel_push_plugin

cevich:
Awesome, looks great.

PR #622, opened 2016-08-12
refactor: make wait_for_ready() reusable

edsantiago:
Other subtests (e.g. kill and possibly run_cidfile) are
experiencing spurious failures that look suspiciously like
the race condition that prompted wait_for_ready() in
run_sigproxy, i.e. we may be starting tests on containers
that aren't fully running yet.

This moves wait_for_ready() into a method of AsyncDockerCmd.
It's kind of complicated because we need to determine our
own container's ID, and that's not trivial or robust, but
for our purposes I think it's safe enough.

There is no corresponding method for DockerCmd, because it
only makes sense if invoked with --detach, and even in that
case there's no sane way to get the container ID.

In order to keep things simple, this is purely a refactor
with no extra functionality. If this code is acceptable, I'll
follow up with PRs for other failing tests.

### DIFF
--- a/dockertest/dockercmd.py
+++ b/dockertest/dockercmd.py
@@ -10,7 +10,7 @@ import time
 from autotest.client import utils
 from subtestbase import SubBase
 from xceptions import DockerNotImplementedError
-from xceptions import DockerTestError
+from xceptions import DockerExecError, DockerTestError
 
 
 class DockerCmdBase(object):
@@ -334,6 +334,49 @@ class AsyncDockerCmd(DockerCmdBase):
                                          stdin=stdin, close_fds=True)
         return self.cmdresult
 
+    def wait_for_ready(self, timeout=None, timestep=0.2):
+        """
+        Monitor the output of a container (including docker logs, in
+        case stdout is detached), waiting for the string 'READY'. Return
+        as soon as we see it. If we don't, throw a meaningful exception.
+
+        :raises DockerExecError: on timeout.
+        """
+        end_time = time.time() + timeout
+        while time.time() <= end_time:
+            time.sleep(timestep)
+            if 'READY' in self.stdout:
+                return
+            # Also check docker logs
+            cid = self.container_id
+            if cid is not None:
+                logs = DockerCmd(self.subtest, 'logs', [self.container_id])
+                logs.execute()
+                if 'READY' in logs.stdout:
+                    return
+
+        # Never saw READY. Did container exit? If so, help user understand why
+        if self.done:
+            msg = "Container exited before READY"
+            if self.exit_status == 0:
+                msg += " (normal exit status)"
+            else:
+                msg += "; exit status = %d" % self.exit_status
+            stderr = self.stderr
+            if stderr:
+                msg += "; stderr='%s'" % stderr
+            raise DockerExecError(msg)
+
+        # Container still running. Must be a timeout.
+        msg = "Timed out waiting for container READY"
+        stdout = self.stdout
+        if stdout:
+            # Juuuuuuust in case it appeared since we last checked
+            if 'READY' in stdout:
+                return
+            msg += "; stdout='%s'" % stdout
+        raise DockerExecError(msg)
+
     def wait(self, timeout=None):
         """
         Return CmdResult after waiting for process to end or timeout
@@ -378,6 +421,27 @@ class AsyncDockerCmd(DockerCmdBase):
             raise DockerTestError("Attempted to get pid before execute()"
                                   " called.")
         return self._async_job.sp.pid
+
+    @property
+    def container_id(self):
+        """
+        Try to discover our own container ID or name. Return None if we can't.
+        """
+
+        # The simple case: if we are a docker attach command, assume that
+        # subargs are zero or more flags plus a container ID or name.
+        if self.subcmd == 'attach':
+            return self.subargs[-1]
+
+        # Non-attach command. Find our PID, iterate over all containers,
+        # get their PID (via inspect). If we find a match, return the CID.
+        pid = self.process_id
+        cids = utils.run('docker ps -q', verbose=False).stdout.splitlines()
+        for cid in cids:
+            c_pid = utils.run('docker inspect --format {{.State.Pid}} ' + cid)
+            if int(c_pid.stdout) == int(pid):
+                return cid
+        return None
 
     # Override base-class property methods to give up-to-second details
 
--- a/dockertest/dockercmd.py
+++ b/dockertest/dockercmd.py
@@ -334,7 +334,7 @@ class AsyncDockerCmd(DockerCmdBase):
                                          stdin=stdin, close_fds=True)
         return self.cmdresult
 
-    def wait_for_ready(self, timeout=None, timestep=0.2):
+    def wait_for_ready(self, cid=None, timeout=None, timestep=0.2):
         """
         Monitor the output of a container (including docker logs, in
         case stdout is detached), waiting for the string 'READY'. Return
@@ -348,9 +348,10 @@ class AsyncDockerCmd(DockerCmdBase):
             if 'READY' in self.stdout:
                 return
             # Also check docker logs
-            cid = self.container_id
+            if cid is None:
+                cid = self.container_id
             if cid is not None:
-                logs = DockerCmd(self.subtest, 'logs', [self.container_id])
+                logs = DockerCmd(self.subtest, 'logs', [cid])
                 logs.execute()
                 if 'READY' in logs.stdout:
                     return
--- a/dockertest/dockercmd.py
+++ b/dockertest/dockercmd.py
@@ -345,7 +345,8 @@ class AsyncDockerCmd(DockerCmdBase):
         end_time = time.time() + timeout
         while time.time() <= end_time:
             time.sleep(timestep)
-            if 'READY' in self.stdout:
+            stdout = self.stdout
+            if 'READY' in stdout:
                 return
             # Also check docker logs
             if cid is None:
@@ -353,7 +354,8 @@ class AsyncDockerCmd(DockerCmdBase):
             if cid is not None:
                 logs = DockerCmd(self.subtest, 'logs', [cid])
                 logs.execute()
-                if 'READY' in logs.stdout:
+                stdout = logs.stdout
+                if 'READY' in stdout:
                     return
 
         # Never saw READY. Did container exit? If so, help user understand why
@@ -370,11 +372,7 @@ class AsyncDockerCmd(DockerCmdBase):
 
         # Container still running. Must be a timeout.
         msg = "Timed out waiting for container READY"
-        stdout = self.stdout
         if stdout:
-            # Juuuuuuust in case it appeared since we last checked
-            if 'READY' in stdout:
-                return
             msg += "; stdout='%s'" % stdout
         raise DockerExecError(msg)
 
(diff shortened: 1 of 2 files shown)


cevich:
Interesting, ya I've seen the test-before-started problem before, super annoying.  Let's see whatchagothere...

dockertest/dockercmd.py
```diff
@@ -334,6 +334,49 @@ def execute(self, stdin=None):
...
+            time.sleep(timestep)
+            if 'READY' in self.stdout:
+                return
+            # Also check docker logs
+            cid = self.container_id
+            if cid is not None:
+                logs = DockerCmd(self.subtest, 'logs', [self.container_id])
+                logs.execute()
+                if 'READY' in logs.stdout:
+                    return
```
cevich: Suggestion:  IIUC, doesn't stdout always go into the logs w/n it's detached?  Maybe it's good enough to only check the logs.

dockertest/dockercmd.py
```diff
@@ -334,6 +334,49 @@ def execute(self, stdin=None):
...
+        """
+        end_time = time.time() + timeout
+        while time.time() <= end_time:
+            time.sleep(timestep)
+            if 'READY' in self.stdout:
+                return
+            # Also check docker logs
+            cid = self.container_id
+            if cid is not None:
+                logs = DockerCmd(self.subtest, 'logs', [self.container_id])
```
cevich: I realize your going for 'simple' here.  Just for consideration- Maybe have the method accept an instance as a parameter, and check ``instance.stdout`` here instead of creating a DockerCmd.  Reason is, it adds a LOT of flexibility to the exit condition while only moving one line of code from here to the caller.

dockertest/dockercmd.py
```diff
@@ -334,6 +334,49 @@ def execute(self, stdin=None):
...
+                logs.execute()
+                if 'READY' in logs.stdout:
```
cevich: Here too.  But less important.  It's nice to give control to the caller by allowing flexibility in the search string.  Not-required, just a suggestion.

dockertest/dockercmd.py
```diff
@@ -334,6 +334,49 @@ def execute(self, stdin=None):
...
+                    return
+
+        # Never saw READY. Did container exit? If so, help user understand why
+        if self.done:
```
cevich: Consider moving this check right to the top.  It will help prevent TOCTU races in the caller code.

dockertest/dockercmd.py
edsantiago: I _think_ checking docker logs would be enough; I just wanted to avoid running it unless absolutely necessary. That may be premature optimization, though. I'll rethink it.

dockertest/dockercmd.py
edsantiago: Interesting idea. I'll test it out, thank you.

dockertest/dockercmd.py
```diff
@@ -334,6 +334,49 @@ def execute(self, stdin=None):
...
+                msg += "; stderr='%s'" % stderr
+            raise DockerExecError(msg)
+
+        # Container still running. Must be a timeout.
+        msg = "Timed out waiting for container READY"
+        stdout = self.stdout
+        if stdout:
+            # Juuuuuuust in case it appeared since we last checked
+            if 'READY' in stdout:
+                return
```
cevich: This doesn't seem right to me.  It presumes the caller cares more about the output than the timeout, which may not be the case.  Otherwise, why stop at double-checking just once?  It's equally likely the string could appear in the microseconds between line 377 and 378.  IMHO, better to ``while time.time() < end_time:`` (since the code is counting "0" seconds elapsed) and leave the double-checking out.  The caller can do it if it's really necessary.

dockertest/dockercmd.py
edsantiago: I considered that for a long time, and decided that it was currently unnecessary. If such flexibility is needed in the future, it's easy to add. And if we don't need it, it's much less confusing to leave it simple and unambiguous.

dockertest/dockercmd.py
```diff
@@ -379,6 +422,27 @@ def process_id(self):
...
+    @property
+    def container_id(self):
+        """
+        Try to discover our own container ID or name. Return None if we can't.
+        """
+
+        # The simple case: if we are a docker attach command, assume that
+        # subargs are zero or more flags plus a container ID or name.
+        if self.subcmd == 'attach':
+            return self.subargs[-1]
```
cevich: Nice!

dockertest/dockercmd.py
edsantiago: I need to think about this. The function is `wait_for_ready()`. It isn't `wait_for_ready_and_make_sure_the_container_is_still_running()`. All we do is check for the string READY. At present the only invocations also expect the container to continue running, but that may not always be the case. I will add `and not self.done` to the while loop, though, to avoid wasting time.

dockertest/dockercmd.py
```diff
@@ -379,6 +422,27 @@ def process_id(self):
...
+        # Non-attach command. Find our PID, iterate over all containers,
+        # get their PID (via inspect). If we find a match, return the CID.
+        pid = self.process_id
+        cids = utils.run('docker ps -q', verbose=False).stdout.splitlines()
+        for cid in cids:
+            c_pid = utils.run('docker inspect --format {{.State.Pid}} ' + cid)
+            if int(c_pid.stdout) == int(pid):
+                return cid
+        return None
+
```
cevich: Hmmmm, I don't think the CID can ever change in the context of this class (I don't remember if we allow re-use after the container exits).  Consider caching the found CID in a private instance variable.  The caller may be referencing this property inside a loop (in fact ``wait_for_ready()``) does this.  Spawning processes is "expensive" and if the system is under particularly high load, can lead to all sorts of problems (depending on how tight the loop is and how quickly the ``utils.run()`` returns).

dockertest/dockercmd.py
```diff
@@ -379,6 +422,27 @@ def process_id(self):
                                   " called.")
         return self._async_job.sp.pid
 
+    @property
+    def container_id(self):
+        """
+        Try to discover our own container ID or name. Return None if we can't.
+        """
```
cevich: Optional idea:  Maybe add a 'cidfile=None' parameter.  When non-None, if the file exists, read it and return.  This will be much faster than going searching for the CID through the process tree (assuming there are a huge number of processes/containers to search).

Another idea: Optional parameter for the container's name.  Looking up the CID by search of a ``dockertest.DockerContainers.list_containers()`` list is much faster than a potential boat-load of ``docker inspect`` calls.

subtests/docker_cli/run_sigproxy/run_sigproxy.py
```diff
@@ -81,25 +81,6 @@ def init_test_specific_variables(self):
...
-            if 'READY' in container_cmd.stdout:
-                return
-            # Some containers run detached; use 'docker logs' to check output
-            logs = DockerCmd(self, 'logs', [self.sub_stuff['container_name']])
-            logs.execute()
-            if 'READY' in logs.stdout:
-                return
-        self.failif(container_cmd.done, "Container exited before ready")
-        raise DockerTestFail("timed out waiting for container READY")
-
```
cevich: good riddance!

dockertest/dockercmd.py
cevich: ...also in a great number of tests, the calling code will already know the container's name.

cevich:
Okay @edsantiago gave a more through read-through.  Made a few suggestions.  I think really the only big nit's for me are the double-checking for READY and caching of CID, the rest are just suggestions.  In any case, I think this is a great addition with potential for wide-spread use.  Nice work.

dockertest/dockercmd.py
cevich: Ahh, good idea, I like it.

dockertest/dockercmd.py
cevich: Okay, no problem.

dockertest/dockercmd.py
cevich: Actually the extra DockerCmd call is probably much slower than checking seld.stdout.  I was mostly thinking of helping to keep the code simple.  In any case, a caller supplied data-source would also help make the method simpler.

cevich:
BTW: I don't think it's worthwhile to go nuts with adding a unittest for this.  If you have a mind-to, go for it, otherwise I know writing unittests for time/sequence sensitive stuff like this is a royal PITA.  It's okay to just leave the subtest as the unittest IMHO.

dockertest/dockercmd.py
edsantiago: OK. Can't check `instance.stdout` because half the docker commands are run with `--detach`. But see later notes for a different approach.

edsantiago: Function signature changed to accept a cid or name; this lets us bypass calling `container_id`, but I'm leaving the function in anyway because the cid param is optional.

dockertest/dockercmd.py
edsantiago: The reason for this is that the `self.stdout` grabbed in line 372 is not the same as the one checked earlier in the function. This one is being displayed to the user as part of an error message, and I've seen (on two occasions) moronic error messages that tell me "the expected string 'X' is not in 'XYZ'", and it's obvious that X **is** in XYZ, but the string must've appeared in the small window between checking and producing the error message.

I know it's duplication, and it's horrible, but without a way to preserve the last-checked stdout I think this is necessary. The alternative is to completely remove the quoted stdout, just telling the reader "tough luck, figure it out on your own". And I think that's worse than duplication. (Which takes a lot for me to say.)

dockertest/dockercmd.py
edsantiago: New signature passes in cid or name.

edsantiago:
re: unit tests: I'd love to but it's too docker- and timing-dependent. I couldn't think of a sane way to test it.

dockertest/dockercmd.py
cevich: Okay, I think I follow.  This will also help if the timeout used is so short, that ``stdout`` isn't actually ever checked.  

> preserve the last-checked stdout 

Having actually written something that solves that, I know it's non-trivial.  Still, un-helpful error messages from test code, shouldn't be solved in generic code (as annoying as they are).  That just moves the problem and introduces a case where it's possible to exceede the timeout, but still pass the test  I agree though that tracking examined lines isn't easy, and duplicating may be the best option.

What about making docker do the 'already seen' legwork by tracking the timestamps (since logging a line requires time):

```python
first_timestamp = last_timestamp = now()  # there's a function for this somewhere
n_lines=0
...
while time.time() <= end_time:
    ...
        logs = DockerCmd(self.subtest, 'logs', [, "--timestamps", "--since", last_timestamp, "--tail", "1", self.container_id])
        if logs:
            last_timestamp = # parse it out
            n_lines += 1
```

Then to produce a helpful error message, run ``docker logs --since=%{first_timestamp} --tail=%{n_lines}``.

or something to that effect.  Could that work?

I'm okay with the duplication, my main fear is creating of a API method with a kind of fuzzy-timeout.  A seriously loaded-down system can produce a lot of fuzz :grimacing: 

OTOH, if it's going to block you for more than a few hours, forget it, we can just merge this as-is.

cevich:
> re: unit tests: 

Yep, I Totally understand.  No worries.

dockertest/dockercmd.py
```diff
@@ -345,15 +345,17 @@ def wait_for_ready(self, cid=None, timeout=None, timestep=0.2):
...
             # Also check docker logs
             if cid is None:
                 cid = self.container_id
             if cid is not None:
                 logs = DockerCmd(self.subtest, 'logs', [cid])
                 logs.execute()
-                if 'READY' in logs.stdout:
+                stdout = logs.stdout
+                if 'READY' in stdout:
                     return
```
cevich: Ahh there ya go, I had a feeling it was something simple.

PR #623, opened 2016-08-15
start/simple subtest: extend 'docker logs' timeout

edsantiago:
'docker logs -f' doesn't seem to be real-time: in simple
manual tests, I've seen it take 15 seconds for the STARTED
message to appear in the docker logs output. This is, I
believe, what causes the start/simple test to fail seemingly
at random on some runs. FWIW 'docker logs' (without the -f)
has no such lag. I don't know if this is a stdout buffering
issue or if -f is implemented by sleep-check (vs select).

Workaround: call `docker logs` directly, in a loop, waiting for
two STARTEDs.

Also: removed unused code and config settings, cleaned up
some confusing comments.

### DIFF
--- a/subtests/docker_cli/start/simple.py
+++ b/subtests/docker_cli/start/simple.py
@@ -23,13 +23,8 @@ from dockertest.images import DockerImage
 
 class simple(subtest.SubSubtest):
 
-    def _init_stuff(self):
-        """ Initialize stuff """
-        self.sub_stuff['container_name'] = None     # name of the container
-
     def initialize(self):
         super(simple, self).initialize()
-        self._init_stuff()
         config.none_if_empty(self.config)
         # Get free name
         docker_containers = DockerContainers(self)
@@ -38,7 +33,6 @@ class simple(subtest.SubSubtest):
 
     def _start_container(self, name):
         """ Create, store in self.sub_stuff and execute container """
-        self.sub_stuff['container_name'] = name
         if self.config.get('run_options_csv'):
             subargs = [arg for arg in
                        self.config['run_options_csv'].split(',')]
@@ -52,27 +46,29 @@ class simple(subtest.SubSubtest):
         subargs.append("'echo STARTED: $(date); while :; do sleep 0.1; done'")
         container = AsyncDockerCmd(self, 'run', subargs)
         container.execute()
-        utils.wait_for(lambda: container.stdout.startswith("STARTED"), 5,
-                       step=0.1)
+        ok = utils.wait_for(lambda: container.stdout.startswith("STARTED"), 9,
+                            step=0.1)
+        self.failif(not ok, "Timed out waiting for STARTED from container.\n"
+                    "Output: '%s'" % container.stdout)
 
     def run_once(self):
         # Execute the start command
         super(simple, self).run_once()
         name = self.sub_stuff['container_name']
-        err_msg = ("Start of the %s container failed, but '%s' message is not "
-                   "in the output:\n%s")
-        # Nonexisting container
-        missing_msg = self.config['missing_msg']
+        # Container does not yet exist; 'start' should fail.
         result = mustfail(DockerCmd(self, "start", [name]).execute(), 1)
-        self.failif(missing_msg not in str(result), err_msg
-                    % ("non-existing", missing_msg, result))
+        self.failif_not_in(self.config['missing_msg'], str(result),
+                           "'docker start <nonexistent container>' failed"
+                           " (as expected), but docker error message did not"
+                           " include expected diagnostic.")
 
-        # Running container
+        # Now run the container. The first "start" here should be a NOP.
         self._start_container(name)
         result = mustpass(DockerCmd(self, "start", [name]).execute())
 
-        # Stopped container
+        # Stop container, then restart it.
         mustpass(DockerCmd(self, "kill", [name]).execute())
+        mustpass(DockerCmd(self, "wait", [name]).execute(), 137)
         result = mustpass(DockerCmd(self, "start", [name]).execute())
 
     def postprocess(self):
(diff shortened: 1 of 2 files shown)


subtests/docker_cli/start/simple.py
```diff
@@ -52,39 +46,39 @@ def _start_container(self, name):
...
         result = mustfail(DockerCmd(self, "start", [name]).execute(), 1)
-        self.failif(missing_msg not in str(result), err_msg
-                    % ("non-existing", missing_msg, result))
+        self.failif_not_in(self.config['missing_msg'], str(result),
+                           "'docker start <nonexistent container>' failed"
+                           " (as expected), but docker error message did not"
+                           " include expected diagnostic.")
 
-        # Running container
+        # Now run the container. The first "start" here should be a NOP.
```
cevich: Meaning the one below on 72?  IIRC, that was added to verify an early BZ about double-starts causing errors.

subtests/docker_cli/start/simple.py
```diff
@@ -52,39 +46,39 @@ def _start_container(self, name):
...
         self._start_container(name)
         result = mustpass(DockerCmd(self, "start", [name]).execute())
 
-        # Stopped container
+        # Stop container, then restart it.
         mustpass(DockerCmd(self, "kill", [name]).execute())
+        mustpass(DockerCmd(self, "wait", [name]).execute(), 137)
```
cevich: Okay, looks good.

subtests/docker_cli/start/simple.py
edsantiago: Ooh, nice. That was in no way obvious from the (lack of) comments or the git history. Do you have any more details? It would be helpful to update the comment, so a future maintainer doesn't just rip it out as seemingly-useless code. E.g. "docker-1.5 aborted if you tried two consecutive start commands"

subtests/docker_cli/start/simple.py
```diff
@@ -52,39 +46,39 @@ def _start_container(self, name):
...
         logs = AsyncDockerCmd(self, "logs", ['-f', name])
         logs.execute()
-        utils.wait_for(lambda: logs.stdout.count("\n") == 2, 5, step=0.1)
-        out = logs.stdout
-        self.failif(out.count("\n") != 2, "The container was executed twice, "
-                    "there should be 2 lines with start dates, but is "
-                    "%s.\nContainer output:\n%s" % (out.count("\n"), out))
+        ok = utils.wait_for(lambda: logs.stdout.count("\n") == 2, 20, step=0.1)
+        self.failif(not ok, "Timed out waiting for second STARTED message.\n"
+                    "Output: '%s'" % logs.stdout)
```
cevich: Hrmmm, me thinks we have absolutely zero control over the buffering/log-lag.  Perhaps 'docker events' would be more reliable?  It shouldn't be too much effort to make the parser from the events sub-test part of the API.  Anyway, that's a bigger issue that I think this PR is about.

Another idea may be to simply drop the logs checking.  I'm thinking the output checking in ``run_once()`` maybe along with a second ``docker start -i`` check is enough.  What do you think?

subtests/docker_cli/start/simple.py
cevich: Ya, I can try and find the bug reference, but it was years ago....

cevich:
Overall looks good.  Not sure if there are any implications of my suggestions on other sub-subtests or not.

subtests/docker_cli/start/simple.py
cevich: ...here it is: https://bugzilla.redhat.com/show_bug.cgi?id=1096293

edsantiago: Thank you!

subtests/docker_cli/start/simple.py
edsantiago: `docker events` changed output format between 1.9 and 1.10. There's code to deal with that in the events test, I don't want to replicate it.

On reflection, just how bad is it to run `docker logs` every half-second anyway? I'm going to try that and see how it works.

subtests/docker_cli/start/simple.py
```diff
@@ -52,39 +46,43 @@ def _start_container(self, name):
...
-        logs.execute()
-        utils.wait_for(lambda: logs.stdout.count("\n") == 2, 5, step=0.1)
-        out = logs.stdout
-        self.failif(out.count("\n") != 2, "The container was executed twice, "
-                    "there should be 2 lines with start dates, but is "
-                    "%s.\nContainer output:\n%s" % (out.count("\n"), out))
+
+        def started_twice():
+            result = mustpass(DockerCmd(self, "logs", [name]).execute())
+            return result.stdout.count("STARTED") == 2
```
cevich: Ahh, okay, and they changed 'docker logs' so it doesn't limit the output to just a few lines (``--tail=all`` is default).  This is fine then.

subtests/docker_cli/start/simple.py
```diff
@@ -52,39 +46,43 @@ def _start_container(self, name):
...
+
+        ok = utils.wait_for(started_twice, 10, step=0.5)
```
cevich: IIRC you picked 20-seconds for some reason, are you sure 10 is good enough?

cevich:
mmmkay, this LGTM now, nice work.

subtests/docker_cli/start/simple.py
edsantiago: 20 seconds was to wait for buffering lag. 10 should be more than enough; I choose it over 5 because every so often on my tests I get timeouts with 5. Maybe OSP7 load, who knows,

cevich: Okay, good 'nuf, just wanted to double-check.

PR #624, opened 2016-08-16
get tests working under docker 1.12

edsantiago:
- kill, kill_stress: docker now silently ignores SIGPIPE
   (see docker commit 55a367d), so passthrough (--sig-proxy)
   no longer works. Add it to the skip list.
 - negativeusage:
   - ip1: remove --force option from tag. Option no longer
     exists in docker 1.12, and it isn't necessary in 1.10
   - error message changes.
 - run_cgroup_parent: allow any prefix before error (docker,
   docker-current, /usr/bin/docker-current, etc)
 - login: error message changes. "no successful auth challenge"
   is gone. "401 Unauthorized" remains, and that's good enough.
 - login: add a FIXME noting that the --email option to login
   is now deprecated
 - tag: -f (force) option removed in 1.12
 - top: better diagnostics

### DIFF
--- a/subtests/docker_cli/login/login.py
+++ b/subtests/docker_cli/login/login.py
@@ -161,6 +161,9 @@ class login_base(SubSubtest):
         """
         if password is None:
             password = self.sub_stuff['passwd']
+        # FIXME: --email option is deprecated in 1.12, and will be
+        # removed in 1.13. We still need it for <= 1.10; otherwise
+        # docker prompts from stdin, and we hang.
         dc = DockerCmd(self, 'login', ['--username', self.sub_stuff['user'],
                                        '--password', password,
                                        '--email', 'nobody@redhat.com',
@@ -258,8 +261,6 @@ class login_fail(login_base):
         cmdresult = self.sub_stuff['cmdresult']
         OutputGood(cmdresult, ignore_error=True)
         mustfail(cmdresult, 1)
-        self.failif_not_in("no successful auth challenge", cmdresult.stderr,
-                           "stderr from failed docker login")
         self.failif_not_in("401 Unauthorized", cmdresult.stderr,
                            "stderr from failed docker login")
 
--- a/subtests/docker_cli/top/runsleep.py
+++ b/subtests/docker_cli/top/runsleep.py
@@ -33,8 +33,8 @@ class runsleep(base):
         OutputGood(self.sub_stuff['run_dkrcmd'].cmdresult)
         OutputGood(self.sub_stuff['top_dkrcmd'].cmdresult)
         pstable = TextTable(self.sub_stuff['top_dkrcmd'].stdout)
-        self.failif_ne(len(pstable), 1)
+        self.failif_ne(len(pstable), 1, "Number of rows returned by top")
         psrow = pstable[0]
-        self.failif_ne(psrow['USER'], 'root')
-        self.failif(int(psrow['PID']) == 1)
-        self.failif_ne(psrow['COMMAND'], self.COMMAND)
+        self.failif_ne(psrow['USER'], 'root', 'Expected user')
+        self.failif(int(psrow['PID']) == 1, 'Process PID is 1')
+        self.failif_ne(psrow['COMMAND'], self.COMMAND, 'Expected command')
--- a/config_defaults/subtests/docker_cli/kill.ini
+++ b/config_defaults/subtests/docker_cli/kill.ini
@@ -6,7 +6,7 @@ wait_start = 3
 run_options_csv = --attach=stdout
 subsubtests = random_num_ttyoff,random_name_ttyoff,run_sigproxy_ttyoff,attach_sigproxy_ttyoff
 #: which signals should not be used (uncatchable signals)
-skip_signals = 9 17 19 27 18 20
+skip_signals = 9 13 17 18 19 20 27
 #: checking output produced by signal
 check_stdout = Received %%s, ignoring...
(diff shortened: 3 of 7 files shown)


subtests/docker_cli/login/login.py
```diff
@@ -161,6 +161,9 @@ def docker_login(self, password=None):
         """
         if password is None:
             password = self.sub_stuff['passwd']
+        # FIXME: --email option is deprecated in 1.12, and will be
+        # removed in 1.13. We still need it for <= 1.10; otherwise
+        # docker prompts from stdin, and we hang.
```
cevich: Ugg, that's annoying.  Does it give a warning in 1.12 but still accept the option?  if so, this can be fixed by a > 1.12 version-check.  No prob. leaving it as a FIXME here for now.

subtests/docker_cli/login/login.py
```diff
@@ -258,8 +261,6 @@ def postprocess(self):
         cmdresult = self.sub_stuff['cmdresult']
         OutputGood(cmdresult, ignore_error=True)
         mustfail(cmdresult, 1)
-        self.failif_not_in("no successful auth challenge", cmdresult.stderr,
-                           "stderr from failed docker login")
         self.failif_not_in("401 Unauthorized", cmdresult.stderr,
                            "stderr from failed docker login")
```
cevich: Is that going to break in older versions of docker, since they'll (presumably) spit out the old "no successful" message?

subtests/docker_cli/top/runsleep.py
```diff
@@ -33,8 +33,8 @@ def postprocess(self):
...
         pstable = TextTable(self.sub_stuff['top_dkrcmd'].stdout)
-        self.failif_ne(len(pstable), 1)
+        self.failif_ne(len(pstable), 1, "Number of rows returned by top")
         psrow = pstable[0]
-        self.failif_ne(psrow['USER'], 'root')
-        self.failif(int(psrow['PID']) == 1)
-        self.failif_ne(psrow['COMMAND'], self.COMMAND)
+        self.failif_ne(psrow['USER'], 'root', 'Expected user')
+        self.failif(int(psrow['PID']) == 1, 'Process PID is 1')
+        self.failif_ne(psrow['COMMAND'], self.COMMAND, 'Expected command')
```
cevich: There are some Bugs associated with this test...IIRC it was something to do with PID 1 not showing up under some circumstances.  So your changes are fine here.

cevich:
Look great, and I know this took a bunch of work despite the small number of actual changes.  Just a few questions on one test, and they're probably just me typing with my posterior again.  Feel free to comment 'har har' and merge this w/o answering in that case :grinning:

subtests/docker_cli/login/login.py
edsantiago: Yes, it emits a warning which I only saw by accident. (We really, really need to figure out a way to log all output and capture unexpected warnings). I have no idea what docker 1.13 will do if given the option, or if they really will go through with removing it. I'm just hoping they'll leave it as an undocumented NOP option, hence the FIXME.

subtests/docker_cli/top/runsleep.py
edsantiago: Reason for changes. The docker `top` command is broken in 1.12 (bz1362691). The messages from this test did not help me track down the cause. These changes helped.

subtests/docker_cli/login/login.py
edsantiago: No, because: old docker = both messages, new docker = just 401. Since we don't check all output, the extra "no successful auth challenge" message from old-docker will just be ignored.

edsantiago:
FWIW full test suite is failing because: broken exit codes (bz1362279, bz1364238); broken top command (bz1362691); and two likely race conditions related to `wait_for_ready()`

subtests/docker_cli/login/login.py
cevich: > We really, really need to figure out a way to log all output and capture unexpected warnings

``DockercmdBase`` would be the place.  Maybe a method that fires off a thread that dumps ``docker log -f`` output into ``logging.debug()`` - autotest will scoop that up automatically.  Another option would be to dump it into a file in ``self.subtest.resultsdir`` (I think that's the attribute name).

edsantiago: What I meant was more along the lines of: less sloppy checking. Instead of "does output contain the substring X", something like "is output a single line of the form <anchored-regex>". That way new warnings will not be ignored.

subtests/docker_cli/top/runsleep.py
cevich: Ahh okay.  Don't be afraid to put BZ references into comments.  Yep, that's a very similar / same to the bug I was thinking about.  Let me see if I can't dig it up, may be some helpful comments in it...here it is:  https://bugzilla.redhat.com/show_bug.cgi?id=1096286

cevich:
> FWIW full test suite is failing

Do you have python-bugzilla available/installed?  Should be able to automaticly exclude those tests by updating the whiteboard in the bugs with:  ``docker-autotest:docker_cli/whatever/whatever-else``.  If a bug is failing multiple tests, then just add another magic string on the next-line or space-separated (to the 'whiteboard' field).

On Atomic, since you can't install python-bugzilla, just drop a copy into the docker-autotest directory (which is on the python path).

cevich:
Okay, this LGTM, merging...

PR #625, opened 2016-08-18
More fixes for race conditions

edsantiago:
- kill subtest: use new wait_for_ready()
 - run_cidfile subtest: allow time for container to
   appear in 'docker ps' output. Also, document a
   rare and hard-to-fix SIGPIPE.

### DIFF
--- a/subtests/docker_cli/kill/kill_utils.py
+++ b/subtests/docker_cli/kill/kill_utils.py
@@ -129,7 +129,8 @@ class kill_base(subtest.SubSubtest):
         else:
             self._init_container_normal(name)
 
-        time.sleep(self.config['wait_start'])
+        cmd = self.sub_stuff['container_cmd']
+        cmd.wait_for_ready(timeout=self.config['wait_start'])
 
         # Prepare the "kill" command
         if self.config.get('kill_options_csv'):
--- a/subtests/docker_cli/run_cidfile/run_cidfile.py
+++ b/subtests/docker_cli/run_cidfile/run_cidfile.py
@@ -85,7 +85,7 @@ class basic(subtest.SubSubtest):
                         custom_dockercmd=None):
         """
         Starts container
-        :warning: When dkrcmd_cls is of Async type, there is no guarrantee
+        :warning: When dkrcmd_cls is of Async type, there is no guarantee
                   that it is going to be up&running after return.
         """
         def do_nothing(resutls):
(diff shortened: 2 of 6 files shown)


subtests/docker_cli/kill/kill_utils.py
```diff
@@ -129,7 +129,8 @@ def initialize(self):
         else:
             self._init_container_normal(name)
 
-        time.sleep(self.config['wait_start'])
+        cmd = self.sub_stuff['container_cmd']
+        cmd.wait_for_ready(timeout=self.config['wait_start'])
```
cevich: There's another copy of ``kill_utils.py`` that maybe needs this update as well.  I can't remember why the subtests were split in two - requiring duplicating the utils code.  I have an issue open to merge them, but maybe there's a better way to fix the problem.  Until that appears, it's probably best to at least keep the two files in-sync. with each-other.

subtests/docker_cli/run_cidfile/run_cidfile.py
```diff
@@ -137,6 +137,8 @@ def run_once(self):
                                                lambda x: mustfail(x, 125)))
         self._check_failure_cidfile_present(containers[-1])
         # cidfile already exists (exited container)
+        # FIXME: this occasionally throws SIGPIPE, presumably because
+        #  container isn't fully ready. This is a tough one to solve.
```
cevich: Does docker not catch it?  Can the test-code catch it and flag-it maybe?

cevich:
One change needs duplicating.  Otherwise LGTM.

subtests/docker_cli/run_cidfile/run_cidfile.py
edsantiago: SIGPIPE in the sense that the test is sending "exit\n" to a container that isn't ready yet. I think. It's infrequent, and I don't have time right now to characterize it. I'm adding the FIXME because once we get CI working this test will start failing occasionally, and I want the reader to understand that it's probably a false alarm.

subtests/docker_cli/kill/kill_utils.py
edsantiago: Good catch! Thank you. I hate those duplicates but don't know how to fix them. Push coming soon.

subtests/docker_cli/run_cidfile/run_cidfile.py
cevich: Ahh right, okay, I'll keep an eye out for them.  Hopefully your other race-fixes will help with this test as well.

subtests/docker_cli/kill/kill_utils.py
cevich: you and me both, sadly I was sick of arguing with the guy who wrote that test so just told him to dupe it.  Those tests are much too complicated anyway, probably the best thing is to just scrap and re-write the important ones simpler.  Until someone has motivation for that...we duplicate code happily and enjoy our coffee/tea when working on the cleaner-parts :grinning:

cevich:
Excellent work as always.

PR #627, opened 2016-08-23
New test for mount namespaces

edsantiago:
In particular, a way to detect that either docker is run
with MountFlags=latest in systemd unit file, or the kernel
issue requiring that flag is fixed (not expected until 7.3+).

### DIFF
--- a/subtests/docker_cli/namespaces/namespaces.py
+++ b/subtests/docker_cli/namespaces/namespaces.py
@@ -1,0 +1,71 @@
+"""
+Summary
+---------
+
+Tests relating to namespaces. As of 2016-08-23 only mount namespaces
+but it's likely we'll need tests for user namespaces.
+
+Operational Summary
+----------------------
+
+#. See individual tests.
+"""
+
+from autotest.client import utils
+from dockertest import subtest
+from dockertest.dockercmd import AsyncDockerCmd
+from dockertest.images import DockerImage
+from dockertest.subtest import SubSubtest
+
+
+class namespaces(subtest.SubSubtestCaller):
+    pass
+
+
+class namespaces_base(SubSubtest):
+    pass
+
+
+class rmdir_mount(namespaces_base):
+    """
+    Motivation for this test was the August 2016 attempt to package
+    docker-latest-1.12: the initial builds didn't work, docker daemon
+    wouldn't start (various BZs, not important now). The workaround was
+    to remove MountFlags=slave from the systemd unit file (1.12.0-5.el7).
+    This actually seemed to work fine, and docker-autotest mostly passed,
+    but as it turns out that workaround would've triggered problems in
+    the field. This test is an attempt to catch one such problem. It
+    passes on docker-latest-1.12.0-12.el7, fails if docker is run
+    without MountFlags=slave. See rhbz1347821 for further info.
+    """
+
+    def initialize(self):
+        super(rmdir_mount, self).initialize()
+
+    def run_once(self):
+        """
+        Run a background container with --rm; while that's running,
+        unshare its mount namespace in a longer-running process.
+        When the container finishes, we expect the --rm to succeed.
+        """
+        super(rmdir_mount, self).run_once()
+        fin = DockerImage.full_name_from_defaults(self.config)
+        subargs = ['--rm', fin, 'bash', '-c', '"echo READY; sleep 15"']
+        dkrcmd = AsyncDockerCmd(self, 'run', subargs)
+        self.sub_stuff['dkrcmd'] = dkrcmd
+        dkrcmd.execute()
+        dkrcmd.wait_for_ready(timeout=5)
+        self.sub_stuff['unshare'] = utils.run("unshare -m bash -c 'sleep 20'")
+
+    def postprocess(self):
+        """
+        If the bug triggers, we'll see:
+          Error response from daemon: Unable to remove filesystem for <sha>:
+             remove /var/lib/docker-latest/containers/<sha>/shm:
+             device or resource busy
+        """
+        super(rmdir_mount, self).postprocess()
+        self.failif_ne(self.sub_stuff['dkrcmd'].stderr, '',
+                       'stderr from docker command')
+        self.failif_ne(self.sub_stuff['unshare'].stderr, '',
+                       'unexpected stderr from unshare command')
--- a/subtests/docker_cli/namespaces/namespaces.py
+++ b/subtests/docker_cli/namespaces/namespaces.py
@@ -11,8 +11,10 @@ Operational Summary
 #. See individual tests.
 """
 
+import os
 from autotest.client import utils
 from dockertest import subtest
+from dockertest.containers import DockerContainers
 from dockertest.dockercmd import AsyncDockerCmd
 from dockertest.images import DockerImage
 from dockertest.subtest import SubSubtest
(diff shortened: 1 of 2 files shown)


subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+from dockertest.images import DockerImage
+from dockertest.subtest import SubSubtest
+
+
+class namespaces(subtest.SubSubtestCaller):
+    pass
+
+
+class namespaces_base(SubSubtest):
+    pass
```
cevich: Hrmmm, not sure why an empty base-class is needed.

subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Motivation for this test was the August 2016 attempt to package
+    docker-latest-1.12: the initial builds didn't work, docker daemon
+    wouldn't start (various BZs, not important now). The workaround was
+    to remove MountFlags=slave from the systemd unit file (1.12.0-5.el7).
+    This actually seemed to work fine, and docker-autotest mostly passed,
+    but as it turns out that workaround would've triggered problems in
+    the field. This test is an attempt to catch one such problem. It
+    passes on docker-latest-1.12.0-12.el7, fails if docker is run
+    without MountFlags=slave. See rhbz1347821 for further info.
```
cevich: excellent, helpful docstring.  Boy what a bugger of a problem.

subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def initialize(self):
+        super(rmdir_mount, self).initialize()
```
cevich: Unless it's an extension-point for a work-in-progress, there's no need for a method that just calls super, that happens automatically w/o defining the method.

subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+        When the container finishes, we expect the --rm to succeed.
+        """
+        super(rmdir_mount, self).run_once()
+        fin = DockerImage.full_name_from_defaults(self.config)
+        subargs = ['--rm', fin, 'bash', '-c', '"echo READY; sleep 15"']
+        dkrcmd = AsyncDockerCmd(self, 'run', subargs)
+        self.sub_stuff['dkrcmd'] = dkrcmd
+        dkrcmd.execute()
+        dkrcmd.wait_for_ready(timeout=5)
+        self.sub_stuff['unshare'] = utils.run("unshare -m bash -c 'sleep 20'")
```
cevich: Any other way to tickle this one that doesn't rely on the wall-clock?  Maybe use signaling instead?

subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+    def run_once(self):
+        """
+        Run a background container with --rm; while that's running,
+        unshare its mount namespace in a longer-running process.
+        When the container finishes, we expect the --rm to succeed.
+        """
+        super(rmdir_mount, self).run_once()
+        fin = DockerImage.full_name_from_defaults(self.config)
+        subargs = ['--rm', fin, 'bash', '-c', '"echo READY; sleep 15"']
+        dkrcmd = AsyncDockerCmd(self, 'run', subargs)
```
cevich: I think this is going to break if for whatever reason, the image doesn't exist.  AsyncDockerCmd won't wait for the implied-pull to finish before continuing execution here.  It's more complicated in the short term, but long-term writing out dependency on wall-clock will make test much more reliable.

subtests/docker_cli/namespaces/namespaces.py
```diff
@@ -0,0 +1,71 @@
...
+        self.sub_stuff['dkrcmd'] = dkrcmd
+        dkrcmd.execute()
+        dkrcmd.wait_for_ready(timeout=5)
```
cevich: oh oops, missed this.  Nevermind about implied-pull comment above.

subtests/docker_cli/namespaces/namespaces.py
edsantiago: Superstition on my part. I see this syntactic sugar (`class foo; class foo_base`) in almost all the existing tests, I just copy it. What do you suggest? Should I just:

    - class rmdir_mount(namespaces_base):
    + class rmdir_mount(SubSubtest):

?

subtests/docker_cli/namespaces/namespaces.py
edsantiago: My bad. This actually did something in my first iteration. Removed.

subtests/docker_cli/namespaces/namespaces.py
edsantiago: Boy did I spend a long time pondering that. I don't like the time dependency either but couldn't think of an alternative. This processes's job is to simply sit there until the container process finishes and docker tries to clean it. I could think of no way to detect "docker is done" reliably. I'll give it some more thought today, maybe I can use `docker ps` or `docker events`

subtests/docker_cli/namespaces/namespaces.py
cevich: Yeah, your fix is what I suggested! Excellent mind-reading :grinning: 

There shouldn't be many/any base-classes that are empty.  There are reasons for sub-subtests to be empty though- Normally because they're only difference is in configuration option values.  If there are subtests with empty base-classes I'd consider that a style-bug.

This is all mainly because of my current bias against extraneous code, given how badly I bit myself in the butt with ADEPT.  I even suggested a few in API additions you've made in the past, and you (rightly so) shot me down with: "If it's __actually__ needed at some future point, then it can be added.  For now, best to keep it simple."

subtests/docker_cli/namespaces/namespaces.py
cevich: Ya, it's a tricky subject, I have no doubt you considered it heavily, it's a hard problem to solve.

You especially can't use files here (?) because you're mucking around with mount namespaces, ya?  For sure there's also a desire to remain "close" to the original bug reproducer.

 In the past, I've had the best luck with using signaling, which is why I suggested it.

I'm okay with leaving it using the sleeps for now given the complexity involved.

cevich: Ahh here you go:

``sleep 10m & SLEEPID=$! trap "kill $SLEEPID" USR1 && echo "READY" && fg``

Should work in a regular shell and in docker.  Sending ``USR1`` to the parent of the shell, or via ``docker kill -s SIGUSR1 <name>`` should cause them to exit.

cevich: This works:

``docker run -ti --rm --name foobar docker.io/fedora bash -ic 'sleep 10m & SLEEPID=$! trap "kill $SLEEPID" USR1 && echo "READY" && fg'``

then in another terminal/process

``docker exec foobar kill -USR1 20``

cevich: clarification:  I got PID 20 from ``stdout`` of the ``docker run...`` command.

edsantiago:
Signals were too complicated: the quoting was a nightmare, as was extracting the sleep pid from the container. And the stuff with interactive containers, -i for both docker and bash, and then having --sig-proxy not work because of it, was really confusing. This approach isn't great either but I think ultimately it's smaller and simpler. Thanks for pushing me to do this and for your work getting signals working - that was instructive and will come in handy some day.

cevich:
I was going to suggest using semaphores :D

PR #628, opened 2016-08-29
Fix test names & tags clashing with other results

cevich:
When results from multiple runs and/or multiple hosts are combined, the
results from one test will overwrite others if the names + tags match.
Distinguish results by including the system's hostname along with the
test sequence number in result tags.

Also remove disused control.example file


cevich:
__note__: Picked systems hostname instead of a UUID because test-results are organized into directories by hostname w/in our internal CI environment.

control
```diff
@@ -1,7 +1,7 @@
 AUTHOR = "Chris Evich <cevich@redhat.com>"
 DOC = "Runs various tests for Docker"
 NAME = "Docker"
-TIME = "LONG"
+IME = "LONG"
```
edsantiago: Looks accidental.

control
```diff
@@ -487,7 +490,7 @@ class Step(collections.Callable):
         self.context = context
         if advance:
             self.context.index += 1
-        self.tag = str(self.context.index)
+        self.tag = "%d-%s" % (str(self.context.index), MYHOSTNAME)
```
edsantiago: I suspect you might not have actually tested this... (hint: %d does not go well with str())

edsantiago:
With the `self.tag` line fixed, it's now crashing in `job.py` -- it looks like something somewhere is trying to define a function based on `self.tag`, and that fails because of the dot in the hostname. I think this needs testing.

cevich:
> I think this needs testing.

That's what the CI job is for (but it's dying for reasons I outlined in my e-mail just now)

cevich:
@edsantiago thanks, fixes being fed to CI.

cevich:
This CI run will probably fail as well.  I'm seeing what looks like a syntax-error coming from autotest, likely the problem @edsantiago described above.  I'll investigate more.

cevich:
@edsantiago internal CI is happy now with this.  Looking at the results, the failures were caused by:
    * Default job timeout was only 30-minutes (WAY too short for a full-run).
    * When timeout divided between all subtests, it was too short for a 'docker pull' to complete
    * ``garbage_check`` was clobbering partially pulled layers in a few instances, orphaning images

I've increased the timeout for the job to 300-minutes, but it's not been merged yet.  I'm working on #629  for a pretest that will take care of pre-pulling the test-image, so subtests don't spend their valuable timeout-time doing this (when possible).

cevich:
jenkins, test this please.

cevich:
Argh, damn ``%d`` bites me again.  Force-pushed, testing again.

cevich:
rebased =+ force pushed

cevich:
Woopie ``From 119 tests executed, 109 have passed (pass rate of 91.0%).``  Better and better every time.

PR #630, opened 2016-09-01
Cleanup: better diagnostics from run_cgroups/memory

edsantiago:
I lost much of a day due to missing error messages from
this test. Let's avoid having that happen again.

### DIFF
--- a/subtests/docker_cli/run_cgroups/memory.py
+++ b/subtests/docker_cli/run_cgroups/memory.py
@@ -7,7 +7,7 @@ properly.
 
 from dockertest import xceptions
 from dockertest.dockercmd import DockerCmd
-from dockertest.output import OutputNotBad
+from dockertest.output import mustpass, mustfail, OutputNotBad
 from dockertest.images import DockerImage
 from cgroups_base import cgroups_base
 
@@ -170,6 +170,10 @@ class memory_base(cgroups_base):
         for subargs in self.sub_stuff['subargs']:
             dkrcmd = DockerCmd(self, 'run -d -i', subargs)
             dkrcmd.execute()
+            if self.config['expect_success'] == 'PASS':
+                mustpass(dkrcmd)
+            else:
+                mustfail(dkrcmd, 125)
             memory_containers.append(dkrcmd)
 
     def check_result(self, subargs, long_id):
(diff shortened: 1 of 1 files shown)


subtests/docker_cli/run_cgroups/memory.py
```diff
@@ -170,6 +170,10 @@ def run_once(self):
         for subargs in self.sub_stuff['subargs']:
             dkrcmd = DockerCmd(self, 'run -d -i', subargs)
             dkrcmd.execute()
+            if self.config['expect_success'] == 'PASS':
+                mustpass(dkrcmd)
+            else:
+                mustfail(dkrcmd, 125)
```
cevich: Is that exit specific to 1.10?  Meh, nobody cares about 1.9 now.  nevermind.

subtests/docker_cli/run_cgroups/memory.py
```diff
@@ -185,7 +189,11 @@ def check_result(self, subargs, long_id):
         return self.check_memory(memory_value, cgroup_memory, memory_unit)
 
     def postprocess_positive(self, this_result, passed):
-        self.failif(not isinstance(this_result, dict))
+        if this_result is None:
+            raise xceptions.DockerTestError("Invoked with null results")
+        self.failif(not isinstance(this_result, dict),
+                    ("expected this_result to be a dict; it's a %s" %
+                     this_result.__class__))
```
cevich: Nice!  Yeah, inconsistent typing like this is soooo confusing when reading a test.

subtests/docker_cli/run_cgroups/memory.py
edsantiago: `mustfail()` handles 1.9 / 1.10 exit codes

subtests/docker_cli/run_cgroups/memory.py
```diff
@@ -239,7 +249,8 @@ def postprocess(self):
                 long_id = memory_container.stdout.splitlines()[0].strip()
                 # Throws DockerIOError
                 this_result = self.check_result(subargs, long_id)
-            except (IndexError, xceptions.DockerIOError):
+            except (IndexError, xceptions.DockerIOError) as e:
+                self.logwarning("Ignoring exception: %s" % e)
```
cevich: good catch.

subtests/docker_cli/run_cgroups/memory.py
cevich: awesome

cevich:
Okay, LGTM.  Was going to merge, but I'll leave it to you in case there's anything last-minute.

edsantiago:
Nope, nothing last-minute; I finished it last night and let it sit overnight so any bugs would settle :)

PR #633, opened 2016-09-01
Race condition fix: stop test

edsantiago:
Another few hours lost to a transient failure. On two of my
virts, for reasons I don't understand, docker image spinup
is taking many seconds. More than the 'sleep 3' previously
hardcoded in the stop test. Solution: switch to wait_for_ready()
instead of hardcoded sleep. And, since I'm seeing lots of
these sort of timeouts, make it 10 seconds.

### DIFF
--- a/config_defaults/subtests/docker_cli/stop.ini
+++ b/config_defaults/subtests/docker_cli/stop.ini
@@ -1,7 +1,7 @@
 [docker_cli/stop]
 docker_timeout = 60
 #: delay before this test starts using created containers
-wait_start = 3
+wait_start = 10
 #: modifies the docker run options
 run_options_csv = --rm,--attach=stdout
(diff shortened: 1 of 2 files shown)


config_defaults/subtests/docker_cli/stop.ini
```diff
@@ -1,7 +1,7 @@
 [docker_cli/stop]
 docker_timeout = 60
 #: delay before this test starts using created containers
-wait_start = 3
+wait_start = 10
```
cevich: FWIW, I'm totally against those short timeouts, esp. in CI, nobody's watching/waiting, so go ahead and make them excessively long.

cevich:
LGTM.  CI Is happy with the changes to this test as well.

PR #635, opened 2016-09-09
Functionality for restarting docker daemon

edsantiago:
...including possibly editing run options. Tricky bits:

  - determining docker vs docker-latest
  - preserving a backup of original system options
  - making sure we restore that when done
  - making double-sure, in gc, that we leave no trace

### DIFF
--- a/dockertest/docker_daemon.py
+++ b/dockertest/docker_daemon.py
@@ -8,9 +8,8 @@ Docker Daemon interface helpers and utilities
 import httplib
 import socket
 import json
-from autotest.client.shared import service
+import os
 from autotest.client import utils
-from output import wait_for_output
 
 
 class ClientBase(object):
@@ -112,51 +111,136 @@ class SocketClient(ClientBase):
 # Group of utils for managing docker daemon service.
 
 
-def start(docker_path, docker_args):
+def _which_docker():
     """
-    Start new docker daemon with special args.
+    Returns 'docker' or 'docker-latest' based on setting in
+    /etc/sysconfig/docker.
 
-    :param docker_path: Full path to executable
-    :param docker_args: List of string of command-line arguments to pass
-    :returns: Opaque daemon_process object (not for direct use)
+    Warning: this is not a reliable method. /etc/sysconfig/docker defines
+    the docker *client*; it is perfectly possible for a system to use
+    docker as client and docker-latest as daemon or vice-versa. It's
+    possible, but unsupported, so we're not going to worry about it.
     """
-    # _SpecificServiceManager creates it's methods during __init__()
-    if service.get_name_of_init() == "systemd":
-        # pylint: disable=E1101
-        utils.run("systemctl stop docker.socket", ignore_status=True)
-    service.SpecificServiceManager("docker").stop()  # pylint: disable=E1101
-    cmd = [docker_path]
-    cmd += docker_args
-
-    daemon_process = utils.AsyncJob(" ".join(cmd), close_fds=True)
-    return daemon_process
+    docker = 'docker'
+    with open('/etc/sysconfig/docker', 'r') as docker_sysconfig:
+        for line in docker_sysconfig:
+            if line.startswith('DOCKERBINARY='):
+                if 'docker-latest' in line:
+                    docker = 'docker-latest'
+    return docker
 
 
-def output_match(daemon_process,
-                 timeout=120,
-                 regex=r"-job acceptconnections\(\) = OK \(0\)"):
+def _systemd_action(action):
+    utils.run("systemctl %s %s.service" % (action, _which_docker()))
+    # FIXME: check status
+
+
+def stop():
+    """ stop the docker daemon """
+    _systemd_action('stop')
+
+
+def start():
+    """ start the docker daemon """
+    _systemd_action('start')
+
+
+def restart():
+    """ restart the docker daemon """
+    _systemd_action('restart')
+
+
+def _preserved_extension():
     """
-    Return True if daemon_process output matches regex within timeout period
-
-    :param daemon_process: Opaque daemon_process object (not for direct use)
-    :param regex: Regular expression to search for
-    :param timeout: Maximum time to wait before returning False
+    File extension (including dot) used for our backup files
     """
-    return wait_for_output(daemon_process.get_stderr,
-                           regex,
-                           timeout=timeout)
+    return '.docker-autotest-preserved'
 
 
-def restart_service(daemon_process=None):
+def assert_pristine_environment():
     """
-    Restart the docker service using host OS's service manager
-
-    :param daemon_process: Opaque daemon_process object (not for direct use)
+    Barf if there are any leftover .docker-autotest-preserved files
+    in /etc/sysconfig; this would indicate that a previous test
+    failed to clean up properly, and our system is in an
+    undefined state.
     """
-    if daemon_process:
-        daemon_process.kill_func()
-        daemon_process.wait_for(10)
-    # _SpecificServiceManager creates it's methods during __init__()
-    if service.get_name_of_init() == "systemd":
-        utils.run("systemctl start docker.socket", ignore_status=True)
-    service.SpecificServiceManager("docker").start()  # pylint: disable=E1101
+    for suffix in ['', '-latest']:
+        path = '/etc/sysconfig/docker%s%s' % (suffix, _preserved_extension())
+        if os.path.exists(path):
+            raise RuntimeError("Leftover backup file: %s. System is"
+                               " in undefined state! Please examine that"
+                               " and its original file; if appropriate,"
+                               " mv it back into place." % path)
+
+
+def edit_options_file(remove=None, add=None):
+    """
+    Write a new /etc/sysconfig/docker* file with new OPTIONS string.
+    Preserve the original.
+
+    :param remove: string or list of strings - option(s) to remove from line
+    :param add: string or list of strings - option(s) to add to line
+    """
+    sysconfig_file = '/etc/sysconfig/%s' % _which_docker()
+    sysconfig_bkp = sysconfig_file + _preserved_extension()
+    if os.path.exists(sysconfig_bkp):
+        raise RuntimeError("Backup file already exists: %s" % sysconfig_bkp)
+    sysconfig_tmp = sysconfig_file + '.tmp'
+    with open(sysconfig_file, 'r') as sysconfig_fh_in:
+        with open(sysconfig_tmp, 'w') as sysconfig_fh_out:
+            for line in sysconfig_fh_in:
+                if line.startswith('OPTIONS='):
+                    line = edit_options_string(line, remove=remove, add=add)
+                sysconfig_fh_out.write(line)
+    os.link(sysconfig_file, sysconfig_bkp)
+    os.rename(sysconfig_tmp, sysconfig_file)
+
+
+def edit_options_string(line, remove=None, add=None):
+    """
+    Helper for edit_options_file(). Given an OPTIONS='...' string,
+    returns OPTIONS='...'\n with the given options removed and/or added.
+
+    :param line: string of the form OPTIONS='something'
+    :param remove: string or list of strings - option(s) to remove from line
+    :param add: string or list of strings - option(s) to add to line
+    """
+    if not line.startswith('OPTIONS='):
+        raise ValueError("input line does not start with OPTIONS= : %s" % line)
+    line = line[8:].rstrip()
+    quote = ''
+    if line[0] == '"' or line[0] == "'":
+        quote = line[0]
+        if line[-1] != quote:
+            raise ValueError("mismatched quotes in %s" % line)
+        line = line[1:-1]
+    if remove:
+        removes = remove if isinstance(remove, list) else [remove]
+        for remove_opt in removes:
+            if remove_opt in line:
+                line = line.replace(remove_opt, '')
+    if add:
+        adds = add if isinstance(add, list) else [add]
+        for add_opt in adds:
+            if add_opt not in line:
+                line = line + ' ' + add_opt
+    return 'OPTIONS=%s%s%s\n' % (quote, line.strip(), quote)
+
+
+def revert_options_file():
+    """
+    Revert back to preserved /etc/sysconfig/docker* file.
+
+    This function is safe to invoke even if the preserved file doesn't
+    exist; the only situation in which that makes sense is in a test's
+    cleanup() method if test prep has failed before edit_options_file().
+
+    Note that we automatically invoke restart(): there is no possible
+    situation in which it makes sense to revert options without restarting
+    docker daemon.
+    """
+    sysconfig_file = '/etc/sysconfig/%s' % _which_docker()
+    sysconfig_bkp = sysconfig_file + _preserved_extension()
+    if os.path.exists(sysconfig_bkp):
+        os.rename(sysconfig_bkp, sysconfig_file)
+        restart()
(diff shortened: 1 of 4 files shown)


edsantiago:
Not in actual production use anywhere, but I expect it to be. Tested via adding this to a randomly-picked subtest:

    in initialize: docker_daemon.edit_options_file(add="--help") ; docker_daemon.restart()

    in cleanup: docker_daemon.revert_options_file()

...and various permutations thereof, e.g. adding a real option `--debug=false`, removing the `revert` (causing gc to fail)

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+                if 'docker-latest' in line:
+                    docker = 'docker-latest'
+    return docker
 
-    daemon_process = utils.AsyncJob(" ".join(cmd), close_fds=True)
-    return daemon_process
 
+def _systemd_action(action):
+    utils.run("systemctl %s %s.service" % (action, _which_docker()))
+    # FIXME: check status
```
cevich: Aye, fixme.  systemctl starts services asynchronous to the caller.  Even calling status can be inconclusive for quite some long amount of time.  You never know when docker-storage-setup might decide it needs to stick the docker VM onto an iscsi drive mounted over a POTS ppp connection :grinning:

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+
+
+def start():
+    """ start the docker daemon """
+    _systemd_action('start')
+
+
+def restart():
+    """ restart the docker daemon """
+    _systemd_action('restart')
```
cevich: Ahh okay, hiding the implementation details.  LGTM.

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+
+
+def _preserved_extension():
     """
-    Return True if daemon_process output matches regex within timeout period
+    File extension (including dot) used for our backup files
+    """
+    return '.docker-autotest-preserved'
```
cevich: Some reason not to just make this a module-level constant?

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+    Write a new /etc/sysconfig/docker* file with new OPTIONS string.
+    Preserve the original.
 
+    :param remove: string or list of strings - option(s) to remove from line
+    :param add: string or list of strings - option(s) to add to line
+    """
+    sysconfig_file = '/etc/sysconfig/%s' % _which_docker()
+    sysconfig_bkp = sysconfig_file + _preserved_extension()
+    if os.path.exists(sysconfig_bkp):
+        raise RuntimeError("Backup file already exists: %s" % sysconfig_bkp)
```
cevich: Nice.

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+    sysconfig_tmp = sysconfig_file + '.tmp'
+    with open(sysconfig_file, 'r') as sysconfig_fh_in:
+        with open(sysconfig_tmp, 'w') as sysconfig_fh_out:
+            for line in sysconfig_fh_in:
+                if line.startswith('OPTIONS='):
+                    line = edit_options_string(line, remove=remove, add=add)
+                sysconfig_fh_out.write(line)
+    os.link(sysconfig_file, sysconfig_bkp)
+    os.rename(sysconfig_tmp, sysconfig_file)
```
cevich: Good we never ever never never ever run more than one Docker Autotest at a time on the same host :grinning:

cevich: (seriously, we never do that on purpose)

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
-    Restart the docker service using host OS's service manager
+    if not line.startswith('OPTIONS='):
+        raise ValueError("input line does not start with OPTIONS= : %s" % line)
+    line = line[8:].rstrip()
+    quote = ''
+    if line[0] == '"' or line[0] == "'":
+        quote = line[0]
+        if line[-1] != quote:
+            raise ValueError("mismatched quotes in %s" % line)
+        line = line[1:-1]
```
cevich: May be easier to use ``shlex.split()`` on the options line, then walk the options instead.  Only requires looping once and that function handles all the really strange quoting situations (i.e. line-wraps and comments) automatically.

dockertest/docker_daemon.py
```diff
@@ -112,51 +111,136 @@ def version(self):
...
+    Revert back to preserved /etc/sysconfig/docker* file.
+
+    This function is safe to invoke even if the preserved file doesn't
+    exist; the only situation in which that makes sense is in a test's
+    cleanup() method if test prep has failed before edit_options_file().
 
-    :param daemon_process: Opaque daemon_process object (not for direct use)
+    Note that we automatically invoke restart(): there is no possible
+    situation in which it makes sense to revert options without restarting
+    docker daemon.
```
cevich: Yep, that's fine.

dockertest/docker_daemon.py
cevich: Ohh, there's a side-effect here.  Maybe check/log details on any running containers before stop/restart.  Could ease debugging if a test somehow lost-track.

dockertest/docker_daemon_unittests.py
```diff
@@ -66,5 +66,56 @@ def value_to_json(value):
         self.assertEqual(i.get_json('bar'), [{u'foo': u'bar'}])
         self.assertEqual(i.interface, None)
 
+
+class TestStringEdit(unittest.TestCase):
+    pass
```
cevich: ???

dockertest/docker_daemon_unittests.py
```diff
@@ -66,5 +66,56 @@ def value_to_json(value):
...
+    def test_bad_input_no_prefix(self):
+        import docker_daemon
+        self.assertRaises(ValueError,
+                          docker_daemon.edit_options_string, "OPTINOS=hi")
+
+    def test_bad_input_mismatched_quotes(self):
+        import docker_daemon
+        self.assertRaises(ValueError,
+                          docker_daemon.edit_options_string,
+                          "OPTIONS='missing end quote")
```
cevich: Hmmmm, yep, totally running away from mock_open() aren't ya :grinning: 

I don't blame you.

dockertest/docker_daemon_unittests.py
```diff
@@ -66,5 +66,56 @@ def value_to_json(value):
...
+
+
+# Various combinations of inputs, and their expected output for
+# the edit_options_string() function.
+string_edit_tests = [
+    # original          remove          add             expected
```
cevich: (it's a unittest so IDC) namedtuple's are good for this.

dockertest/docker_daemon_unittests.py
```diff
@@ -66,5 +66,56 @@ def value_to_json(value):
...
+
 if __name__ == '__main__':
+    # Generate a test_NNN function for each test case in table above
+    for i, t in enumerate(string_edit_tests):
+        test_name = 'test_%03d' % i
+        test_ref = test_generator_string_edit("OPTIONS=%s\n" % t[0],
+                                              t[1], t[2],
+                                              "OPTIONS=%s\n" % t[3])
+        setattr(TestStringEdit, test_name, test_ref)
+
```
cevich: With unittest2, there's a 'subtest' class you can use to handle this easier form w/in a regular testcase class.

cevich:
Gave a quick/complete look.  Coming along nicely, I think it will do the business.

dockertest/docker_daemon.py
edsantiago: No, in fact I'd prefer a constant, but I have this vague recollection that Python discourages constants. If there's a best practice, I'd love to do that.

dockertest/docker_daemon.py
edsantiago: ...and if we do by accident, results are undefined anyway...

dockertest/docker_daemon.py
edsantiago: Thank you! It looks like that comes with RHEL7, so I'll take a look.

dockertest/docker_daemon_unittests.py
edsantiago: Leftover. Cleaned. Thanks.

dockertest/docker_daemon_unittests.py
edsantiago: TIL namedtuple, thanks.

dockertest/docker_daemon.py
edsantiago: So, it turns out that:

  1) systemctl is not asynchronous (I didn't think it was). If start/restart fails, it barfs. And
  2) utils.run checks exit status, and dies with a nice error message if != 0.

So all that's necessary here is to remove the FIXME. No need to add anything.

cevich: Interesting, so #1 must depend on both the unit's service-type and what the failure is.  I've been playing around with openvpn (totally different context) and it uses ``Type=forking`` (whereas docker uses Type=notify``.

I just tried it to confirm- If I screw up openvpn's config file, then ``systemctl start openvpn@server.service`` exits 0.  However, if I then run ``systemctl status openvpn@server.service`` it shows it as failed.  But I bet if I screwed up the unit file itself, then the ``start`` would exit() != 0.

Don't you just love how working with systemd is **sooooooooo** simple!

Though, even if at some future point docker's startup type changes. Won't the tests fail anyway if the daemon isn't actually running after ``docker_daemon.start()`` is called?  So maybe this is good enough.

edsantiago: I tested by adding `--asdf` to the OPTIONS line; that caused systemctl to exit failure because docker exited failure. I also tried `--help`, which caused docker to exit success (without running as daemon); this caused systemctl to succeed, but all subsequent tests to fail (because docker daemon isn't running). So I'm inclined to think this will cover all likely error scenarios.

cevich: Yep, I agree, it's probably good enough, and we can adjust as necessary if otherwise.

dockertest/docker_daemon.py
edsantiago: Many blind alleys later, I don't think `shlex.split` is the right tool for this job. Unless I'm missing something I still have to strip `OPTIONS=` myself, and do quote recognition myself, so the only way shlex could help is if I'm invoked to add/remove `--foo` but there's an option such as `OPTIONS='-a -b --disable=--foo'`. That doesn't seem worthwhile. 

It's possible I've misunderstood or missed something important about shlex. What did you have in mind here?

cevich: Oh, maybe it's my mistake misunderstanding this method.  I was thinking something like:

```python
def f(line,adds=[],removes=[]):
    new_ops = []
    ops = shlex.split(line, False)
    while ops:
        _op = ops.pop()
        if _op not in removes:
            new_ops.append('"%s"' % _op)
    try:
        while adds:
            new_ops.append(adds.pop())
    except AttributeError:
            new_ops.append(adds)
    new_ops.reverse()
    return "OPTIONS='%s'" % ' '.join(new_ops)
```

Funky looking output, but I think it will always be shell-correct.  Anyway, it's only 4 lines shorter than what you have, so I suppose it doesn't much matter.

dockertest/docker_daemon_unittests.py
```diff
@@ -67,8 +67,35 @@ def value_to_json(value):
...
+    ]
+
+    def test_edit_options_string(self):
+        import docker_daemon
+        for (opts_in, remove, add, opts_out) in self.string_edit_tests:
+            with self.subTest(name="%s -<%s> +<%s>" % (opts_in, remove, add)):
+                s_in = 'OPTIONS=%s\n' % opts_in
+                expected = 'OPTIONS=%s\n' % opts_out
+                actual = docker_daemon.edit_options_string(s_in, remove, add)
+                self.assertEqual(actual, expected)
```
cevich: Wow, that's dense, I need more coffee :grinning:

cevich:
Yay travis is happy.  okay lessee what Mr. Janky-Jankenz is whining about...

cevich:
...uggg, okay, those are the same six that I've seen fail before, haven't looked into why yet.  Probably test configuration errors.

cevich:
This LGTM.  Ready to merge?

edsantiago:
Jenkins failures:
 - build : it's a timeout. Solution: change the "4" in the TIMEOUT line in control to "25". Yeah, I know it shouldn't be necessary, but it is and it has been ever since I started working with autotest. I just do it automatically on all my new virts.
 - dockerinspect - doesn't work on CentOS, never has
 - run_cgroups - looks like the same underpowered-VM-with-no-swap problem I was seeing
 - start - bz1330224, will not be fixed in 1.10
 - redhat/packaging, redhat/rhel_push_plugin - this is CentOS

Yes, I think I'm ready to merge, thanks!

dockertest/docker_daemon.py
edsantiago: One problem is that `line` is not `-a -b -c`, it is `OPTIONS=<optional quote>-a -b -c<close quote>`. Another is that this adds duplicates, i.e. doesn't check if `adds` is already in the option list.

cevich:
Thanks for taking a closer look.

*  ``build`` - there are actual build failures by the looks of it.  The TIMEOUT line is modified by ADEPT.  Here's what it was changed to: https://url.corp.redhat.com/c567aeb ``Subtest timeout set to 252.6316s``  This could be config. issues w/ the test and the CentOS image or behavior changes not captured in config.

*  ``dockerinspect`` should work fine on CentOS, looks like it needs updating to avoid checking certain branches.  Should be a quick fix, I'll add it to my list.

*  kk, probably need to setup swap then.  Maybe file a bug that atomic host doesn't have any swap.  On my TODO.

* Hmmm start - bz1330224 should be getting skipped.  I see it skipped BZ1344132.  Something must not be working right in ``control`` I'll play with it some more to see why it's not skipping.

* redhat/packaging, redhat/rhel_push_plugin ahh, probably should just skip **all** the redhat/ subtests.  Okay, I can change that easy enough.

dockertest/docker_daemon.py
cevich: Ahh okay.  Could always strip off ``OPTIONS=`` at the head, then add it back in at the end.  Well, I was just hoping we could get away with having a standard module do all the quote-wrangling work for us.  Could always fix duplicate adds by including them into the removes list.  So...

```python
def edit_options_string(line, remove=None, add=None):
    if remove is None:
        remove = []
    if adds is None:
        add = []
    line = line.split('=',1)[1:]
    new_ops = []
    ops = shlex.split(line, False)
    while ops:
        _op = ops.pop()
        if _op not in remove and _op not in add:
            new_ops.append('"%s"' % _op)
    try:
        while adds:
            new_ops.append(add.pop())
    except AttributeError:
            new_ops.append(add)
    new_ops.reverse()
    return "OPTIONS='%s'" % ' '.join(new_ops)
```

Anyway, I don't intend to re-write what you've already written.  What you have is probably fine, I'm just tinkering :smiley:

edsantiago: line = line.split('=',1)[1:]

Removes leading `OPTIONS=` but does not deal with the quotes. I know `shlex` seems very useful but it's not the right tool for this job.

cevich:
[(?) [CI] Test block registry covers docker.io](https://trello.com/c/i8mgkm1g/318-ci-test-block-registry-covers-docker-io)

PR #636, opened 2016-09-12
Adding container as systemd service

afomm:
Adding container as systemd service subtest with static config
A really really rough version


config_defaults/control.ini
```diff
@@ -78,6 +78,10 @@ key_match = docker-autotest
 product = Red Hat Enterprise Linux 7
 component = docker, docker-distribution, docker-registry, atomic,
             rhel-server-atomic, rhel-server-docker, rhel-tools-docker,
+<<<<<<< Updated upstream
```
edsantiago: Eek! These are conflict markers. They indicate that you tried to `git merge` at some point, and git found changes upstream that were incompatible with changes that you made.

@cevich there are git hooks for preventing commits with conflict markers. Do you know if/how to enable them in github?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+                    "Returned system time is not a digit")
+        # stop container
+        self.sysd_action(action = 'stop',
+                            sysd_unit = self.config['sysd_unit_file'])
+
+    def sysd_action(self, action = None, sysd_unit = None):
+        if sysd_unit is 'None':
+            command = "systemctl " + action
+        else:
+            command = "systemctl " + action + " " + sysd_unit
```
edsantiago: One way to do this without duplication might be:

    command = "systemctl " + action
    if sysd_unit:
        command += " " + sysd_unit

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+                                ignore_status=True)
+        system_time_adigit = system_time.stdout.strip().isdigit() 
+        self.failif(not system_time_adigit,
+                    "Returned system time is not a digit")
+        # stop container
+        self.sysd_action(action = 'stop',
+                            sysd_unit = self.config['sysd_unit_file'])
+
+    def sysd_action(self, action = None, sysd_unit = None):
+        if sysd_unit is 'None':
```
edsantiago: This does not do what you think it does! You are comparing to the string 'None', not to the actual python None. Huge difference. See below for a way to avoid this confusion.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+
+class systemd(subtest.Subtest):
+
+    def initialize(self):
+        super(systemd, self).initialize()
+        sysd_unitf = join(self.bindir, self.config['sysd_unit_file'])
+        dockerfilesrc = join(self.bindir, 'Dockerfile')
+        script_src = join(self.bindir, 'p4321-server.py')
+        dockerbuild_dir = join(self.bindir, self.config['tmpblddir'])
+        self.config['dockerbuild_dir'] = dockerbuild_dir
```
edsantiago: I think you want to use `self.sub_stuff`, not self.config.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        self.config['unitf_dest'] = join(self.config['sysd_dest'],
+                                        self.config['sysd_unit_file'])
+        utils.run("mkdir %s" %dockerbuild_dir, ignore_status=True)
+        copy(sysd_unitf, self.config['sysd_dest'])
```
edsantiago: Future maintainers will thank you if you make the destination be `unitf_dest` (i.e. the filename), not sysd_dest (directory name). And again, `self.sub_stuff[]` not config.

config_defaults/control.ini
afomm: Sorry about that. After I rebased & pushed, I noticed my branch isn't working. Should I remove my commit?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class systemd(subtest.Subtest):
+
+    def initialize(self):
+        super(systemd, self).initialize()
+        sysd_unitf = join(self.bindir, self.config['sysd_unit_file'])
+        dockerfilesrc = join(self.bindir, 'Dockerfile')
+        script_src = join(self.bindir, 'p4321-server.py')
+        dockerbuild_dir = join(self.bindir, self.config['tmpblddir'])
```
edsantiago: Hm, probably not. You might want to use `self.tmpdir`

config_defaults/control.ini
edsantiago: No, you need to resolve the conflict. That's waaaay beyond the scope of what I can help with in a github message. Can you read up on git, and git merges in particular, and/or ask someone local for help?

Merge conflicts are complicated and very, very important to understand. Please feel free to take a day or two to read, including working with examples in a scratch repo, and develop a basic understanding.

config_defaults/defaults.ini
```diff
@@ -53,7 +53,8 @@ remove_after_test = yes
 #: CSV of possibly existing **full** image names to preserve.
 preserve_fqins = docker.io/stackbrew/centos:7,
                  docker.io/stackbrew/centos:centos7,
-                 docker.io/stackbrew/centos:latest
+                 docker.io/stackbrew/centos:latest,
+                 p4321:latest
```
edsantiago: I don't think you really want this here. What this does is, when tests finish, garbage collection leaves `p4321:latest` lying around. This is pretty bad, because if the subtest changes it will be running the old version. You want this to be a throwaway image, one that gets deleted after each use.

subtests/docker_cli/systemd/p4321.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Systemd for p4321 service
+After=docker-latest.service
+
+[Service]
+Environment=NAME=p4321
+ExecStartPre=-/usr/bin/docker-latest stop p4321
+ExecStartPre=-/usr/bin/docker-latest rm -f p4321
```
edsantiago: Probably not a good idea. These are only useful if p4321 is running when it isn't supposed to be, and that would be a bug.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        # may able to use build_base class
+        # dkrcmd = DockerCmd(self, 'build', self.config['build_options_csv'])
+        # self.config['dkrcmd'] = dkrcmd
+        # dkrcmd.execute()
+        # 
+        build_args = self.config['docker_path'] + " build --force-rm -t " \
+                    + self.config['name'] + " -f " \
+                    + self.config['dockerbuild_dir'] +"/Dockerfile " \
+                    + self.config['dockerbuild_dir']
+        utils.run(build_args)
```
edsantiago: Yeah, please don't do this (using utils.run to run docker)

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        copy(sysd_unitf, self.config['sysd_dest'])
+        copy(dockerfilesrc, dockerbuild_dir)
+        copy(script_src, dockerbuild_dir)
+        self.build_image()
+
+    def build_image(self):
+        # FIXME
+        # a temporary ugly hack
+        # may able to use build_base class
+        # dkrcmd = DockerCmd(self, 'build', self.config['build_options_csv'])
```
edsantiago: The way to get this to work is to just delete/forget `build_options_csv` and do

    dkrcmd = DockerCmd(self, 'build', ['--force-rm', '-t', self.config['name'], '-f', ...])

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        if sysd_unit is 'None':
+            command = "systemctl " + action
+        else:
+            command = "systemctl " + action + " " + sysd_unit
+        utils.run(command, ignore_status=False)
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        utils.run("rm -fr %s" %self.config['dockerbuild_dir'])
+        utils.run("rm -f %s" %self.config['unitf_dest'])
```
edsantiago: Try `os.unlink()`. And if you use `self.tmpdir` as suggested above, you won't need the rm -rf.

Also: when you delete a systemd unit file, you need to rerun daemon-reload.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        utils.run(build_args)
+
+    def run_once(self):
+        self.sysd_action(action = 'daemon-reload', sysd_unit = 'None')
+        self.sysd_action(action = 'start',
+                            sysd_unit = self.config['sysd_unit_file'])
+        # FIXME
+        # instead of sleep, use (not utils.is_port_free(4321, localhost))
+        time.sleep(5)
+        system_time = utils.run("cat < /dev/tcp/localhost/4321",
```
edsantiago: overkill. Hint: any time you find yourself running `cat` with one argument, you should ask yourself if there's a better way. There is. Basically, just use any standard way to read a line from a file. E.g.

    result = open('/dev/tcp/localhost/4321').read().strip()

Also: the `<` is unnecessary. Just a hint for next time.

edsantiago:
Nice work. Short, mostly clean. Needs some tune-up but is a good start.

cevich:
Ugg, well, ed's beat me to the party by about 10,000 review comments :confounded:  Please walk down and poke me for help untangling the merge markers and whatnot.  Worst I'll do is point and laugh :smiley:

(I've made this mistake a few times before as well)

@edsantiago  Martin gave me good advice once:  Don't overwhelm n00bs with too through of a review.  The top two-or-three major things you find are enough.  The rest will come together over subsequent reviews.  This also gives them a chance to find/fix the other problems on their own.  Also, it gives others a chance to participate in reviewing :P

subtests/docker_cli/systemd/systemd.py
cevich: > overkill. Hint: any time you find yourself running `cat` with one argument
>
>     result = open('/dev/tcp/localhost/4321').read().strip()
>
> Also: the `<` is unnecessary. Just a hint for next time.

This is a phony bash redirect token, that file doesn't exist.  The way 
to do it in python is using the socket module (which isn't easy).  I 
think we can leave this one for fixing later.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,84 @@
...
+        self.sysd_action(action='start',
+                         sysd_unit=self.stuff['sysd_unit_file'])
+        while is_port_free(self.stuff['tcp_port'], self.stuff['tcp_ip']):
+            time.sleep(2)
+        # FIXME
+        # for now, using cat but
+        # will try something similar to open() below
+        # read_port = open('/dev/tcp/localhost/4321').read().strip()
+        system_time = utils.run("cat < /dev/tcp/localhost/4321",
+                                ignore_status=False)
```
cevich: This is a phony bash redirect token (and utils.run uses bash).  So using ``open()`` won't work,  that file doesn't exist.  The way to do it in python is using the ``socket`` module (which isn't easy).  I think we can leave this one running 'cat' for fixing later, before we merge.

cevich: https://docs.python.org/2/howto/sockets.html

afomm:
@edsantiago @cevich - Thanks for the feedback guys. I applied Ed's feedback except for one or two that I need to fix - 'cat ...' being one of them. I force push again early this morning. However, travis is failing on sysd_action with 'Method could be a function'. Please review.

afomm:
@edsantiago @cevich - Travis is happy now.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,80 @@
...
+class systemd(Subtest):
+
+    def initialize(self):
+        super(systemd, self).initialize()
+        self.stuff['tcp_ip'] = self.config['tcp_ip']
+        self.stuff['tcp_port'] = self.config['tcp_port']
+        self.stuff['sysd_unit_file'] = self.config['sysd_unit_file']
+        self.stuff['sysd_unitf_dest'] = self.config['sysd_unitf_dest']
+        self.stuff['name'] = self.config['name']
+        self.stuff['self.tmpdir'] = self.tmpdir
```
edsantiago: I guess I wasn't very clear. `self.stuff` and `.sub_stuff` are for things that are defined in one place and used in another. The lines above are just duplication, because the values in `.stuff` are identical to the ones in `.config`. 

Put another way, this is a useful definition:

    self.stuff['cmd'] = DockerCmd(....something)

See that DockerCmd returns an actual variable, something changing, that needs to be used later in a different function. But this is not useful:

    self.stuff['a'] = self.config['a']

..because all functions in this module have access to self.config['a']. See the difference?

So, you can just delete lines 29-34.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,80 @@
...
+        sysd_unitf = join(self.bindir, 'p4321.service')
+        dockerfilesrc = join(self.bindir, 'Dockerfile')
+        script_src = join(self.bindir, 'p4321-server.py')
+        shutil.copyfile(sysd_unitf, self.stuff['sysd_unitf_dest'])
+        shutil.copy(dockerfilesrc, self.stuff['self.tmpdir'])
+        shutil.copy(script_src, self.stuff['self.tmpdir'])
```
edsantiago: Future maintainers will appreciate it if you condense the above two lines into:

    for build_file in ['Dockerfile', 'p4321-server.py']:
        shutil.copy(join(self.bindir, build_file), self.tmpdir)

subtests/docker_cli/systemd/systemd.py
afomm: So use self.config['something'] throughout?

edsantiago: Yes, exactly.

IIRC I suggested using `sub_stuff` because the previous version of your PR did some combinations of `bindir` and `self.config[something]`. But in this version of your PR there's nothing like that.

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -0,0 +1,20 @@
...
+#: build image ``name``
+name=p4321
+
+#: image building args ``build_options_csv``
+build_options_csv = --force-rm,-t,%(name)s
+
+#: local ip address ``tcp_ip``
+tcp_ip = 127.0.0.1
+
+#: port to use ``tcp_port``
```
edsantiago: Why do you have all these things like ``this`` and ``that`` in the comments? It seems like unnecessary duplication, and it takes away space from an actually helpful comment. But I've never seen that before, and am curious if it's just something I don't know about.

edsantiago: Anyhow, all these comments should probably be filled in a little better with descriptions of what the fields are for.

subtests/docker_cli/systemd/p4321.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Systemd for p4321 service
```
edsantiago: It might be helpful to note that this is a scratch/test service, and perhaps include a note of the form "this is part of docker-autotest, if you see it on a live system it means something has failed very badly". The purpose is to realize that there's a small window during which a test crash could leave this file in the system, and someone might find it, and a helpful comment will save them from wasting time.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,80 @@
...
+-------
+
+Operational Summary
+---------------------
+    1. Copy Unit file
+    2. Build image using predefined Dockerfile & p4321-server.py
+    3. Run image
+    4. Confirm p4321-server script is running
+    5. Stop container
+    6. cleanup
```
edsantiago: Take a step back. Imagine that you don't know anything about this test. Does that description help you? "Copy unit file" -- what unit file? Copy where? What does it do? What would a future maintainer like to know about this test and how it works?

afomm: I agree. I will make it more descriptive since these were just for me.

subtests/docker_cli/systemd/p4321.service
afomm: I agree again.

config_defaults/subtests/docker_cli/systemd.ini
afomm: When an unknown var is used here, Travis complains and I added these to trick it but I have to make it concise & descriptive.

subtests/docker_cli/systemd/systemd.py
afomm: Ed, I really appreciate all these insightful feedback.

config_defaults/subtests/docker_cli/systemd.ini
cevich: IIRC, the port number is hard-coded in the docker file, so probably no reason to have it be a config. option.  Same for the ``tcp_ip``.  Localhost is localhost is 127.0.0.1 everywhere (with only one minor exception I know of).  Ordinarily it would be helpful to have this as an option.  But now that I think about it, for this specific test, I can't see any reason it will ever need to not be localhost.

All the other options you have here seem perfectly reasonable to me.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+        sysd_unitf = join(self.bindir, 'p4321.service')
+        shutil.copyfile(sysd_unitf, self.config['sysd_unitf_dest'])
+        for build_file in ['Dockerfile', 'p4321-server.py']:
+            shutil.copy(join(self.bindir, build_file), self.tmpdir)
+        self.build_image()
+
+    def build_image(self):
+        dkrcmd = DockerCmd(self, 'build', ['--force-rm -t',
+                                           self.config['name'], self.tmpdir])
+        dkrcmd.execute()
```
cevich: What's your thought behind building from ``tmpdir``.  Why not skip the copy entirely and
```python
       dkrcmd = DockerCmd(self, 'build', ['--force-rm -t',
                                          self.config['name'], self.bindir])
```

?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+        time.sleep(5)
+
+    def run_once(self):
+        self.sysd_action(action='daemon-reload')
+        self.sysd_action(action='start',
+                         sysd_unit=self.config['sysd_unit_file'])
+        while is_port_free(self.config['tcp_port'], self.config['tcp_ip']):
+            time.sleep(2)
```
cevich: Minor nit-pick:  The CPU won't even break a sweat if you made this ``time.sleep(0.1)``.  Sometimes shaving off an extra second or 1.736346 seconds is important, esp. if there's an impending timeout involved.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+    def sysd_action(action=None, sysd_unit=None):
+        command = "systemctl " + action
+        if sysd_unit:
+            command += " " + sysd_unit
+        utils.run(command, ignore_status=False)
+        time.sleep(2)
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        utils.run("rm -f %s" % self.config['sysd_unitf_dest'])
```
cevich: ``os.unlink()`` is the python-way.  Though it can be annoying if the file isn't there.   This works well:

```python
import errno
...
filepath = self.config['sysd_unitf_dest']
try:
    os.unlink(filepath)
except OSError, xcpt:
    if xcpt.errno != errno.ENOENT:  # No such file or directory
        raise
    self.logwarning("Failed to remove %s, it's completely invisible!"
                    % filepath)
```

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+        self.sysd_action(action='stop',
+                         sysd_unit=self.config['sysd_unit_file'])
+
+    @staticmethod
+    def sysd_action(action=None, sysd_unit=None):
+        command = "systemctl " + action
+        if sysd_unit:
+            command += " " + sysd_unit
+        utils.run(command, ignore_status=False)
+        time.sleep(2)
```
cevich: don't forget to kill this sleep.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        utils.run("rm -f %s" % self.config['sysd_unitf_dest'])
+        time.sleep(2)
+        self.sysd_action(action='daemon-reload')
```
cevich: There are reasons a test can fail, even before ``run_once()`` or even ``initialize()`` started/finished.  In those cases, it would be wise to avoid unnecessary restarts of systemd.  Perhaps something like this would be a nice touch:

```python
...
        def initialize(self):
            ...
            self.stuff['systemd_touchstone'] = False
            ...

        def sysd_action(self, action=None, sysd_unit=None):
            self.stuff['systemd_touchstone'] = True
            ...

        def cleanup(self):
            # Always remove file if possible
            try: ... unlink() ...
            # Only refresh systemd when required
            if  self.stuff['systemd_touchstone']:
                self.sysd_action(action='daemon-reload')
            ...
```

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,71 @@
...
+        command = "systemctl " + action
+        if sysd_unit:
+            command += " " + sysd_unit
+        utils.run(command, ignore_status=False)
+        time.sleep(2)
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        utils.run("rm -f %s" % self.config['sysd_unitf_dest'])
+        time.sleep(2)
```
cevich: reminder: another sleep to kill

config_defaults/subtests/docker_cli/systemd.ini
edsantiago: @afomm I would assume those Travis warnings come from `./run_checkdocs.py` but I've been unable to reproduce them. BTW what I'm referring to is the backquote-backquote-varname-backquote-backquote strings. Try removing those, writing full param descriptions, and run `./run_checkdocs.py` on your system to confirm they're ok.

@cevich, nice suggestion. Indeed there's no need to make 127.0.0.1 and 4321 configurable.

Also, the comment on docker_timeout is misleading.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        super(systemd, self).initialize()
+        self.stuff['systemd_touchstone'] = False
+        sysd_unitf = join(self.bindir, 'p4321.service')
+        shutil.copyfile(sysd_unitf, self.config['sysd_unitf_dest'])
+        self.build_image()
+
+    def build_image(self):
+        dkrcmd = DockerCmd(self, 'build', ['--force-rm -t',
+                                           self.config['name'], self.bindir])
+        dkrcmd.execute()
```
edsantiago: I think it might be wise to wrap this in `mustpass()`, i.e.

    from dockertest.output import mustpass
    ...
    mustpass(dkrcmd.execute())

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def run_once(self):
+        host = 'localhost'
+        PORT = 4321
+        self.sysd_action(action='daemon-reload')
+        self.sysd_action(action='start',
+                         sysd_unit=self.config['sysd_unit_file'])
```
edsantiago: On reflection, how about ditching the `action=` and `sysd_unit=` ?

    self.sysd_action('start', self.config['etc'])

That reads naturally to anyone who has ever used systemctl.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+            time.sleep(0.1)
+        # Read time from socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((host, PORT))
+        time_from_socket = s.recv(16)
+        s.close()
+        # check if read time is a number.
+        # Maybe add another check - something like len(str(time_from_socket))
+        time_from_socket_adigit = time_from_socket.strip().isdigit()
+        self.failif(not time_from_socket_adigit,
```
edsantiago: Very minor nit: "adigit" isn't the important thing here, it's really the time. What about:

    time_from_socket = s.recv(16).strip()
    ...
    self.failif(not time_from_socket.isdigit())

Again, very minor. I just find it slightly more readable because the tested condition is inside the `failif` and could easily be changed/adapted if necessary without changing variable names.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        self.stuff['systemd_touchstone'] = True
+        command = "systemctl " + action
+        if sysd_unit:
+            command += " " + sysd_unit
+        utils.run(command, ignore_status=False)
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        filepath = self.config['sysd_unitf_dest']
+        try:
```
edsantiago: A slightly cleaner way might be:

    if os.path.exists(filepath):
        os.unlink(filepath)

No complicated exceptions to check for.

afomm:
@edsantiago @cevich - integrated latest feedback. Please review & Thanks!

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,69 @@
...
+        host = 'localhost'
+        PORT = 4321
+        self.sysd_action('daemon-reload')
+        self.sysd_action('start', self.config['sysd_unit_file'])
+        while is_port_free(PORT, host):
+            time.sleep(0.1)
+        # Read time from socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((host, PORT))
+        time_from_socket = s.recv(16).strip()
```
edsantiago: So, there's a race condition here. About one time in four, in my stress tests, s.recv() returns '' (empty string, i.e. EOF). I've wasted a lot of time trying to resolve this in p*server.py, and can't. Instead, the (ugly) workaround I've settled on is to move `socket.socket` ... `s.close` into a separate function that retries until it receives non-null.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,69 @@
...
+        s.close()
+        self.failif(not time_from_socket.isdigit())
```
edsantiago: To help future maintainers, an error message would be nice:

    self.failif(not ...isdigit(), "Data received from container is non-numeric: '%s'" % time_from_socket)

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,69 @@
...
+        self.stuff['systemd_touchstone'] = False
+        sysd_unitf = join(self.bindir, 'p4321.service')
+        shutil.copyfile(sysd_unitf, self.config['sysd_unitf_dest'])
+        dkrcmd = DockerCmd(self, 'build', ['--force-rm -t',
+                                           self.config['name'], self.bindir])
+        mustpass(dkrcmd.execute())
+
+    def run_once(self):
+        host = 'localhost'
+        PORT = 4321
```
edsantiago: Caps make me think global. This should probably be just 'port'.

afomm: OK - I'll make it local :+1:

subtests/docker_cli/systemd/systemd.py
afomm: I had the sleep(X) in several places because I was seeing something similar but I haven't run into it since my changes. I'll add your suggestion.

subtests/docker_cli/systemd/systemd.py
afomm: Yes, a good idea again

afomm:
@edsantiago @cevich - made the changes Ed suggested. I am not sure if "num_read_try = 10" without sleep(0.x) is good/long enought? Please review. - Thanks

subtests/docker_cli/systemd/p4321.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=docker-autotest systemd subtest. If you see this file on a live system it means something has failed very badly
```
cevich: Hehe, nice.

subtests/docker_cli/systemd/p4321.service
```diff
@@ -0,0 +1,11 @@
...
+After=docker-latest.service
```
cevich: Is this functionality only ever to be supported in whatever docker-latest is?

Maybe supply two unit files, one for ``docker``, another for ``docker-latest``.  Then add a config option that names which to use.  That way it's up to the tester which to set.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
+"""
+Summary
+-------
+Testing basic container as systemd service
+
+Operational Summary
+---------------------
+This is to test container as systemd service using a predefined unit file
+(p4321.service) and a script p4321-server.py. An image is built with this
```
cevich: The docstrings are parsed as RST, so it's good to add double-back-quotes around anything that should be in a courier/system font.  Makes them easier to read.  e.g.

This is to test container as systemd service using a predefined unit file
(``p4321.service``) and a ``script p4321-server.py``. An image is built with this
script added. Systemd starts a container using this image & the container
writes current time to port ``4321``. From host, the socket is read and the value
is checked to see if it is a digit. Finally, container is stopped through
systemd and system is cleaned up.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        if self.stuff['systemd_touchstone']:
+            self.sysd_action('daemon-reload')
+
+    @staticmethod
+    def read_socket(host, port):
+        time_from_socket = ''
+        num_read_try = 10
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((host, port))
+        while not len(str(time_from_socket)):
```
cevich: Any reason you can think of why this couldn't just be:  ``while not time_from_socket: ...``?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+            time_from_socket = s.recv(16).strip()
+            num_read_try -= 1
+            if num_read_try == 0:
+                break
```
cevich: This can just be added to the loop condition as well:

```python
while num_read_try or not time_from_socket:
    ...
```

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+            self.sysd_action('daemon-reload')
+
+    @staticmethod
+    def read_socket(host, port):
+        time_from_socket = ''
+        num_read_try = 10
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((host, port))
+        while not len(str(time_from_socket)):
+            time_from_socket = s.recv(16).strip()
```
cevich: Unless there's complete garbage in the socket buffer, then you leave it hanging.  If the 16-bytes you happened to read were a digit, it could look like a PASS.  The size here is a //maximum// value, so it's fine to set it large, like ``512``.

Also, this is a blocking call, so if there's no data, it will just sit there and wait forever.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+            unlink(filepath)
+        if self.stuff['systemd_touchstone']:
+            self.sysd_action('daemon-reload')
+
+    @staticmethod
+    def read_socket(host, port):
+        time_from_socket = ''
+        num_read_try = 10
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((host, port))
```
cevich: You can't predict networking conditions, it's good to stick timeouts on these things.  Normally, there isn't one, or if there is it's very long (minutes).

```python
    s = socket.socket.create_connection((host, port), timeout=10)
    s. socket.settimeout(1.0)  # Sets non-blocking w/ timeout
    while ...
        ...
        try:
            time_from_socket = s.recv(512).strip()  # non-blocking
        except socket.timeout:
            pass  # will try again a few more times
    ...
    s.shutdown(socket.SHUT_RDWR)
```
(untested)

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        while not len(str(time_from_socket)):
+            time_from_socket = s.recv(16).strip()
+            num_read_try -= 1
+            if num_read_try == 0:
+                break
+        s.close()
```
cevich: Good to call s.shutdown() instead.  The close() call will happen automatically when the socket reference is lost (``s`` goes away).

subtests/docker_cli/systemd/p4321.service
edsantiago: Good catch! It might be simpler just to remove this line entirely.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,81 @@
...
+        if exists(filepath):
+            unlink(filepath)
+        if self.stuff['systemd_touchstone']:
+            self.sysd_action('daemon-reload')
+
+    @staticmethod
+    def read_socket(host, port):
+        time_from_socket = ''
+        num_read_try = 10
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
edsantiago: So, couple of things. First, as you suggested, a `sleep` is necessary between retries. `sleep(0.5)` is IMHO a reasonable value.

Second, I think you have to have `socket.socket` inside the while loop, not outside. Retrying `recv()` after EOF doesn't seem to work; yesterday I saw cases where just retrying `recv` just kept getting empty string. I haven't reproduced it today, though.

This is what I have working on my end; but note that it doesn't incorporate @cevich's changes because I haven't taken time to look at `create_connection` or `settimeout`.

def read_socket(host, port):
        num_read_try = 10
        while num_read_try:
            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
            s.connect((host, port))
            time_from_socket = s.recv(16).strip()
            if time_from_socket:
                s.close()
                return time_from_socket
            self.logwarning("empty read from socket")   # just for my debugging
            num_read_try -= 1
            time.sleep(0.5)
        raise SomeSortOfExceptionIndicatingThatWeGotNoDataFromSocket

subtests/docker_cli/systemd/p4321.service
```diff
@@ -0,0 +1,11 @@
...
+
+[Service]
+ExecStart=/usr/bin/docker-latest run -p 4321:4321 --name p4321 p4321
+ExecStop=-/usr/bin/docker-latest stop p4321
+ExecStop=/usr/bin/docker-latest rm -f p4321
```
edsantiago: You also want an ExecStop with `rmi`, to avoid a warning from garbage_check

cevich: or just remove the image in the test's ``cleanup()`` (there's a ``DockerImages.cleanup()`` call you can use IIRC).

edsantiago: Ooops, yes, definitely what @cevich said. Since the `docker build` is done in the test, the `rmi` should be as well.

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -0,0 +1,8 @@
+[docker_cli/systemd]
+docker_timeout = 300
```
afomm: Yes, default is 300. I added that just in case it was changed by user to something shorter in config_custom/defaults.ini. I can remove it if that isn't a good enough reason.

subtests/docker_cli/systemd/p4321.service
afomm: That came from Ed :+1:

subtests/docker_cli/systemd/p4321.service
afomm: Yes, good catch.

subtests/docker_cli/systemd/systemd.py
afomm: No reason and Ed has a good suggestion below.

subtests/docker_cli/systemd/p4321.service
afomm: It might be an overkill but I have the following method to edit the unit file depending on what docker_path is set.
   def edit_unit_file(self, unit_file):
        # edit systemd unit file and set it to value of docker_path
        filedata = None
        dockerpath = self.config['docker_path']
        docker_service = dockerpath.split('/')[-1] + ".service"
        with open(unit_file,'r') as file:
            for line in file:
                if 'After' in line:
                    docker_ufile = line.split('=')[-1].split('.')[0]
                    docker_ufile_path = "/usr/bin/" + docker_ufile
                    service_line = docker_ufile + ".service"
        with open(unit_file,'r') as file:
            filedata = file.read()
            if dockerpath != docker_ufile_path:
                filedata = filedata.replace(docker_ufile_path, dockerpath)
                filedata = filedata.replace(service_line, docker_service)
        with open(unit_file,'w') as file:
            file.write(filedata)

cevich:
I dunno where it went, but I got an e-mail about your unit file editing method.  My recommendation is to keep it simple, just have two files and select between them with a config. option.  If you're really interested in editing the file directly, I'd recommend using ``dockertest.config.ConfigSection()`` as it's unit-tested well and saves you from having to deal with all the low-level IO.

afomm:
@cevich - OK. Removing the edit_unit_file and will go with two files.

afomm:
@edsantiago @cevich - made the suggested changes. Please review.

afomm:
@edsantiago @cevich - I see it is failing when used with docker. I'm fixing it & will push again.

afomm:
@edsantiago @cevich - it is now ready for review again.

subtests/docker_cli/systemd/p4321-docker-latest.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
edsantiago: You don't need two unit files. Please remove this one.

subtests/docker_cli/systemd/p4321-docker.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=docker-autotest systemd subtest. If you see this file on a live system it means something has failed very badly
+After=docker.service
```
edsantiago: I believe that this is the crux of the whole matter. You do not need this `After=` line. Just remove it, and all will be nice and simple again.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,89 @@
...
+from dockertest.output import mustpass
+from dockertest.containers import DockerContainers
+
+
+class systemd(Subtest):
+
+    def initialize(self):
+        super(systemd, self).initialize()
+        self.stuff['systemd_touchstone'] = False
+        if self.config['docker_path'] == '/usr/bin/docker':
```
edsantiago: Again, please remove. This is unnecessary, and it doesn't do what you think it does. (That is: docker_path will *always* be /usr/bin/docker. Never docker-latest. The docker/docker-latest magic happens in `/etc/sysconfig/docker`).

cevich:
Just ignore the ``Red Hat Internal Testing — ADEPT Fail`` for now.  I sent you an e-mail about it.

subtests/docker_cli/systemd/p4321-docker.service
```diff
@@ -0,0 +1,11 @@
...
+
+[Service]
+ExecStart=/usr/bin/docker run -p 4321:4321 --name p4321 p4321
+ExecStop=-/usr/bin/docker stop p4321
+ExecStop=/usr/bin/docker rm -f p4321
+
+[Install]
+WantedBy=multi-user.target
```
cevich: I wonder, does it even ever need to be installed?  IIRC aren't the test requirements just about starting/stopping it?  Maybe I'm remembering wrong.

subtests/docker_cli/systemd/p4321-server.py
```diff
@@ -0,0 +1,32 @@
...
+class forkingtcpserver(SocketServer.ForkingTCPServer):
+
+    address_family = socket.AF_INET
+    allow_reuse_address = True
+    socket_type = socket.SOCK_STREAM
+
+
+if __name__ == "__main__":
+    server = forkingtcpserver(('0.0.0.0', PORT), requesthandler)
+    server.serve_forever()
```
cevich: All this code looks great!  Beautiful!  I especially like the perfect use of whitespace.  Geeze, it almost looks like I wrote it myself :grinning:

subtests/docker_cli/systemd/systemd.py
cevich: Aye, the only reason that config. option exists is to allow building docker (which requires docker) and then pointing at the built binary instead of the installed one.  Basically Ed's right, you can completely forget that the option even exists :grinning:

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,89 @@
...
+        if exists(filepath):
+            unlink(filepath)
+        if self.stuff['systemd_touchstone']:
+            self.sysd_action('daemon-reload')
+
+    def sysd_action(self, action=None, sysd_unit=None):
+        self.stuff['systemd_touchstone'] = True
+        command = "systemctl " + action
+        if sysd_unit:
+            command += " " + sysd_unit
```
cevich: Minor issue:  Generally using the ``+`` token for string concatenation is frowned upon (for performance and developer cheast-beating reasons).  The "proper" way is to use ``str.format()`` or the old-style ``%`` style formatting/substitution.  Though the reasons are complicated, and personally I think for little/simple uses, the ``+`` way is more readable.  No need to change what you have.  Just FYI.

cevich: If you must know: The main argument against using ``+`` is it results in creating additional strings in memory, then immediately throwing them away.  Whereas the formatting operations only end up creating a single string in memory and possibly throwing away only the original reference (if one existed).

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,89 @@
...
+        utils.run(command, ignore_status=False)
+
+    @staticmethod
+    def read_socket(host, port):
+        num_read_try = 10
+        while num_read_try:
+            s = socket.create_connection((host, port), 13)
```
cevich: Good to stick a comment here about why ``13`` was chosen.

If you don't remember:  it's because it's 1-second longer than two DNS timeouts.

subtests/docker_cli/systemd/p4321-docker-latest.service
cevich: @edsantiago that's my fault.  I suggested having two so we can select between one for docker and another for docker-latest.  But you're right, we don't need two if ``After`` is removed.

subtests/docker_cli/systemd/p4321-docker.service
afomm: Fixed - it is now removed.

subtests/docker_cli/systemd/systemd.py
afomm: Thanks. I'll use str.format() in the future.

subtests/docker_cli/systemd/systemd.py
afomm: I made these changes and pushed.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,86 @@
...
+            s.settimeout(1.0)
+            try:
+                time_from_socket = s.recv(512).strip()
+                if time_from_socket:
+                    s.shutdown(socket.SHUT_RDWR)
+                    return time_from_socket
+                num_read_try -= 1
+            except socket.timeout:
+                pass
```
edsantiago: It might be nice to have a fallthrough here, perhaps raising DockerTestError with a helpful message indicating that we never got a response despite N retries.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,86 @@
...
+            time.sleep(0.1)
+        time_from_socket = self.read_socket(host, port)
+        self.failif(not time_from_socket.isdigit(),
+                    "Data received from container is non-numeric: '%s'"
+                    % time_from_socket)
+        self.sysd_action('stop', self.config['sysd_unit_file'])
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        cntr_name = DockerContainers(self).get_unique_name()
```
edsantiago: This does not do what you think it does. `get_unique_name()` just generates a pseudorandom string, it doesn't actually *use* it. Expected use is:

    myname = DockerContainers(self).get_unique_name()
    self.sub_stuff['myname'] = myname
    cmd = DockerCmd(self,'run',['--name', myname, etc, etc])

   ...then cleanup
   ...'rmi', self.sub_stuff['myname']

Calling `get_unique_name` in `cleanup` doesn't do anything, and calling `clean_all()` with that string is like calling `clean_all("this_is_a_completely_nonexistent_name")`

subtests/docker_cli/systemd/systemd.py
afomm: Thanks! Made these changes and pushed.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+        port = 4321
+        self.sysd_action('daemon-reload')
+        self.sysd_action('start', self.config['sysd_unit_file'])
+        while is_port_free(port, host):
+            time.sleep(0.1)
+        time_from_socket = self.read_socket(host, port)
+        self.failif(not time_from_socket.isdigit(),
+                    "Data received from container is non-numeric: '%s'"
+                    % time_from_socket)
+        self.sysd_action('stop', self.config['sysd_unit_file'])
```
cevich: If the ``failif()`` throws, the ``sysd_action('stop', ...)`` will not occur.  Then in ``cleanup()`` the unit file is removed, and daemon reloaded.  But does that mean the service could be left running or in some funky state?  I honestly don't know.

Maybe this is not a problem, @edsantiago what do you think?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+                                           self.config['name'], self.bindir])
+        mustpass(dkrcmd.execute())
+
+    def run_once(self):
+        host = 'localhost'
+        port = 4321
+        self.sysd_action('daemon-reload')
+        self.sysd_action('start', self.config['sysd_unit_file'])
+        while is_port_free(port, host):
+            time.sleep(0.1)
```
cevich: Just realized, this could be an infinite loop e.g. something funky happens with docker/networking.  It will eventually exhaust the autotest-level timeout - a condition which means the test gets killed w/o running ``cleanup()`` (IIRC).  Maybe trying 1000 times (10-seconds), then barfing would be better.  @edsantiago  what do you think?

subtests/docker_cli/systemd/systemd.py
edsantiago: @cevich *excellent* catch! Yes, it actually is a problem and it explains a failure I saw in stress tests yesterday. @afomm, the solution is to move the `sysd_action stop` to:

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+        self.failif(not time_from_socket.isdigit(),
+                    "Data received from container is non-numeric: '%s'"
+                    % time_from_socket)
+        self.sysd_action('stop', self.config['sysd_unit_file'])
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        DockerImages(self).clean_all([self.config['name']])
+        filepath = self.config['sysd_unitf_dest']
+        if exists(filepath):
```
edsantiago: ...here. That is: move `self.sysd_action("stop")` to here, right after `if exists`.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+            time.sleep(0.1)
+        time_from_socket = self.read_socket(host, port)
+        self.failif(not time_from_socket.isdigit(),
+                    "Data received from container is non-numeric: '%s'"
+                    % time_from_socket)
+        self.sysd_action('stop', self.config['sysd_unit_file'])
+
+    def cleanup(self):
+        super(systemd, self).cleanup()
+        DockerImages(self).clean_all([self.config['name']])
```
edsantiago: This needs to be moved further down, to after the `stop`. Otherwise the image is removed while a container is still running, which leaves `<none>`s.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+        filepath = self.config['sysd_unitf_dest']
+        if exists(filepath):
+            unlink(filepath)
+        if self.stuff['systemd_touchstone']:
```
edsantiago: While we're at it, I don't think this touchstone thing is necessary any more. Basically, in cleanup, run `daemon-reload` if filepath exists. Final cleanup could look like:

        super(systemd, self).cleanup()
        filepath = self.config['sysd_unitf_dest']
        if exists(filepath):
            self.sysd_action('stop', self.config['sysd_unit_file'])
            unlink(filepath)
            self.sysd_action('daemon-reload')
        DockerImages(self).clean_all([self.config['name']])

Sorry for the last-minute requests.

cevich: Yes, my apologies also.  Didn't want to take it through so many cycles, it's my fault I should have taken the bird's eye view of everything earlier.  Ed's suggested-solution LGTM and there's no shame in copy-paste (unless it's me that's doing it excessivly) :grinning:

subtests/docker_cli/systemd/systemd.py
edsantiago: @cevich out of curiousity, why not:

    utils.wait_for(lambda: is_port_free(port, host), 10)

cevich: an even better idea

subtests/docker_cli/systemd/systemd.py
```diff
@@ -0,0 +1,87 @@
...
+            try:
+                time_from_socket = s.recv(512).strip()
+                if time_from_socket:
+                    s.shutdown(socket.SHUT_RDWR)
+                    return time_from_socket
+                num_read_try -= 1
+            except socket.timeout:
+                pass
+        raise DockerTestError("Failed to read from socket after %s tries"
+                              % num_read_try)
```
afomm: Thanks a lot guys! I applied these changes and pushed.

cevich:
[(8) [CI]Add automated test that verifies Docker containers can be pull/run/start/stop as systemd services - 1198599](https://trello.com/c/V5BwwE62/63-8-ci-add-automated-test-that-verifies-docker-containers-can-be-pull-run-start-stop-as-systemd-services-1198599)

PR #637, opened 2016-09-13
Fix a few failing jobs in docker autotest CI / ADEPT

cevich:
N/B: There's an un-squashed **fixup!** here

### DIFF
--- a/config_defaults/defaults.ini
+++ b/config_defaults/defaults.ini
@@ -19,9 +19,6 @@ docker_path = /usr/bin/docker
 #: Global docker client command options to use (CSV)
 docker_options =
 
-#: Global docker daemon options to use (CSV)
-daemon_options = daemon,-D,--selinux-enabled
-
 #: Max runtime in seconds for any docker command (auto-converts to float)
 docker_timeout = 300.0
 
--- a/dockertest/subtestbase.py
+++ b/dockertest/subtestbase.py
@@ -10,6 +10,7 @@ modules.
 import logging
 import traceback
 from xceptions import DockerTestFail
+from xceptions import DockerTestNAError
 from config import get_as_list
 
 
@@ -45,6 +46,9 @@ class SubBase(object):
         "postprocess": "postprocess()",
         "cleanup": "cleanup()"
     }
+
+    #: Path to file indicating which Red Hat release this is
+    redhat_release_filepath = "/etc/redhat-release"
 
     def __init__(self, *args, **dargs):
         super(SubBase, self).__init__(*args, **dargs)
@@ -166,6 +170,25 @@ class SubBase(object):
                     return
         raise DockerTestFail("%s not in %s '%s'" %
                              (expect_s, description, haystack))
+
+    @classmethod
+    def failif_not_redhat(cls, xcept=DockerTestNAError):
+        """
+        Raises an exception if this is a non-RHEL/RHELAH System
+
+        :param xcept: Non-default exception to raise
+        :raise DockerTestNAError: By default
+        """
+        try:
+            with open(cls.redhat_release_filepath, 'rb') as rhrel:
+                if 'Enterprise' not in rhrel.read():
+                    raise xcept("Test only intended for a Red Hat "
+                                "Enterprise Linux System.")
+        except IOError, anotherone:
+            raise xcept("Is this a Red Hat Enterprise system?  "
+                        "Error reading %s: %s."
+                        % (cls.redhat_release_filepath,
+                           str(anotherone)))
 
     @classmethod
     def log_x(cls, lvl, msg, *args):
--- a/dockertest/subtestbase_unittests.py
+++ b/dockertest/subtestbase_unittests.py
@@ -5,6 +5,7 @@
 
 import os
 import sys
+from tempfile import mkstemp
 import types
 from unittest2 import TestCase, main
 
@@ -42,6 +43,7 @@ class DockerTestFail(Exception):
 
 # Mock module and exception class in one stroke
 setattr(mock('xceptions'), 'DockerTestFail', DockerTestFail)
+setattr(mock('xceptions'), 'DockerTestNAError', DockerTestFail)
 
 # END   boilerplate crap needed for subtests, which should be refactored
 ###############################################################################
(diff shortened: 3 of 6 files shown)


cevich:
Getting network errors for internal CI - centos atomic upgrade reporting 'Network is unreachable'

config_defaults/defaults.ini
```diff
@@ -19,9 +19,6 @@ docker_path = /usr/bin/docker
 #: Global docker client command options to use (CSV)
 docker_options =
 
-#: Global docker daemon options to use (CSV)
-daemon_options = daemon,-D,--selinux-enabled
```
edsantiago: Oh, good catch!

cevich:
I have a fixup!-in-waiting, watching CI results since they seem to be running now.

problem is improper class in a super() call for ``subtests/redhat/rhel_push_plugin/rhel_push_plugin.py``

dockertest/subtestbase.py
```diff
@@ -168,6 +172,21 @@ def failif_not_in(needle, haystack, description=None):
...
+
+        :param xcept: Non-default exception to raise
+        :raise DockerTestNAError: By default
+        """
+        try:
+            with open(cls.redhat_release_filepath, 'rb') as rhrel:
+                if 'Enterprise' not in rhrel.read():
+                    raise xcept("Expecting a Red Hat Enterprise Linux System")
+        except IOError:
+            cls.logwarning("Problem checking /etc/redhat-release")
```
edsantiago: Is this really wise? If `/etc/redhat-release` is unreadable/nonexistent/whatever, why would we want to proceed? And if we do proceed, shouldn't the warning at least be courteous enough to offer the reason it failed?

(and, if you keep this code, please use % cls.redhat_release_filepath)

subtests/redhat/rhel_push_plugin/rhel_push_plugin.py
```diff
@@ -24,6 +24,10 @@ class rhel_push_plugin(subtest.SubSubtestCaller):
 
     """ SubSubtest caller """
 
+    def initialize(self):
+        self.failif_not_redhat()
+        super(packaging, self).initialize()
```
edsantiago: Dontcha love python2 super...

cevich:
Internal CI shows this is doing what it's suppose to:

```
docker/subtests/redhat/rhel_push_plugin.57_1360555924   TEST_NA
```

dockertest/subtestbase.py
cevich: TBH, I'm not sure what to do in this case.  Figured I'd toss in a logging message so you'd notice and give me a better solution :grinning: 

Thinking about it more...it seems like in all cases, raising a ``TestNAError()`` (with a different message) is what it should do.  Ya?

subtests/redhat/rhel_push_plugin/rhel_push_plugin.py
cevich: Hrmmm, there's a subtle problem here.  I was hoping to have these tests not contribute toward a FAIL (in CI).  However, top-level TestNAErrors do have that effect :-1:  But moving this call into the sub-subtests feels wrong to me.

cevich:
Not sure if updating ``config_defaults/control.ini`` every time a ``redhat`` subtests is added makes sense (it would be a PITA).  Thinking the best thing is to work this into ADEPT's default ``control.ini.j2`` template instead.  It's really just the docker-autotest and ADEPT CI jobs the top-level TestNAErrors affect.

dockertest/subtestbase.py
edsantiago: Is this code ever going to run on non-{RH,CentOS,Fedora}? If so then yes, raising `TestNAError` is probably the right thing to do.

cevich:
The ``redhat/push_plugin`` test is doing the right thing now, but opened #638 for a ``redhat/packaging`` problem.  In any case, I'll squish this up and add any needed polish before force-pushing.

cevich:
Okay, looks like internal CI is happy with this.  @ed mind taking a last look before I merge?

cevich:
Okay, looks like internal CI is happy with this. @edsantiago  mind taking a last look before I merge?

dockertest/subtestbase.py
```diff
@@ -168,6 +172,23 @@ def failif_not_in(needle, haystack, description=None):
...
+        """
+        Raises an exception if this is a non-RHEL/RHELAH System
+
+        :param xcept: Non-default exception to raise
+        :raise DockerTestNAError: By default
+        """
+        try:
+            with open(cls.redhat_release_filepath, 'rb') as rhrel:
+                if 'Enterprise' not in rhrel.read():
+                    raise xcept("Expecting a Red Hat Enterprise Linux System")
```
edsantiago: So I'm still a little hazy on when/how the `redhat/*` tests are invoked. (I'm actually a little hazy on whether docker-autotest is used anywhere outside RH). Are redhat/* tests invoked on all runs? On some? Is it configurable somehow?

The reason it matters is that, IMO, if run on all docker-autotest invocations, a more suitable exception message might be "this is not a RHEL system, move along" or equivalent. As stated above, "Expecting" implies that this is an unexpected situation.

PS Oooooh, nice! github: "instead of commenting right away, your comments will be grouped"!

dockertest/subtestbase.py
```diff
@@ -168,6 +172,23 @@ def failif_not_in(needle, haystack, description=None):
...
+        except IOError, anotherone:
+            raise xcept("Error reading %s: %s"
+                        % (cls.redhat_release_filepath,
+                           str(anotherone)))
```
edsantiago: ...similarly here. If these tests are part of standard docker-autotest runs, a user on <non-RH-CentOS-or-Fedora> might prefer a message such as "This doesn't look like a Red Hat system".

dockertest/subtestbase_unittests.py
```diff
@@ -72,6 +74,36 @@ class DockerTestFail(Exception):
...
 
+class TestNotRedHat(TestCase):
+
+    def setUp(self):
+        import subtestbase
+        self.subtestbase = subtestbase
+        # Saves some typing
+        self.stbsb = self.subtestbase.SubBase
+        pfx = 'subtestbase_unitest'
+        (self.fd,
```
edsantiago: Why preserve fd? Making this `None` saves maintainer effort, and also lets you remove the `os.close()` in `tearDown`.

dockertest/subtestbase.py
cevich: Everything runs on everything by default.  So the TestNA's are important.

Controlling this is really up to the caller and configuration.  So for CI, I have some changes staged (in ADEPT) that modify the ``config.ini`` exclude list based on the type of host (i.e. atomic or regular) and product (rhel or non).

dockertest/subtestbase_unittests.py
cevich: Nope, then it would leak a file (at least until the unittest process exited.)  Docs state it's the caller's responsibility to close.

dockertest/subtestbase.py
cevich: S'okay, I prefer your wording, let me try fixing it...

dockertest/subtestbase_unittests.py
edsantiago: OK, then how about closing immediately after `mkstemp`? 

What bothers me is keeping that fd while all the subsequent code does reopens.

cevich: Ahh sure, but it's not very faithful to the unittesting spirit if "peeking under the covers" is allowed :grinning:

cevich:
Oh, durr, totally missed that the file is just re-opened in the test method.  Okay, there, that's better now.  Sorry, it's been a loooooooooong day :confounded:

edsantiago:
Sorry, I missed that there was a push. LGTM, merging.

PR #639, opened 2016-09-14
packaging test: RHEL Atomic doesn't have magic.py

edsantiago:
Let's try invoking 'file' directly.


edsantiago:
This of course assumes that Atomic has `file`. I don't currently have a host to test with.

cevich:
Heh, I love these simple fixes :+1:

cevich:
> This of course assumes that Atomic has file. I don't currently have a host to test with.

Yes it does.

PR #640, opened 2016-10-03
Fix image counting in build subtest

cevich:
Subtests cannot assume the default test image exists or not at the start
of any test.  If a part of a build sub-subtests builds an image that
includes pulling the default test image, and uses the img_count
postprocessing command, the count will be off-by-one.

Fix this condition by excluding appearance of the any image that looks
like the default image, from the before/after build comparison.

N/B: This _does_not_ fix a possible mis-count condition where the pulled
dependent image, only sometimes matches the default image (by
configuration.  In these situations, the test configuration needs to be
modified to account for the difference (one way or another).


edsantiago:
Doesn't 091096b6b16fed157e93d2e174ae37f444f15974 guarantee that the default test image is pulled in?

cevich:
> Doesn't 091096b guarantee that the default test image is pulled in?

Only for the first test really.  There are tests that delete the default image, so I'm not sure it's an assumption that's safe to make.

cevich:
Re: the miscount when build base-image != default image

This needs to be addressed in configuration.  So for ADEPT, I'll put in a change to the ``subtests.ini`` template that should take care of this: https://github.com/cevich/autotest-docker-enabled-product-testing/pull/58

edsantiago:
Looks OK, but that's a lot of hoops to jump through. Much more so with that ADEPT pr. Did you consider adding something like this to build.initialize() ?

    base_img = self.config['base_img'] or di.default_image    # base_img defined in https config
    DockerCmd(self, 'pull', base_img)

cevich:
> Did you consider adding something like this to build.initialize() ?

I did not consider that.  Wouldn't it mean using it for __all__ the non-local builds? i.e. editing multiple config. sections and sync'ing with the remote dockerfile.  Oh, in fact, I forgot to consider Fedora, let me include that as well...

edsantiago:
I'm not sure how to parse that, but my thinking is that if a config.ini file references a Dockerfile that in turn references image `foo`, the config.ini file would be a nice place to declare `base_img = foo` as well. Makes it easy for the reader and maintainer. Or at least easier, I think, than special centos-fedora-rhel hardcoding in ADEPT. Does that make sense?

cevich:
Ya, makes sense.

What I mean is, e.g. for build_git or build http or whatever other sub-subtest builds from a remote dockerfile.  They'd all need to use a ``base_image = foo``.

Anyway, you're probably right, as it would keep all the special-case-ness coded into the config, rather than the test.  Unf. Red Hat Internal CI bustid ATM.  So I got to fix that first before I go further on this and the ADEPT PR :frowning_face:

cevich:
jenkins, test this please.

edsantiago:
LGTM, and I see you even fixed the commit message.

cevich:
OIC why it's failing, it's building from 'https://raw.githubusercontent.com/autotest/autotest-docker/master/subtests/docker_cli/build/full/Dockerfile' which is still using the old ``:centos7`` tag (this PR fixes that).

PR #641, opened 2016-10-03
Skip failing tests on docker 1.10

edsantiago:
rhbz#1330224 causes failures in docker-1.10 but is fixed
in 1.12 (it was never a problem pre-1.10). The bug will
not be fixed in any RH build of docker-1.10, so let's just
skip it to make CI happy.


cevich:
Uggg, sorry, dunno why github didn't notify me about this.

edsantiago:
Oh, no worries. I thought you were working on an ADEPT-based solution.

cevich:
Nope, been trying to fix the ADEPT and Docker Autotest CI jobs.  Looking at the code here though, I wonder if this isn't a better solution.  Mainly because we can just drop it after 1.10 is no longer anywhere to be seen.

cevich:
Haha! Will you look at that, it's running and not blocking the Jenkins Queue :confounded:

cevich:
CI is happy with this.  ``From 115 tests executed, 110 have passed (pass rate of 95.0%)`` but two of those are ``TestNA``'s (marked as failure by Jenkins) and the start test is green now  :grinning:

PR #645, opened 2016-10-13
dockerinspect, more Labels key filtering needed

cevich:
The tree for image JSON, under 'ContainerConfig' also needs
'Labels' key excluded from checking.  These are build-time
keys, basically arbitrary key's / values that can not and
should not ever be validated.


edsantiago:
Much nicer than hardcoding CentOS exceptions, thank you!

cevich:
Heh, I'm sure ___"they"___ will find a way to break this test in...oh...four weeks :grinning:

cevich:
Wow, look at that 

```
Oct 13 14:47:42 	docker/subtests/docker_cli/dockerinspect.10_372443729 	
GOOD
	2
```

Completed and passing in TWO seconds :grinning:

PR #646, opened 2016-10-14
Fix docker_test_images clobber defaults.ini

cevich:
For whatever reason, dockertest.config.ConfigSection.merge_write()
cannot properly merge with the 'DEFAULTS' section.  Add a note/comment
to this effect, and update docker_test_images to perform the merge
with currently loaded defaults.  Also append a comment to the
updated file indicating the source of the modifications.


cevich:
(I'm lazy, that's why I didn't fix ``dockertest.config.ConfigSection.merge_write()`` and add a unittest for this)

cevich:
@edsantiago @afomm Could take a quick double-check for me?  This fixes a problem that's causing silent, wide-spread erroneous testing.  Thanks.

edsantiago:
I've been running it and testing it on my 7.3 virt. Seems to be working, at least as tested by `cat tests/docker/config_custom/defaults.ini`, but I'm having a hard time understanding the background of the problem. Still looking.

cevich:
re: cat test

Jeeze man, that's the same way I tested it :smirk_cat: 

background: w/o this fix, if this pretest runs using all default options, it will clobber your ``config_custom/defaults.ini`` with a single DEFAULTS item, 'preserve_fqins = foo'.

The purpose of the 'build' feature is for use in: https://trello.com/c/YiuGqB1m

edsantiago:
So, the new code does not preserve `__example__`. Is that important?

cevich:
> So, the new code does not preserve __example__. Is that important?

IIRC, that should always get filtered out.  ``__example__`` is only useful in the context of ``config_defaults``.  It's used there to flag extra checking that certain key's values are different from the defaults.  I'll double check my memory on this...

cevich:
So the way it works is, when a subtest is being initialized, setting up it's config, it checks if anything is in the ``__example__`` key and if so, warns about it.  So all the checking happens on load in ``config.config().load_config_sec()``.  I'll need to read it more carefully, but it appears that the only ``__example__`` values that are loaded come from the ``config_defaults`` files.

dockertest/config.py
```diff
@@ -203,6 +203,7 @@ def merge_write(self, fileobject):
         scp = SafeConfigParser()
         # Safe if file doesn't exist
         scp.read(fileobject.name)
+        # N/B: This won't work with DEFAULTS
```
edsantiago: Do you know why? If so can you explain it in the comment? I really don't understand this. `ConfigParser` special-cases DEFAULT (singular), but I don't see why the plural should also be affected.

cevich: I don't, I looked at it briefly but couldn't figure it out.  It may be something in the ``ConfigParser`` module.

PR #647, opened 2016-10-18
run_cgroups : behavior change in RHEL 7.3

edsantiago:
RHEL 7.3 appears to change the behavior of cgroups when
docker is invoked with "-m 0". Given these commands:

    # img=$(docker run -d -i -m 0 centos /bin/bash)
    # cat /sys/fs/cgroup/memory/system.slice/docker-$img.scope/memory.limit_in_bytes

...we seem to get the following results in different RHELs:

    7.2 : 9223372036854775807    ! 0x7FFFFFFFFFFFFFFF
    7.3 : 9223372036854771712    ! 0x7FFFFFFFFFFFF000

Deal with either, and explain why, and use a hex constant
for readability.


edsantiago:
NOTE: I have not tested this on RHEL 7.2, because I don't currently have a virt setup. Am hoping ADEPT will test it.

cevich:
Interesting, exactly one page (4096 bytes) difference. Maybe there was an off-by-one bug fixed somewhere?

cevich:
There was a bug on this test a loooooong time ago, opened by it's original author.  There's some maths stuff going on here that escapes my memory.  Trying to find the bug...

cevich:
Here it is: https://bugzilla.redhat.com/show_bug.cgi?id=1101486

cevich:
Ahh, better one: https://bugzilla.redhat.com/show_bug.cgi?id=1101943

edsantiago:
OK, from reading those, I get the sense that the "-m 0" test is poorly implemented: it should instead compare the memory.limit_in_bytes value from "-m 0" against docker with no "-m" argument, and make sure they're the same. 

But that's weeks of work (the whole test needs revamping) and right now I just want to get rid of a cry-wolf bug. Is it OK to just merge this to get our tests passing?

cevich:
I agree the test is poorly implemented, it's very old, and dates to a time where our mission was quantity over quality.    However, I'm a little afraid about tweaking the test to pass, in case there really is a problem here.  I just don't know/remember enough to be sure one way or another.

In other words, the numerical representation of "unlimited" and "a very large number" may both be large enough that from many perspectives, they are equivalent.  Until some rare instance comes along 5 years from now (assuming no other change) and all of a sudden "big number" != "unlimited" any more, and "bad stuff" ensues.

Do you think it worth worth opening a low-severity bug against this behavior change to double-check with kernel eng.?  

That at least allows skipping the test (by whiteboard) until we have an answer one way or another.  Worst case, it's intended behavior and we can just merge this.  Best case, there's a real problem that needs fixing.

Or maybe am I just being paranoid?

cevich:
Side-note: I fixed the Jenkins-killing problem the other week, so asked the CI job for docker-autotest be re-enabled.

edsantiago:
I would really prefer not to spend much time on this. From my perspective, this is what has changed:

    Between RHEL 7.3 and 7.3, the kernel changed the definition of "unlimited memory".

If the `run_cgroups/memory_no_cgroup` test were correctly implemented, as in, comparing `-m 0` against the no-minus-m case, we wouldn't be having this conversation: the limit values would match, and docker-autotest would not report anything. 

The proper solution is for us to fix docker-autotest, remove the hardcoding. This seems especially pressing given what looks like upcoming changes in kernel 3.12 (RHEL 7.3 is 3.10):

    https://www.redhat.com/archives/libvir-list/2013-December/msg00435.html

I've filed issue #648

cevich:
Thanks for pointing it out again.  It just computed, about the number matching the no-m case, sorry, little slow today.  Yeah, band-aid application + new issue is good 'nuf.  Thanks for your patients.

PR #649, opened 2016-10-18
Replace run_ping with run_install

cevich:
With some base images (those lacking iputils), the run_ping test must
perform two operations.  Both are required for the sub-subtest to
pass, which can make it difficult to identify higher-level repo. /
install problems vs ping-related problems.

This commit adds a new sub-subtest, and a new option ``install_cmd``.  When
non-empty, this command will be logically ANDed along with ``cmd``.
A simplistic check for ``bash -c`` was added since ``&&`` is
shell-specific.

The new ``run_install`` test uses the ``install_cmd`` twice.  Once to
verify that installation works, including with existing packages.
The run_ping test was updated to use this option instead of jamming both
commands together on it's own.  Though the eventual docker command is
unchanged, it makes the configuration cleaner and more modular.

### DIFF
--- a/config_defaults/subtests/docker_cli/run.ini
+++ b/config_defaults/subtests/docker_cli/run.ini
@@ -1,5 +1,5 @@
 [docker_cli/run]
-subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_tmpfs,run_passwd,run_ping
+subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_tmpfs,run_passwd,run_install
 #: The most basic sub-subtests can be generated from a generic
 #: class at runtime, set 'yes' to enable this feature.
 generate_generic = no
@@ -19,7 +19,6 @@ exit_status = 0
 [docker_cli/run/run_true]
 generate_generic = yes
 cmd = /bin/true
-exit_status = 0
 
 [docker_cli/run/run_false]
 generate_generic = yes
@@ -31,15 +30,6 @@ generate_generic = yes
 #: String to search for with ``grep`` (see ``cmd`` option value)
 expected_status = "Password set"
 cmd = 'passwd --status root | grep -qv %(expected_status)s'
-exit_status = 0
-
-[docker_cli/run/run_ping]
-generate_generic = yes
-__example__ = ping_url
-#: Host to ping
-ping_url = www.google.com
-cmd = 'ping -c 10 -q %(ping_url)s'
-exit_status = 0
 
 [docker_cli/run/run_names]
 cmd = sleep 2s
@@ -52,14 +42,12 @@ run_append_name = no
 last_name_sticks = yes
 #: The number of ``--name`` options to generate for the command line.
 names_count = 1000
-exit_status = 0
 
 [docker_cli/run/run_tmpfs]
 #: --tmpfs parameter
 tmpfs_path = /run
 run_options_csv = --tmpfs,%(tmpfs_path)s
-cmd = findmnt -n -t tmpfs %(tmpfs_path)s
-exit_status = 0
+cmd = 'findmnt -n -t tmpfs %(tmpfs_path)s'
 
 [docker_cli/run/run_interactive]
 run_options_csv = --interactive,--rm
@@ -75,7 +63,6 @@ secret_sauce = 4c93bb78d98f
 # Yes it's ugly / crude, but it's simple and works in all but most
 # heavily loaded cases.
 cmd = 'sleep 5s && echo "%(secret_sauce)s"'
-exit_status = 0
 
 [docker_cli/run/run_remote_tag]
 #: Change this to an image remotely available within test environment
@@ -85,3 +72,9 @@ docker_timeout = 60
 remote_image_fqin = stackbrew/centos:7
 run_options_csv =
 cmd = /bin/true
+
+[docker_cli/run/run_install]
+#: Full command to execute which installs something inside the container
+install_cmd = yum install --disablerepo="*" --enablerepo="base" --assumeyes yajl
+#: Full command to run in commited container to verify installation
+verify_cmd = bash -c "echo '[{}]' | json_verify -q"
--- a/subtests/docker_cli/run/run_install.py
+++ b/subtests/docker_cli/run/run_install.py
@@ -1,0 +1,33 @@
+from run import run_base
+from dockertest.dockercmd import DockerCmd
+from dockertest.output.validate import mustpass
+
+
+class run_install(run_base):
+    """Verify installing packages in a container is functional"""
+
+    def init_subargs(self):
+        cont = self.sub_stuff['cont']
+        # Name will be used for image, must be lower-case
+        self.sub_stuff['name'] = name = cont.get_unique_name().lower()
+        self.sub_stuff['subargs'] += ["--name %s" % name,
+                                      self.sub_stuff['fqin'],
+                                      self.config['install_cmd']]
+
+    def postprocess(self):
+        super(run_install, self).postprocess()
+
+        name = self.sub_stuff['name']
+        # ancestor method must have been successful
+        self.sub_stuff['containers'].append(name)
+
+        # images with same names as containers are confusing
+        eman = name[-1::-1]  # the name, backwards
+        mustpass(DockerCmd(self, 'commit',
+                           [name, "%s:latest" % eman]).execute())
+        # mustpass() was successful
+        self.sub_stuff['images'].append(name)
+
+        eman = name[-1::-1]  # committed image
+        subargs = ['-i', '--rm', eman, self.config['verify_cmd']]
+        mustpass(DockerCmd(self, 'run', subargs).execute())


cevich:
woopsie, forgot ``yum ... ``__INSTALL__``...``  :grinning:

cevich:
jenkins, test this please.

config_defaults/subtests/docker_cli/run.ini
```diff
@@ -85,3 +75,29 @@ docker_timeout = 60
...
+[docker_cli/run/run_install]
+generate_generic = yes
+__example__ = base_repo_name
+run_options_csv = -i,--rm
+#: Repo./Channel name to enable for yum commands
+base_repo_name = base
+#: Packages to install
+install_rpms = iputils
+#: Command to run to install rpms (yum or dnf)
+install_cmd = yum --disablerepo "*" --enablerepo "%(base_repo_name)s" install -y %(install_rpms)s
```
edsantiago: You'll need to wrap this in single-quotes. The next one (line 100) too.

cevich:
Note: CI is going to fail on the ping test b/c: https://github.com/cevich/autotest-docker-enabled-product-testing/pull/60

config_defaults/subtests/docker_cli/run.ini
cevich: Tried that, won't work b/c then ``install_cmd`` and ``cmd`` will be under different quoting.  Instead, I removed all quoting, and just do it in the test-code instead.  If that doesn't work...then we can come up with something else.

subtests/docker_cli/run/run.py
```diff
@@ -46,7 +46,14 @@ def init_subargs(self):
...
         subargs += get_as_list(self.config['bash_cmd'])
-        subargs.append(self.config['cmd'])
+        install_cmd = self.config['install_cmd'].strip()
+        if install_cmd:
+            subargs.append("'%s && %s'" % (install_cmd, self.config['cmd']))
+            if 'bash' not in subargs and '-c' not in subargs:
+                self.logwarning("Using install_cmd && cmd w/o bash -c could "
+                                "lead to cancer: %s", subargs)
+        else:
+            subargs.append("'%s'" % self.config['cmd'])
```
cevich: Here, just add quotes how they're needed.  Also added a comment in config, mentioning that single-quotes will be used.

edsantiago:
Weird: it worked on my test setup. Pulling now & rereviewing. (But I still think this test should be disabled in RHEL)

subtests/docker_cli/run/run.py
```diff
@@ -46,7 +46,16 @@ def init_subargs(self):
...
-        subargs.append(self.config['cmd'])
+        install_cmd = self.config['install_cmd'].strip()
+        cmd = self.config['cmd'].strip()
+        if install_cmd:
+            subargs.append("'%s && %s'" % (install_cmd, cmd))
+            if 'bash' not in subargs and '-c' not in subargs:
+                self.logwarning("Using install_cmd && cmd w/o bash -c could "
+                                "lead to cancer: %s", subargs)
+        # Don't try to run empty commands
+        elif not cmd:
```
edsantiago: `not`?

cevich:
That's why I'm relying on CI to verify the changes.  The setup is the same no matter who sends the PR.  

This new ``run_install`` test was always an implied part of ``run_ping``, so it's good to move it out and have it verified separately.

The ``run_ping`` test previously discovered a nasty [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1142311) in docker.  Then it refuted a report that there was a regression on the original fix almost a year later.  So we need both tests, and we need them to pass.

subtests/docker_cli/run/run.py
cevich: oops

cevich:
> Can you get this working on RHEL

That's certainly my goal, and for installing packages too.  It's all about getting the repos. correctly disabled/enabled, and dealing with the MTU bug.

cevich:
OIC, it's barfing now b/c quotes in config. https://github.com/cevich/autotest-docker-enabled-product-testing/pull/60 gets rid of.  Okay, need to focus on that PR now then since it's got fixes for the "Wednesday Shenanigans" and also adds proper swap setup (needed to fix other tests).

edsantiago:
It's not just the quotes. There's the `--enablerepo=base` which fails with "repository not found"

cevich:
> There's the --enablerepo=base which fails with "repository not found".

Ahh right, thanks for reminding me.  I need to fix that in the ADEPT template used to render the test configuration (since it's different for RHEL).

subtests/docker_cli/run/run.py
```diff
@@ -46,7 +46,16 @@ def init_subargs(self):
         fqin = self.sub_stuff['fqin']
         subargs.append(fqin)
         subargs += get_as_list(self.config['bash_cmd'])
-        subargs.append(self.config['cmd'])
+        install_cmd = self.config['install_cmd'].strip()
```
edsantiago: So, taking a step back, does `install_cmd` actually give any benefit? It seems to increase code complexity and break quoting (fixed here in `run.ini` but leaves existing `config_custom` files which will need to be edited). It looks to me like the approach started off worthwhile but didn't pan out. What would you think of reverting back to just simple `cmd`?

cevich:
@edsantiago squash commit totally kills run_ping and replaces with non-generic run_install sub-subtest.  Far simpler this way IMHO.  Thrown together and untested..  Waiting on CI for that.

subtests/docker_cli/run/run_install.py
```diff
@@ -0,0 +1,29 @@
...
+
+class run_install(run_base):
+    """Verify installing packages in a container is functional"""
+
+    def init_subargs(self):
+        cont = self.sub_stuff['cont']
+        self.sub_stuff['name'] = name = cont.get_unique_name()
+        self.sub_stuff['subargs'] += ["--name %s" % name,
+                                      self.sub_stuff['fqin'],
+                                      self.sub_stuff['install_cmd']]
```
edsantiago: You've probably realized by now that this should be `config`

edsantiago: I'm seeing a hang in `run_once()`

cevich: nope, was out to lunch, so thanks!

subtests/docker_cli/run/run_install.py
```diff
@@ -20,10 +20,13 @@ def postprocess(self):
         # ancestor method must have been successful
         self.sub_stuff['containers'].append(name)
 
-        mustpass(DockerCmd(self, 'commit', [name, name]).execute())
+        # images with same names as containers are confusing
+        eman = name[-1::-1]
```
edsantiago: Cute, but causes suffering to the maintenance programmer.

Also, please use `tolower()` to make sure there are no caps in the name. Caps lead to a docker barf on commit.

cevich: hrmmm, I'll double-check, but pretty sure ``get_unique_name()`` always does lowercase...

cevich: ...ahh, you're right.  It's done for images but not for container names.  good catch.

cevich:
@edsantiago thanks for your help, all the ``run/run_*`` tests are passing now, including the new one:

```
10/25 15:48:13 DEBUG|subtestbas:0202| 	run: Instantiating sub-subtest: docker_cli/run/run_install
10/25 15:48:13 INFO |subtestbas:0202| 		run_install: initialize()
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: WARNING: Recommended options not customized:
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: WARNING: docker_repo_tag
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: WARNING: docker_repo_name
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: WARNING: preserve_cnames
10/25 15:48:13 WARNI|subtestbas:0202| 		run_install: WARNING: Test results may be externally dependent! (See debug log for details)
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: run_install configuration:
...cut...
10/25 15:48:13 INFO |subtestbas:0202| 		run_install: run_once()
10/25 15:48:13 DEBUG|subtestbas:0202| 		run_install: Executing Command: /usr/bin/docker run --name run_install_efbl docker.io/stackbrew/centos:7 yum install --disablerepo="*" --enablerepo="base" --assumeyes yajl
10/25 15:48:18 INFO |subtestbas:0202| 		run_install: postprocess()
10/25 15:48:18 DEBUG|subtestbas:0202| 		run_install: Executing Command: /usr/bin/docker commit run_install_efbl lbfe_llatsni_nur:latest
10/25 15:48:20 DEBUG|subtestbas:0202| 		run_install: Executing Command: /usr/bin/docker run -i --rm lbfe_llatsni_nur bash -c "echo '[{}]' | json_verify -q"
10/25 15:48:20 INFO |subtestbas:0202| 		run_install: cleanup()
10/25 15:48:20 DEBUG|subtestbas:0202| 		run_install: Cleaning run_install_efbl
10/25 15:48:20 DEBUG|subtestbas:0202| 		run_install: Cleaning run_install_efbl
```

Good to merge?

edsantiago:
Not passing on my system, and I don't think it ever will, but if this will pass in ADEPT and we can get rid of the cry-wolf, I'm all for merging.

cevich:
> Not passing on my system

Hmmm.  Any idea why?  The Dockerautotest CI job is running on CentOsAtomic + centos base image.  Are you testing it on RHEL perhaps?

edsantiago:
Of course I'm testing it on RHEL.

cevich:
> Of course I'm testing it on RHEL.

Ahh okay, that needs a different configuration.  Can't use "Base" as the channel name, have to use "rhel-7-server-rpms" instead.  That's taken care of (in theory still) w/in ADEPT.   I'm still debating w/n to add a pair of RHEL hosts to the CI jobs.

edsantiago:
That seems to imply that we're not already testing RHEL, which kind of alarms me.

cevich:
We are, just further down the line.  I was referring to the ADEPT and Docker Autotest CI jobs.  They're separate from the regular (internal) jobs that run.  They only run on CentOS Atomic, but I'm testing an extension that will run both docker-current and docker-latest on them.

PR #652, opened 2016-10-24
bz1385924 workaround: run_cgroups_parent 'pids'

edsantiago:
...is broken in docker-1.10 on RHEL 7.3. It's fixed in 1.12
but won't be in any 1.10 build. So, don't test the pids
cgroup when docker server is 1.10.


cevich:
Bahh, kommandir ran out of space.  Fixing...

cevich:
jenkins, test this please.

subtests/docker_cli/run_cgroup_parent/run_cgroup_parent.py
```diff
@@ -110,6 +110,10 @@ def postprocess(self):
                     raise DockerTestFail("cgroup line does not conform to"
                                          " '<n>:<controller>:<path>': '%s'"
                                          % line)
+                # bz1385924: 'pids' fails in docker-1.10; not worth fixing.
+                if m.group(2) == 'pids':
```
cevich: If one controller was added, another could be as well (net, user, etc.).  What's your thought on having a ``controller_skip_csv=pids`` config. option?  Not necessary ATM to be sure, but down the line it could save somebody else a bunch of time.

edsantiago: It depends on what the future behavior is: will later docker handle the new controllers, or not? Because right now we need `controller_skip_if_docker_1_10_csv=pids` or `controller_skip_csv=pids_but_only_if_docker_1_10`. I'm reluctant to spend much time trying to second-guess the possible ways this can break in the future.

cevich: Oh your right, it's per-version as well.  Yeah, forget it.

PR #653, opened 2016-10-26
Add pulling and building image to container as systemd service

afomm:
Pull image and build image using systemd

Signed-off-by: Afom Michael tmichael@redhat.com


subtests/docker_cli/systemd/systemd.py
```diff
@@ -5,58 +5,42 @@
...
+class systemd_base(SubSubtest):
+
+    def initialize(self):
+        super(systemd_base, self).initialize()
+        self.sub_stuff['dirname'] = self.parent_subtest.bindir
+        self.sub_stuff['unit_src'] = \
+            '{}/{}'.format(self.sub_stuff['dirname'],
+                           self.config['sysd_unit_file'])
+        shutil.copyfile(self.sub_stuff['unit_src'],
+                        self.config['sysd_unit_dest'])
```
edsantiago: Duplication: each sysd_unit_dest is exactly `/etc/systemd/system/` + `sysd_unit_file`. You can save that duplication by removing all `sysd_unit_dest` config entries, and making the above 'os.path.join("/etc/systemd/system", self.config['sysd_unit_file'])`

subtests/docker_cli/systemd/systemd_build.py
```diff
@@ -0,0 +1,28 @@
...
+        for k, v in var_to_edit.items():
+            self.init_unitfile(k, v)
+
+    def run_once(self):
+        self.sysd_action('daemon-reload')
+        self.sysd_action('start', self.config['sysd_unit_file'])
+
+    def cleanup(self):
+        super(systemd_build, self).cleanup()
+        DockerImages(self).clean_all([self.config['image_name']])
```
edsantiago: Duplication: each `cleanup` is identical. Why not move them to the parent?

subtests/docker_cli/systemd/systemd_build.py
```diff
@@ -0,0 +1,28 @@
+"""
```
edsantiago: Nice breakdown. Well done. But `build` and `pull` are so small, do they really deserve a separate file for each? They could easily fit in `systemd.py`. I agree that `run` is longer and more complicated, so it deserves a separate file. @cevich , thoughts?

subtests/docker_cli/systemd/systemd_build.py
```diff
@@ -0,0 +1,28 @@
...
+Build docker image using Dockerfile and systemd unit file -
+``docker-build.service`` is used. This unit file is copied to system and
+minor edition is made. A systemd service is started using this file to build
```
edsantiago: edit, not edition. "edition" is something else completely.

afomm: Good catch! Thanks for reviewing even the description.

subtests/docker_cli/systemd/systemd_build.py
afomm: And use something like image_name for all three in the ini file instead of name, image_name, and image_to_pull?

subtests/docker_cli/systemd/systemd_build.py
afomm: They will get even smaller when I move/fix the cleanup.

afomm:
Hi - I made changes based on Ed's feedback. Please review.

edsantiago:
Sorry for the lack of comment. I've been looking at it this morning, running it on docker-1.12 (pass), but  also been busy testing the new docker-1.12.3-1 build.

afomm:
@edsantiago - np. I know busy day.

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -1,8 +1,17 @@
 [docker_cli/systemd]
-docker_timeout = 300
+subsubtests = systemd_pull, systemd_build , systemd_run
+
 #: Systemd unit file
+sysd_unit_file = docker-pull.service
+#: image name to pull, build, run
+image_name = rhel7:latest
```
cevich: This could break on non-RHEL systems (i.e. Fedora or CentOS) depending on how subscriptions/entitlement enforcement is worked out.  This is a problem for fixing on another day.

subtests/docker_cli/systemd/systemd_build.py
cevich: Yes, the convention (by other tests) is to just bundle the small ones in the main module file.

subtests/docker_cli/systemd/systemd_build.py
```diff
@@ -0,0 +1,23 @@
...
+
+from systemd import systemd_base
+from dockertest.images import DockerImages
+
+
+class systemd_build(systemd_base):
+
+    def initialize(self):
+        super(systemd_build, self).initialize()
+        DockerImages(self).clean_all([self.config['image_name']])
```
cevich: This should be assumed (clean environment).  It keeps the writers of ``cleanup()`` methods on their toes :grinning:

subtests/docker_cli/systemd/systemd.py
```diff
@@ -65,22 +52,16 @@ def sysd_action(action=None, sysd_unit=None):
...
-                if time_from_socket:
-                    s.shutdown(socket.SHUT_RDWR)
-                    return time_from_socket
-                num_read_try -= 1
-            except socket.timeout:
-                pass
-            time.sleep(1.0)
-        raise DockerTestError("Failed to read from socket with %d tries"
-                              % max_read_tries)
+    def init_unitfile(self, separator=None, changeto=None):
```
cevich: Any parameter required to be a non-None value, shouldn't both default to None and not be checked if it is None (before use).  How about ``separator='='``?

cevich: Also, is there ever a case where ``changeto`` should really ever be None?  Maybe switch the parameters around (so changeto is required):

``def init_unitfile(self, changeto, separator='='):``

subtests/docker_cli/systemd/systemd.py
```diff
@@ -65,22 +52,16 @@ def sysd_action(action=None, sysd_unit=None):
...
+        tmp_unitfile = '{}.{}'.format(self.sub_stuff['unit_dest'], "tmp")
+        with open(self.sub_stuff['unit_dest'], "r") as srcfile:
```
cevich: If test dies right here, it will leak a ``/etc/systemd/system/blah.tmp`` file that's never cleaned up.  It would be better to make the file in ``self.tmpdir`` and then ``os.rename()`` it into place.  Anything in ``self.tmpdir`` is automatically cleaned up by autotest.

An alternate approach is to read the entire file into a buffer (a list of lines), modify the buffer, and ``file.writelines()`` it out in place (instead of copying it).  There's a lot less disk I/O involved doing it this way, though you wouldn't really notice with such a small file.

Though for the purposes of this test, either way is fine.

subtests/docker_cli/systemd/systemd_build.py
```diff
@@ -0,0 +1,23 @@
...
+        var_to_edit = {'NAME=': self.config['image_name'],
+                       'DOCKERFILEDIR=': self.sub_stuff['bindir']}
+        for k, v in var_to_edit.items():
```
cevich: I think pylint or pep8 will get mad at you for using one-letter variable names.

subtests/docker_cli/systemd/systemd_pull.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def initialize(self):
+        super(systemd_pull, self).initialize()
+        # DockerImages(self).clean_all([self.config['image_to_pull']])
+        DockerImages(self).clean_all([self.config['image_name']])
+        # TODO
+        # set NAME in docker-pull.service to name provided in systemd.ini
+        # cfgsec = ConfigSection(dict(), 'docker_cli/systemd')
+        # cfgsec.set('ExecStart', cfgsec.get('ExecStart') % imagename)
+        # temp hack using init_unitfile but should use cfgsec.set()
```
cevich: Naa, what you have here is fine.

subtests/docker_cli/systemd/systemd_pull.py
```diff
@@ -0,0 +1,28 @@
...
+from dockertest.images import DockerImages
+# from dockertest.config import ConfigSection
+
+
+class systemd_pull(systemd_base):
+
+    def initialize(self):
+        super(systemd_pull, self).initialize()
+        # DockerImages(self).clean_all([self.config['image_to_pull']])
+        DockerImages(self).clean_all([self.config['image_name']])
```
cevich: Some reason ``clean_all()`` needs to happen twice (once in ancestor's ``initialize()`` and another here)?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -5,58 +5,45 @@
...
+class systemd_base(SubSubtest):
+
+    def initialize(self):
+        super(systemd_base, self).initialize()
+        self.sub_stuff['bindir'] = self.parent_subtest.bindir
+        self.sub_stuff['unit_dest'] = join("/etc/systemd/system",
+                                           self.config['sysd_unit_file'])
+        shutil.copyfile(join(self.sub_stuff['bindir'],
+                             self.config['sysd_unit_file']),
+                        self.sub_stuff['unit_dest'])
```
cevich: It's a little confusing that the test copies the unitfile into place, but then the build test re-writes it into a temp file, and copies it into place again.  Maybe move this ``copyfile()`` into the sub-subtest's ``initialize()`` right after they call ``super(...)``?

subtests/docker_cli/systemd/systemd_build.py
afomm: Changing before anyone gets mad :-)

config_defaults/subtests/docker_cli/systemd.ini
edsantiago: I like using `busybox`. It's always there, and it's very small (1MB) so it's a quick pull.

subtests/docker_cli/systemd/docker-build.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=docker-autotest systemd build image subtest. If you see this file on a live system it means something has failed very badly
+
+[Service]
+Type=oneshot
+Environment=NAME=rhel7_p4321
```
edsantiago: Since this is going to be altered anyway, how about `NAME=this-will-be-replaced`?

subtests/docker_cli/systemd/docker-build.service
```diff
@@ -0,0 +1,10 @@
...
+Environment=OPTIONS=--force-rm
+Environment=DOCKERFILEDIR=/usr/local/autotest/client/tests/docker/subtests/docker_cli/systemd
```
edsantiago: Same as above, only much more so. Make it very obvious that this is a garbage value, eg `DOCKERFILEDIR=this-will-be-replaced-by-the-test`. And try to make a habit of doing so for values that will be replaced.

The reason is that `/usr/local/autotest/blah/blah` looks "real". It looks like an actual value, and my first thought on reading it was hey, wait, no, it should be `/var/lib/autotest`, and besides there's too much duplication here, and I spent a few minutes figuring out that the value is thrown away.

When you have strings that are going to be replaced, make it obvious that they are dummy data. If you want them to look like an example, maybe use /this/is/an/example or some other convention. But don't use values that resemble real data, because they'll trick your maintenance programmer.

subtests/docker_cli/systemd/systemd.py
edsantiago: Agree with @cevich. One addition: `separator` is misnamed. And I think the function is too. And I'm not sure I like the multiple invocations and copies. I think there's got to be a cleaner way, I'm just not sure what that is. Maybe:

    def write_unitfile(self, env):
         ...open original (git-controlled) file for reading
         ...open output file (/etc/systemd) for writing
          ...while read:
               if line.startswith('Environment='):
                     extract the right-hand side, split on the '=' sign
                     if left-hand side in env dict:
                         line = 'Environment={}={}'.format(left-hand-side, env[right-hand-side])
                     write line to output file

config_defaults/subtests/docker_cli/systemd.ini
cevich: I forget, is the image used here is also referenced in ``FROM`` to build that ``4321`` test image?  IMHO, it would be good to get them all in sync, but that probably means updating the dockerfile's ``FROM`` line.  We do this in the ``build`` test.

hmmm

Yeah, let's replace them all with the CentOS base image.  That's (more) guaranteed to work forever & wherever.  I'm fine with having it just use ``image_name`` for the foreseeable future, though it would be a nice update to make it work with the conventional mechanisms.  It's more important that the systemd feature get tested than all the niggly details are right from the start.

afomm:
Hi guys, applied these feedbacks except editing the Dockerfile. Can you please review?

cevich:
Yep, looking now.

edsantiago:
Gaaaaah! Why did I get @cevich's email notification, but not @afomm's ???

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -1,8 +1,17 @@
 [docker_cli/systemd]
-docker_timeout = 300
+subsubtests = systemd_pull, systemd_build , systemd_run
```
cevich: My OCD briefly took over and reminded me to make a comment about  commas and whitespace and perfection.  There is is.

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -1,8 +1,17 @@
...
+
 #: Systemd unit file
+sysd_unit_file = docker-pull.service
+#: image name to pull, build, run
+image_name = centos:latest
+
+[docker_cli/systemd/systemd_build]
+sysd_unit_file = docker-build.service
+image_name = rhel7_p4321
```
cevich: Should this be ``centos_p4321``?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -5,58 +5,51 @@
...
 Operational Summary
 ---------------------
-This is to test container as systemd service using a predefined unit file
-(``p4321.service``) and a ``script p4321-server.py``. An image is built with
-this script added. Systemd starts a container using this image & the container
-writes current time to port ``4321``. From host, the socket is read and the
-value is checked to see if it is a digit. Finally, container is stopped
-through systemd and system is cleaned up.
+This is to test container as systemd service using a predefined unit file.
+Pulling, building, and running are performed.
```
cevich: Should probably add a ``Prerequisites``  section added here mentioning something about any configured ``image_name`` value and ``Dockerfile`` base image must already exist or be retrievable.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -65,22 +58,32 @@ def sysd_action(action=None, sysd_unit=None):
...
-                if time_from_socket:
-                    s.shutdown(socket.SHUT_RDWR)
-                    return time_from_socket
-                num_read_try -= 1
-            except socket.timeout:
-                pass
-            time.sleep(1.0)
-        raise DockerTestError("Failed to read from socket with %d tries"
-                              % max_read_tries)
+    def write_unitfile(self, var_to_edit):
```
cevich: Not sure ``var_to_edit`` describes what it is.  What about something like ``needle_haystack_map``?

edsantiago:
(answer: 3800-second email delays on sendgrid.net)

config_defaults/subtests/docker_cli/systemd.ini
edsantiago: The `rhel7` in the name is misleading

subtests/docker_cli/systemd/systemd.py
```diff
@@ -65,22 +58,32 @@ def sysd_action(action=None, sysd_unit=None):
...
+        with open(self.sub_stuff['unit_src'], "r") as srcfile:
+            lines = srcfile.readlines()
+        with open(self.sub_stuff['unit_dest'], "w") as tmpfile:
+            for line in lines:
+                for unitf_var_name, changeto in var_to_edit.items():
+                    if unitf_var_name in line:
+                        str_to_change = \
+                            line.partition(unitf_var_name)[2].strip()
+                        line = line.replace(str_to_change, changeto)
```
cevich: This is getting a bit too deeply nested, take a step back, consider a simpler way.  What strikes me right away is that the code is modifying the input values, why not just start out with the proper value?

Think, what does python have built-in that is able to take a string, and swap out specific character sequences with keywords, by values found by looking up that keyword in a dictionary?

Oh wait, look, you're already passing in a dictionary to the function anyway :grinning: 

Once you have figured this out, just set up your input (unit) file lines, so they include the tokens ("special" character sequences).  Then setup the values in your dictionary so the output is what you need.

Hint: You're already using something similar in ``sysd_action()``, just w/o the keyword part.

subtests/docker_cli/systemd/docker-build.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=docker-autotest systemd build image subtest. If you see this file on a live system it means something has failed very badly
+
+[Service]
+Type=oneshot
+Environment=NAME=set-by-the-test-from-systemd.ini
```
edsantiago: Why two different strings? I like `to-be-set-by-test` (below), it's short and descriptive. Why not use that for `NAME=` also? Imagine yourself as the maintenance programmer. You see this:

    NAME=to-be-set-by-test
    DOCKERFILEDIR=to-be-set-by-test

Your first thought is, aha, those are identical values in contexts where they should be different, that clearly means that I should look elsewhere, in one other place, to see how they're set. But if you see:

    NAME=one-thing
    DOCKERFILEDIR=another

...you might wonder: why are those different? What is the reason for using one replaceable value in one place, and another in the other?

subtests/docker_cli/systemd/systemd.py
edsantiago: Would you consider renaming to plural? `vars_to_edit`? Singular is misleading because there can be (and so far, always are) multiple vars.

subtests/docker_cli/systemd/systemd_run.py
```diff
@@ -0,0 +1,66 @@
...
+                                           self.config['image_name'],
+                                           self.sub_stuff['bindir']])
+        mustpass(dkrcmd.execute())
+
+    def run_once(self):
+        host = 'localhost'
+        port = 4321
+        super(systemd_run, self).sysd_action('daemon-reload')
+        super(systemd_run, self).sysd_action('start',
+                                             self.config['sysd_unit_file'])
```
cevich: No need for ``super()`` here, that's only necessary for "special" methods (like ``__init__()`` or ``run_once()``) or where you must have a guarantee of where in the class ancestry it's looked up (maybe it's called from a method with the same name). For regular-ole-methods, just pretend as if they're already defined in the class.  So, ``self.sysd_action(...)`` in this case.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -5,58 +5,51 @@
...
-        port = 4321
         self.sysd_action('daemon-reload')
-        self.sysd_action('start', self.config['sysd_unit_file'])
-        utils.wait_for(lambda: not is_port_free(port, host), 10)
-        time_from_socket = self.read_socket(host, port)
-        self.failif(not time_from_socket.isdigit(),
-                    "Data received from container is non-numeric: '%s'"
-                    % time_from_socket)
+        for act in 'start', 'status':
+            self.sysd_action(act, self.config['sysd_unit_file'])
```
edsantiago: I'm not seeing a `postprocess()`, and that makes me slightly uncomfortable. The systemd start & status (in `run_once`) seem to catch every error I can throw at them, but I can't help thinking that there should be some more error & status checking. I can't define it exactly so I'll keep pondering it today.

subtests/docker_cli/systemd/systemd.py
cevich: One more clue.  Lines w/o tokens will throw exceptions, which can be made useful like this:

```python
try:
     replacement_line = ...super secret sauce...
except KeyError:
    replacement_line = original_line
...
unitfile.write(replacement_line)
```
```

File: subtests/docker_cli/systemd/systemd.py [hunk 23]
```diff
@@ -65,22 +58,32 @@ def sysd_action(action=None, sysd_unit=None):
...
-                num_read_try -= 1
-            except socket.timeout:
-                pass
-            time.sleep(1.0)
-        raise DockerTestError("Failed to read from socket with %d tries"
-                              % max_read_tries)
+    def write_unitfile(self, var_to_edit):
+        with open(self.sub_stuff['unit_src'], "r") as srcfile:
+            lines = srcfile.readlines()
+        with open(self.sub_stuff['unit_dest'], "w") as tmpfile:
```
edsantiago: `tmpfile` here is dangerously misleading! You can do two things here:

  * Rename variable to `outfile`, or
  * go back to your (correct and proper) earlier way, in which you write a tempfile and then atomically rename it into place.

The second form is correct, because it is super-dangerous to write to non-tempfiles: if your program crashes, or the disk fills up, or there's any of a million different errors, you end up with a corrupt `/etc/systemd/system/foo.unit`.

subtests/docker_cli/systemd/systemd.py
edsantiago: Followup on postprocess: maybe running `DockerImages(self).list_imgs_full_name()`, stripping registry & tags (eg docker.io/centos:latest -> centos), and checking for `image_name`?

afomm:
Thanks for the great feedback. I'll apply & send a fixed version.

subtests/docker_cli/systemd/systemd.py
afomm: Thanks @cevich. I had reduced the unit files to something similar but then I reverted back thinking that it might not be as flexible. I'll integrate your feedback. Right now, my write_unitfile (below) treats the file lines as list - thanks to your previous suggestion. I'll send my update to PR.
    def write_unitfile(self, needle_haystack_map):
        with open(self.sub_stuff['unit_src'], 'r') as srcfile:
            lines = srcfile.readlines()
            for subkey in needle_haystack_map.keys():
                needle_found = next(hays for hays in lines if subkey in hays))
                indx = lines.index(needle_found)
                lines[indx] = '{}{}\n'.format(subkey,needle_haystack_map[subkey])
        with open(self.sub_stuff['unit_dest'], 'w') as outfile:
            outfile.writelines(lines)

afomm:
@cevich @edsantiago - I need to add postprocess as suggested by Ed. Otherwise, this should have all the feedback. Please review.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -5,82 +5,124 @@
...
+                prep = ''
+            # edit lines with matching string to dict keys
+            for key in vars_dict.keys():
+                ndl_found = next((ndl for ndl in lines if key in ndl), None)
+                if ndl_found is not None:
+                    indx = lines.index(ndl_found)
+                    lines[indx] = '{}{}{}\n'.format(prep, key, vars_dict[key])
+        with open(tmp_write, 'w') as outfile:
+            outfile.writelines(lines)
+        copyfile(tmp_write, tmpdirf)
```
cevich: This still seems slow and clunky for what amounts to a search/replace of a single key/value.  I think I suggested before, there's no reason to iterate line-by-line. It's going to be much simpler (and faster) to read the entire file into a single string variable (it will preserve the newlines).  Then just do one search/replace across the entire thing all at once.  Then write the entire string out.

Also, I'm not sure why you've got the double-indirection of using an environment variable ``NAME``, and then referencing that from the other values.  If you're manipulating the file with code anyway, it's easier/faster to just search down all the tokens and replace them with duplicate values.  That's why I suggested just sticking the format tokens directly into your template unit files.

Unless I'm misunderstanding something, and you have to use the ``NAME`` environment for some reason.

cevich: [Here's my old comment about this](https://github.com/autotest/autotest-docker/pull/653#discussion_r90657604).  Basically you can boil this entire method down to five lines:

```python

    def write_unitfile(self, substitutions):
        srcfile = open(...)
        destfile = open(...)
        with open(self.sub_stuff['unit_src'], "r") as srcfile,
             open(self.sub_stuff['unit_dest'], "w") as destfile:
               destfile.write(srcfile.read().format(**substitutions))
```

Then, with a properly prepared dictionary, call it with ``self.write_unitfile(self.sub_stuff)``

subtests/docker_cli/systemd/systemd_run.py
```diff
@@ -0,0 +1,77 @@
...
+            s.settimeout(2.0)
+            try:
+                time_from_socket = s.recv(512).strip()
+                if time_from_socket:
+                    s.shutdown(socket.SHUT_RDWR)
+                    return time_from_socket
+                num_read_try -= 1
+            except socket.timeout:
+                pass
+            time.sleep(1.0)
```
cevich: In all cases, it's "nice" to shutdown the socket.  Maybe:

```python
            try:
                ...
            except:
                ...
            finally:
                time.sleep(1.0)
                s.shutdown(socket.SHUT_RDWR)
```

Also, I think pylint may get mad about using single-character variable names.  IIRC, three-characters is the minimum.

subtests/docker_cli/systemd/systemd.py
cevich: Summary: Ed's right, but checking if the image is there is common to all sub-subtests, so you can add that to the base class's ``postprocess()``.  For docker run, just move the "checking" lines from ``run_once()`` to ``postprocess()`` and add a call to super (to check the image is there).  For build and pull, they can just inherit and use the base class method (checking for the pulled or built image).

subtests/docker_cli/systemd/systemd.py
edsantiago: Full agreement with @cevich : this has grown waaaaaaaay too complex for what it does. And in addition:

  * The moment you started using this for rewriting Dockerfile, you should've renamed the function: Dockerfile isn't a unitfile, so the function name is now highly misleading.
  * The `copyfile` at the end is super dangerous: it leaves the tmp file behind. You don't want that. You don't want to leave any trace. That should be `os.rename()`.
  * the 'Environment=' was fine when this was write_unitfile but now it's dangerous: since this is no longer a function dedicated to unit files, this function should not know about Environment=. That should be left to the caller now.

One possible way to do this (completely untested, it's late and I'm tired):

    def sed_file_in_place(path, replacements):
        tmpfile = path + '.tmp'
        os.unlink(tmpfile)           # in case there's a dropping from a previous run
        s = open(path).read()
        for key, newvalue in replacements:
            searchfor = re.compile(key + '.*$', re.MULTILINE)
            s = re.sub(searchfor, key + newvalue, s)
        ...writing to tmpfile, then os.rename(), left as exercise for the reader

subtests/docker_cli/systemd/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:latest
+FROM rhel:latest
```
afomm: Well, I guess it has to be something like {from_img} since it is edited during test run

subtests/docker_cli/systemd/systemd_run.py
afomm: @cevich when I do that, I am getting "Failed to read from socket with %d tries" error frequently. So how about:
```
            finally:
                time.sleep(1.0)
            skt.shutdown(socket.SHUT_RDWR)

**Comment by afomm** [comment]
@edsantiago @cevich - it is ready for your review again. Thanks!

File: subtests/docker_cli/systemd/systemd_run.py (hunk 25 above)
Comment (cevich) [inline subtests/docker_cli/systemd/systemd_run.py]: Nope, the point is, exception or not, the code in the ``finally`` block always runs.  But I don't doubt you're getting an error.  Hmmmm.  Maybe:

```python
finally:
    time.sleep(...)
    if skt:
        socket.shutdown(...)
```

may need to tweak the ``if`` condition, not sure.  Check the python docs, there's probably a "in a bad state" call or attribute you can use.  If not, it's probably fine to totally ignore the finally exception like this:

```
finally:
    time.sleep(...)
    try:
        skt.shutdown(...)
    except:
        pass
```

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
-                    "Data received from container is non-numeric: '%s'"
-                    % time_from_socket)
+        for act in 'start', 'status':
+            self.sysd_action(act, self.config['sysd_unit_file'])
+
+    def postprocess(self):
+        super(systemd_base, self).postprocess()
+        img_lst = ''
+        for img in DockerImages(self).list_imgs_full_name():
+            img_lst = '{} {}'.format(img_lst, img.split('/', -1)[-1])
```
edsantiago: Augh. That's really awkward. What about:

    img_list = [x.split('/', -1)[-1] for x in DockerImages(self).list_imgs_full_name()]

(untested)

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
+
+
+class systemd_build(systemd_base):
+
+    def initialize(self):
+        # update dict, edit unitfile & copy it to /etc/systemd/system/
+        super(systemd_build, self).initialize()
+        build_dict = {'dockerfiledir': self.sub_stuff['bindir'],
+                      'options': self.config['unit_opts']}
+        vars_dict = self.sub_stuff['vars_dict']
```
edsantiago: Where is this sub_stuff set?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
+        self.sed_file(vars_dict)
+        copyfile(self.sub_stuff['tmpdirf'], self.sub_stuff['unitf_dst'])
+        # copy Dockerfile to tmpdir, edit, then copy to bindir for build
+        self.sub_stuff['tmpdirf'] = join(self.sub_stuff['tmpdir'],
+                                         'Dockerfile')
+        tmpdirf = self.sub_stuff['tmpdirf']
+        self.sub_stuff['dkrfl_bk'] = '{}.bk'.format(tmpdirf)
+        for f in [tmpdirf, self.sub_stuff['dkrfl_bk']]:
+            copyfile(self.sub_stuff['dkrfl'], f)
+        vars_dict = {'img_from': self.config['img_frm']}
```
edsantiago: Whoa. I'd like you to take a close look at this, possibly in conjunction with the subsequent sed_file. Please tell me what it's doing (what it's really doing, not what you think now that it's doing).

I'm sorry for not pointing it out directly, but this is a major bug and I think it's important for you to track it down on your own.

edsantiago:
obtw @afomm instead of doing `git commit --amend && git push --force`, you can do `git commit -m"fixup! blah blah blah" && git push`. The "fixup!" tells git (and the person doing the merge) that all these commits can be squashed at merge time. Doing this helps your reviewer identify what has changed between one review check and the next. I think it might also actually send email notification, unlike push --force.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
+            self.sysd_action(act, self.config['sysd_unit_file'])
+
+    def postprocess(self):
+        super(systemd_base, self).postprocess()
+        img_lst = ''
+        for img in DockerImages(self).list_imgs_full_name():
+            img_lst = '{} {}'.format(img_lst, img.split('/', -1)[-1])
+        if self.config['image_name'] not in img_lst:
+            raise DockerTestError("Image %s is not present"
+                                  % self.config['image_name'])
```
cevich: Could be written smaller/simpler, but this is OK for now I think.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
 
     def cleanup(self):
-        super(systemd, self).cleanup()
-        filepath = self.config['sysd_unitf_dest']
+        super(systemd_base, self).cleanup()
+        filepath = self.sub_stuff['unitf_dst']
         if exists(filepath):
             self.sysd_action('stop', self.config['sysd_unit_file'])
```
cevich: If ``super(...)`` or ``sysd_action`` fails for any reason.  The unit file won't be deleted.  I think you need a ``try:...finally:...`` here to be extra sure the file gets removed and daemon is reloaded.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
+# To test building an image using systemd unitfile
+#
+# Operational Summary
+# ---------------------
+#
+# #. Edit unitfile ``docker-build.service`` & copy it to /etc/systemd/system
+# #. Edit Dockerfile
+# #. Systemd service is started using this file to build an image.
+# #. Systemd service is stoped and image removed
+# """
```
cevich: Kinda strange to have a docstring inside comments.  Maybe just stick it in as the docstring for ``class systemd_build()``?

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,155 @@
...
+#
+# To test pulling an image using systemd unitfile
+#
+# Operational Summary
+# ---------------------
+#
+# #. Edit unitfile ``docker-pull.service`` & copy it to /etc/systemd/system
+# #. Systemd service is started using this file to pull an image.
+# #. Systemd service is stoped and image removed
+# """
```
cevich: Same here.  Just stick it into the class.

subtests/docker_cli/systemd/systemd.py
afomm: @edsantiago super has:
```self.sub_stuff['vars_dict'] = {'name': self.config['image_name']}```
But I think I should have the classes set their own instead?

edsantiago: Yes, I forgot to check the parent. Thanks. 

I think you spoke with Chris earlier, and I'm mostly in agreement with his opinion.

afomm:
@edsantiago @cevich - I added most of the suggestions but still looking into the bug Ed identified (I've a few ideas but not sure yet). By the way, I used "!fixup..." - did you get any notification? Please review.

edsantiago:
Yes! I received an email notifying me of your fixup commit. Looking now.

cevich:
Closing in favor of #673

PR #654, opened 2016-10-26
Remove failing --tmpfs tests

edsantiago:
PR 608 added three tests for invalid use of --tmpfs. All three
turned out to be fragile: docker-1.12 fails in a different
manner than 1.10. After an hour's effort I'm unable to come up
with any equivalent tests that fail similarly in 1.10 & 1.12,
so it seems best to just remove them entirely. There is still
an actual test for --tmpfs functionality.

### DIFF
--- a/config_defaults/subtests/docker_cli/negativeusage.ini
+++ b/config_defaults/subtests/docker_cli/negativeusage.ini
@@ -1,5 +1,5 @@
 [docker_cli/negativeusage]
-subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, iv5, iv6, iv7, ip1
+subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, ip1
 #: Optional docker subcommand (string). Automatically generated substitution keys:{n}
 #: {n}
 #: {t} * ``%%(FQIN)s`` **-** valid, default test image name{n}
@@ -69,24 +69,6 @@ subcmd = run
 subarg = -e,PATH=/tmp,%%(FQIN)s,true
 stderr = exec\: \"true\"\: executable file not found in \$PATH
 extcmd = 127
-
-[docker_cli/negativeusage/iv5]
-subcmd = run
-subarg = --tmpfs,/dev,%%(FQIN)s
-stderr =  no such file or directory
-extcmd = 127
-
-[docker_cli/negativeusage/iv6]
-subcmd = run
-subarg = --tmpfs,:rw,%%(FQIN)s
-stderr =  open /proc/self/task/1/attr/exec: no such file or directory
-extcmd = 125
-
-[docker_cli/negativeusage/iv7]
-subcmd = run
-subarg = --tmpfs,.,%%(FQIN)s
-stderr =  Message: Failed to open /dev/null - open /dev/null: permission denied
-extcmd = 125
 
 [docker_cli/negativeusage/ip1]
 subcmd = tag


afomm:
Hello Ed,

The return values might have changed. I'll check it out.

edsantiago:
Exit status change, and error message change. Like, completely different results. Word of warning, I think it will take hours of effort to come up with tests that will work for both 1.10 and 1.12, and I think it's very low ROI.

cevich:
Thanks for taking a look at this.  I'm sure this one is ripe for exactly the problem you describe (now and in the future).  I'm all for killing tests that are not useful.  Additionally, I think the core-problem here isn't going to go away (given nature of the test).

Instead of treating the sub-subtests as one-size-fits-all, how about adding some version-comparison / limiting config. option(s)?

Tthat would enable us to have different sets of negative tests run depending on the docker version.  Since all the sub-subtests here are generated from config,  the subtest could simply not generate the not-applicable ones.  Though having them toss TestNA would be fine too.

What do you think?  I could make a trello card for this idea, it doesn't have to be jammed into this PR.

config_defaults/subtests/docker_cli/negativeusage.ini
```diff
@@ -70,24 +70,6 @@ subarg = -e,PATH=/tmp,%%(FQIN)s,true
...
-subcmd = run
-subarg = --tmpfs,:rw,%%(FQIN)s
-stderr =  open /proc/self/task/1/attr/exec: no such file or directory
-extcmd = 125
-
-[docker_cli/negativeusage/iv7]
-subcmd = run
-subarg = --tmpfs,.,%%(FQIN)s
-stderr =  Message: Failed to open /dev/null - open /dev/null: permission denied
-extcmd = 125
```
cevich: I'm a little hesitant to nuke this one.  Numerous bugs have been found/fixed by feeding null into places where it's unexpected.  IIRC there was a BZ filed on this @afomm do you remember?

config_defaults/subtests/docker_cli/negativeusage.ini
```diff
@@ -1,5 +1,5 @@
 [docker_cli/negativeusage]
-subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, iv5, iv6, iv7, ip1
+subsubtests = op1, op2, op3, if1, ov1, iv1, iv3, iv4, ip1
```
cevich: Some of these are referenced in ``subtests.ini`` generated from template in ADEPT.  I'll fix that to allow CI to actually run this test w/o barfing.

cevich: Hrmmm, oops, totally wrong.  There's nothing special done for this test in CI.  Wonder why it failed on mismatch w/ the ``subsubtests`` list...

edsantiago:
>Instead of treating the sub-subtests as one-size-fits-all, how about adding some version-comparison / limiting config. option(s)?

I'm not sure. But to take a step back, I think we should define the purpose of negativetest. We seem to be using it for two purposes:

* Testing well-defined and documented responses to invalid input
* Testing GIGO, as in just throwing random crap at docker and seeing how it responds

The tmpfs tests in this PR seem to be a case of the latter. They are definitely invalid commands, it's just not clear how docker should fail. The fact that docker exit status changes between 126 and 127 (command cannot be invoked vs command cannot be found) to me implies that these are undefined corner cases whose explicit behavior we shouldn't count on.

If we want to keep the tests, could we just add them to the tmpfs test, instead of negativeusage? Just add `mustfail(docker run with-invalid-arguments)`, without checking exact error status or message?

cevich:
>    Testing well-defined and documented responses to invalid input
>    Testing GIGO, as in just throwing random crap at docker and seeing how it responds 

Your analysis is correct, though I would also add:

*  Confirm Docker handles unexpected [negative] situations appropriatly - no tracebacks, kernel or go panic messages.

> If we want to keep the tests, could we just add them to the tmpfs test

What happened in the past was, mixing positive / negative tests together lead to really complicated code and confusing configuration.  Mostly because of how config. options are inherited, so any negative sub-subtests had to ignore a lot of stuff.  Also, there was a great amount of code duplication across all these separate negative tests, it made more sense to stick all of them into a purpose-built subtest.

So I don't think going back to the old way is a good idea.  It's slightly more work to add version-comparison options, but it will make dealing with the GIGO variety much easier - b/c the output/exit codes change so frequently and inconsistently across versions.  What I'm suggesting is, we merge this PR in, and open up a card to fix the fundamental problem so at least the null-option sub-subtest can be re-added.

The other thing that's probably become more of a necessary addition is, monitoring syslog / dmesg.  It's possible that one of the daemons freaks out, rather than the client.  That's also a problem worth catching.

Anyway, your changes here exposed a oddball bug I never considered, pushing to a PR that will allow this PR to properly run through CI...

cevich:
... #655 should be applied first, then we can ask jenkins to re-test this one.

afomm:
----- Original Message -----
> Thanks for taking a look at this.  I'm sure this one is ripe for exactly the
> problem you describe (now and in the future).  I'm all for killing tests
> that are not useful.  Additionally, I think the core-problem here isn't
> going to go away (given nature of the test).
> 
> Instead of treating the sub-subtests as one-size-fits-all, how about adding
> some version-comparison / limiting config. option(s)?
> 
I was thinking along the same lines but not sure how much work is needed to accomplish that.

- Afom

cevich:
[(?) [CI] Version-dependent testing](https://trello.com/c/tdAhquxo/316-ci-version-dependent-testing)

cevich:
> not sure how much work is needed to accomplish that.

Let's have that conversation over in trello...

config_defaults/subtests/docker_cli/negativeusage.ini
cevich: Gah, *squeek*, *squeek* ahh ha!  It wasn't a speck on my glasses!  Really should clean them more often :grinning: 

```Timestamp: 2016-10-27 19:10:20.065693229 +0000 UTC
Code: System error

Message: Failed to open /dev/null - open /dev/null: permission denied

Frames:
---
0: setupRootfs
Package: github.com/opencontainers/runc/libcontainer
File: rootfs_linux.go@81
---
1: Init
Package: github.com/opencontainers/runc/libcontainer.(*linuxStandardInit)
File: standard_init_linux.go@59
---
2: StartInitialization
Package: github.com/opencontainers/runc/libcontainer.(*LinuxFactory)
File: factory_linux.go@247
---
3: initializer
Package: github.com/docker/docker/daemon/execdriver/native
File: init.go@35
---
4: Init
Package: github.com/docker/docker/pkg/reexec
File: reexec.go@26
---
5: main
Package: main
File: docker.go@20
---
6: main
Package: runtime
File: proc.go@188
---
7: goexit
Package: runtime
File: asm_amd64.s@1998
docker: Error response from daemon: Cannot start container a2fdec625689f1646c4da64a1f647d2b2065ea31ea4cf86a9abb376337491a6a: [9] System error: invalid character 'r' looking for beginning of value.
```

Right, I think Afom showed that to me before.  Yuk.  I half-wonder if we shouldn't file a BZ on that.  I'd much prefer if docker just threw a simple error like: "tmpfs paths must be absolute".  Then if so, that begs the question of, shouldn't that format of output be caught by ``OutputGood``?

Sorry, lots of pondering, few answers.

cevich:
Opened https://bugzilla.redhat.com/show_bug.cgi?id=1389545 for the relative-path backtrace.

edsantiago:
I'm a bit lost as to where this stands. My understanding is that as of the time of this comment, the goal is to merge #655 then test this under Jenkins. But is there anything else for me to do? Anything I've missed in the discussion?

cevich:
Doh! @afomm already opened a similar bug on iv7, totally forgot!  
https://bugzilla.redhat.com/show_bug.cgi?id=1344132

Hmm, okay, I guess I'll link the two and let engineering sort them out.  Apparently it's a problem in Fedora as well.

config_defaults/subtests/docker_cli/negativeusage.ini
cevich: @edsantiago 

Okay, with two bugs open now on ``iv7``, and #655 merged, let's keep the iv7 test - though it will need to be updated once the bug(s) are fixed.  The others (``iv5`` and ``iv6``) can go.  To be considered for adding back in when/if https://trello.com/c/tdAhquxo is completed.  Sound reasonable?

cevich:
> I'm a bit lost as to where this stands.

Ya, working right now to update the commit IDs in CI for this job...

config_defaults/subtests/docker_cli/negativeusage.ini
edsantiago: Unfortunately, the test still fails. docker-1.10 exits with 125, 1.12 with 126.

I would really like to get rid of all failing tests. I really want to get ADEPT running cleanly, so when it fails I can actually look at the failure instead of shrugging and saying "oh, it always fails". If and when there is a defined behavior for `--tmpfs .`, I'd like to add a test for it. But right now I really just want to get tests passing.

cevich: The ``iv7`` test should be extra skipped, everywhere, it's flagged in **two** bugzillas now :grinning:

Maybe ``yum install -y --enablerepo=extras python-bugzilla`` ?

cevich:
...okay, updated all the jobs to use the latest HEAD from this repo.

cevich:
jenkins, test this please.

config_defaults/subtests/docker_cli/negativeusage.ini
edsantiago: >The iv7 test should be extra skipped

Only if it's called iv7. I tried reenabling it, but called it `iv5` because the gap offends me and will confuse future maintenance programmers. That resulted in a total barf: `DockerTestError: Sub-subtest(s) set(['docker_cli/negativeusage/iv7']) referenced in control include, exclude, and/or subthings but not defined in subsubtests configuration option: ['op1', 'op2', 'op3', 'if1', 'ov1', 'iv1', 'iv3', 'iv4', 'iv5', 'ip1']` (yes, I rebased on to master with pr655). So I disabled the bugzilla check because I wanted to actually test the code.

Bottom line: I'm not comfortable with band-aiding by disabling the test, because once (if) that BZ gets closed, we will still have a situation in which docker 1.10 exits 125, 1.12 126, and docker-FUTURE exits with TBD. And tests will break, and we'll have to scramble to find a solution that works on all versions of docker.

I totally agree that a test would be nice. I want to write that test *on our terms*, on our time, not as an emergency dictated by failing tests whenever that BZ happens to get closed. Does that make sense?

cevich:
there it goes

cevich:
Drat, something's still broken (or not fixed).

```
DockerTestError: Sub-subtest(s) set(['docker_cli/negativeusage/iv7']) referenced in control include, exclude, and/or subthings but not defined in subsubtests configuration option: ['op1', 'op2', 'op3', 'if1', 'ov1', 'iv1', 'iv3', 'iv4', 'ip1']
```

edsantiago:
Yeah, see my comment from 30 minutes ago. I'm concerned that autotest can be DoS'ed by mistakes or typos or just plain invalid data in a BZ.

cevich:
> invalid data in a BZ.

We control the data in BZ.  If it's wrong, then we fix it, including clearing the whiteboard if necessary.  We can also change the field used for internal jobs, e.g. use the qe-whiteboard (internal-only).  But those are also only band-aids.  The trello-card I opened can address the longer-term problems.  I agree, the urgent problem is to now allow random BZ data to lead to test breakage.

Let me check the job that just ran/failed and make sure it's pulling in the right commits.  If so, I'll set up a test job in the development environment that only runs ``negativeusage`` so we can iterate on a fix faster...

cevich:
...OIC, the CI job is using the commit from this PR (as it's suppose to) - so it doesn't include the fix I just merged.  Can you rebase / re-push this so we can see if the critical bug is fixed?

cevich:
Okay, lessee what CI says about those.  I made https://url.corp.redhat.com/398f0d5 so it only runs the negativeusage test as well.

cevich:
@edsantiago looks like it's working now.  No more error on missing sub-subtest.

cevich:
FWIW: What I did was ssh into the node running the test, and ``tail -f /var/lib/autotest/client/results/default/debug/client.INFO``.  Saw the ``negativeusage`` test fly by and there was no more error about missing config section for ``docker_cli/negativeusage/iv7``.  Double-checked ``control.ini`` and ``client.DEBUG`` and made sure it was searching & marking the nonexistent  sub-subtest for exclusion (it was).

edsantiago:
Weird. I'm still seeing a test crash on my setup.

cevich:
> Weird. I'm still seeing a test crash on my setup.

Aye, that's why I rely on the dockerautotest CI job.  The "setup" is exactly the same all the time.  It also has the side-effect of being nearly the exact same setup used for production testing :grinning: Soon-ish, I want ADEPT to be able to do all this locally, no janky-jenkins in the mix.

PR #655, opened 2016-10-27
Fix bzexclude

cevich:
All sub-subtests defined in a subtest's ``subsubtests`` config.
option must have corresponding config. sections.  However, if a
sub-subtest is excluded by an open bugzilla, the code assumes
it's configuration section must exist.  This is not always the
case, esp. if the excluded sub-subtest was removed from config.

Introduce a new property that grants access to only the exclusions
returned by bugzilla (if any).  During initialization of either
"caller" subtest, refrain from including these bugzilla excluded
items in the subsubtest_names attribute.

### DIFF
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -11,7 +11,6 @@ loading the specified configuration section (see `configuration module`_)
 # Pylint runs from a different directory, it's fine to import this way
 # pylint: disable=W0403
 
-import logging
 import tempfile
 import os.path
 import imp
@@ -246,6 +245,8 @@ class SubSubtest(subtestbase.SubBase):
         """
         Form subsubtest configuration by inheriting parent subtest config
         """
+        # Many branches needed to keep this method compact
+        # pylint: disable=R0912
         subsubtest_config = all_configs.get(name, {})
         # don't redefine the module
         _config = copy.deepcopy(parent_config)  # a copy
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -211,7 +211,6 @@ class SubSubtest(subtestbase.SubBase):
         self.config_section = self.make_name(pscs)
         # Allow child to inherit and override parent config
         all_configs = config.Config()
-        # make_subsubtest_config will modify this
         parent_config = self.parent_subtest.config
         # subsubtest config is optional, overrides parent.
         if self.config_section not in all_configs:
@@ -281,16 +280,6 @@ class SubSubtest(subtestbase.SubBase):
             else:
                 _config[key] = val
         return _config
-
-    def make_subsubtest_config(self, all_configs, parent_config,
-                               subsubtest_config):
-        """
-        Deprecated, use make_config() instead, will be removed soon
-        """
-        del subsubtest_config  # not used
-        logging.warning("SubSubtest.make_subsubtest_config() is deprecated!")
-        self.config = self.make_config(all_configs, parent_config,
-                                       self.config_section)
 
 
 class SubSubtestCaller(Subtest):
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -153,26 +153,47 @@ class Subtest(subtestbase.SubBase, test.test):
         """
         self.log_step_msg('postprocess_iteration')
 
+    def _control_ini_section(self, section):
+        if self._control_ini is None:
+            self._control_ini = {}  # empty set of caches
+        cache = self._control_ini.get(section, None)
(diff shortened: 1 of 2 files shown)


cevich:
travis? where are ya buddy?

dockertest/subtest.py
```diff
@@ -154,27 +153,37 @@ def postprocess_iteration(self):
...
             fullpath = os.path.join(self.job.resultdir, 'control.ini')
             # Control-file & behavior cannot be assumed, file may not exist.
             try:
-                self._control_ini = config.ConfigDict('Control')
+                self._control_ini = config.ConfigDict(section)
                 self._control_ini.read(open(fullpath, 'rb'))
             except (IOError, OSError, Error):
                 self.logwarning("Failed to load reference '%s' and/or"
-                                "it's '[Control]' section.", fullpath)
+                                "it's '[%s]' section.", fullpath, section)
```
edsantiago: As long as you're touching this line, could you s/it's/ its/? (remove apostrophe, add leading space to separate from preceding and/or) -- thanks

dockertest/subtest.py
```diff
@@ -377,7 +381,8 @@ def initialize(self):
             if len(children) < 1:
                 return  # No children specified
             # Any children NOT listed in fullnames are 'undefined'
-            if children - fullnames:
+            # unless they were excluded by (an old) bugzilla
+            if children - fullnames - bzexclude:
                 msg = ("Sub-subtest(s) %s referenced in control include, "
                        "exclude, and/or subthings but not defined in "
                        "subsubtests configuration option: %s"
```
edsantiago: The line right below this, shouldn't it be `% (children i fullnames - bzexclude` ? And if so, would it make sense to refactor that long list into something like

    foo = children - fullnames - bzexclude
    if foo:

dockertest/subtest.py
cevich: done

dockertest/subtest.py
cevich: Oops, yep, you're right again.

dockertest/subtest.py
```diff
@@ -154,27 +153,37 @@ def postprocess_iteration(self):
...
-        """
-        Represent operational control.ini's [Control] section as a dict or None
-        """
+    def _control_ini_section(self, section):
         if self._control_ini is None:
             fullpath = os.path.join(self.job.resultdir, 'control.ini')
             # Control-file & behavior cannot be assumed, file may not exist.
             try:
-                self._control_ini = config.ConfigDict('Control')
+                self._control_ini = config.ConfigDict(section)
```
edsantiago: Oh dear, one more: shouldn't this be `self._control_ini[section]` ? With similar changes above and below?

cevich:
Thanks @edsantiago .  Looks like the CI run caught a bug that's in the ``control`` as well.  Force-pushed fixes.

cevich:
> today's not looking like the day I'll grok it

May that day never come :grinning: 

This method, in combination with a few others solve these problems:

* Autotest client ``--args`` options should override ``control.ini`` options
* The bugzilla ``control.ini`` options are optional (bugzilla module missing or disabled)
* Autotest treats the ``control`` file itself as an content, not code (directly), so this applies also to ``control.ini`` - it may not exist or have different content than expected.
* The control file must be self-contained, it can't rely on any test-code.
* The test code must not rely on specific behavior or include code from the control file.
* The control file can't do final processing of sub-subtests as they can't be loaded (they depend on subtest code/logic).
* Some sub-subtests are generated dynamicly, but also must respect all of the above.

There are probably more as well.  It's a nightmare and took a REALLY long time to get it behaving well.

dockertest/subtest.py
cevich: Mmmmm, I don't think so, but maybe I'm misunderstanding your concern.

edsantiago: foo = self.control_config
    bz = self.bugzilla_config
    ! in the second invocation of _control_ini_section(), self._control_ini has been set (by the first invocation) to config.ConfigDict('control')

cevich: Oh snaps, your right, I totally ignored ``self`` dot.  Thanks!

edsantiago:
Tentative LGTM. Keep in mind I only reviewed changes, without fully understanding context. I have not tested heavily, just a few simple autotest runs.

I assume you'll fix the redefinition of `cache` issue.

cevich:
> I assume you'll fix the redefinition of cache issue.

Ya, I stripped out a pylint disable that was there, now I know why it was there :frowning:

PR #658, opened 2016-11-04
New subtest: run_flags - check daemon argv

edsantiago:
Make sure docker daemon is running with the
expected command-line options.


edsantiago:
https://trello.com/c/DRzLe51S/225-8-docker-devicemapper-enable-daemon-storage-options-dm-use-deferred-removal-and-dm-use-deferred-deletion-by-default

subtests/docker_cli/run_flags/run_flags.py
```diff
@@ -0,0 +1,94 @@
...
+
+        raise DockerTestFail("missing option: dm.use_deferred_deletion")
+
+    @staticmethod
+    def _is_rhel_or_centos():
+        """ True if this system is RHEL or CentOS """
+        try:
+            with open('/etc/redhat-release', 'r') as rh_release_fh:
+                rh_release = rh_release_fh.read()
+            if 'Red Hat Enterprise Linux' in rh_release:
```
cevich: This will match on Atomic Host as well.  Good.

subtests/docker_cli/run_flags/run_flags.py
```diff
@@ -0,0 +1,94 @@
...
+            return False
+        return False
+
+    def _check_kernel_support(self):
+        """
+        Fail with N/A if this is a RHEL/CentOS host with too-old kernel.
+        """
+        if not self._is_rhel_or_centos():
+            return
+        required = LooseVersion("3.999")    # FIXME once we ID a good version
```
cevich: FIXME comment suggests maybe this should be a configuration option?

Not sure the ``_is_rhel_or_centos()`` is necessary, kernel version and docker version should be good enough IMHO.  
That allows this to run on Fedora to potentially detect problems earlier.  It's easy for ADEPT to swap in another/different config. option value depending on the distro.

subtests/docker_cli/run_flags/run_flags.py
```diff
@@ -0,0 +1,94 @@
...
+
+Operational Summary
+-------------------
+Get the full command line of the currently-running docker or docker-latest
+daemon. Check for presence of expected option flags.
+
+"""
+
+import os
+from distutils.version import LooseVersion   # pylint: disable=E0611,F0401
```
cevich: Hmmm, guess I'm wrong.

BTW, you can use ``git commit --fixup <ID> /path/to/file`` and it'll do the right thing so ``git rebase -i --autosquash`` works later.

dockertest/docker_daemon.py
```diff
@@ -152,6 +152,27 @@ def restart():
     _systemd_action('restart')
 
 
+def pid():
+    """ returns the process ID of currently-running docker daemon """
+    result = _systemd_action('show --property=MainPID').stdout
+    if not result.startswith('MainPID='):
+        raise RuntimeError("Unexpected result from systemctl: '%s'" % result)
+    return int(result[8:])
```
edsantiago: Fixed. That had been my first wish, I just didn't want to deal with line breaks. Thanks for the incentive to do the right thing.

dockertest/docker_daemon.py
```diff
@@ -152,6 +152,27 @@ def restart():
...
+    """
+    Returns the command line (argv) for the currently-running docker daemon,
+    as read from /proc/<pid>/cmdline. We don't use 'systemctl show' because
+    that includes unexpanded variables. Return value is a list of strings.
+    """
+    cmdline_file = os.path.join('/proc', str(pid()), 'cmdline')
+    cmdline_args = []
+    with open(cmdline_file, 'r') as cmdline_fh:
+        cmdline_args += cmdline_fh.read().split('\0')
+    return cmdline_args
```
edsantiago: Fixed, thanks

subtests/docker_cli/run_flags/run_flags.py
edsantiago: `is_rhel_or_centos`: removed. Thanks, much cleaner now. I can assume that CentOS will track RHEL, and that Fedora will have much more recent kernels.

Config setting: I'm moderately opposed to that, because I don't see it being configurable. It's a value we'll discover some day. Once we do we'll make a new commit here (have to do that whether it's a config value or in code), and forget about it forever. It should never need to be touched again, which means it shouldn't offer the illusion of configurability. But there are still uncertainties about this, so if you feel strongly about making it configurable I'd like to hear your thoughts.

cevich:
> so if you feel strongly about making it configurable I'd like to hear your thoughts.

I was thinking in terms of the bigger-picture set of CI jobs.  AFAIK, the RHEL Kernel version != Fedora kernel version (and maybe CentOS too?).  Unless, WRT this bug they're essentially the same, in that case what you have is fine (fix it later if we need to).

edsantiago:
What I meant was: whatever RHEL kernel version we end up with here, Fedora will already be multiple versions ahead (so the greater-than test will automatically pass). CentOS, I'm assuming will track RHEL, but I'd like to worry about when we get there.

PR #659, opened 2016-11-08
Prefetch centos, to fix failing build test

edsantiago:
build/https_file and build/rm_false tests are failing
because their Dockerfiles pull in centos, so img_count()
finds two new post-run images instead of one.

This patch solves it, but not ideally. It's duplication on top
of more duplication. A better solution might piggyback off
preserve_fqins but that too is fragile. Submitting as a
compromise to get tests passing.


cevich:
Hmmm, this might actually work.

cevich:
> Submitting as a compromise to get tests passing.

This only affects test runs on RHEL / Fedora + Atomic ya? Maybe a better solution would be to change the base-image used in the dockerfile to something uncommon (like busybox) then update all the image-count checks to "2".

Another idea is, add custom configurations for these tests in ADEPT's ``subtests.ini.j2`` template (which is intended for exactly this purpose).

cevich:
Re: CI run results- the ``build/git_path`` fail is on me, still working on a fix for it (and some other related things).

edsantiago:
> This only affects test runs on RHEL / Fedora + Atomic ya? 

I don't know. I haven't delved deeply enough. I see it on the nightly ADEPT run on RHEL. It probably wouldn't fail on CentOS because the centos image is already pulled.

> change the base-image used in the dockerfile to something uncommon (like busybox) then update all the image-count checks to "2".

That should work but it's even worse duplication: all of postproc_cmd_csv (very long string) needs to be duplicated, and worse, in two places.

> add custom configurations for these tests in ADEPT's subtests.ini.j2 

Possibly, but see above for potential duplication. I'll let you decide that, because I'm not ready to dive into ADEPT hacking yet.

cevich:
>even worse duplication

Good thing you never saw this test _before_ the command-pattern was deployed.  These were ALL basically hard-coded in python instead of config values :frowning_face: 

> Possibly, but see above for potential duplication.

What if we just add this to ``[docker_cli/build]``:

```ini
docker_repo_name = centos
docker_repo_tag = 7
docker_registry_host = docker.io
docker_registry_user = stackbrew
```

Hmmm, meh, not much/any better that your solution.  I think I like your fix better.

Only thing it's possibly missing is a comment about why the centos value is there.  Something like "helps some build sub-subtests image counting"

cevich:
LGTM, let's get this merged.

cevich:
eek! did you not squash?

cevich:
whew! okay, you did, cool.

edsantiago:
Huh? Yes, squash-and-merge. Looks squashed on my pull. What are you seeing?

cevich:
BTW: I've got a fix for the CI failure making it's way to prod.

PR #660, opened 2016-11-10
New tool to create Jenkins jUnit

edsantiago:


edsantiago:
DO NOT MERGE. Experimental.

Is there any way you can run ADEPT using this instead of the existing result2junit? You need to feed it --name, e.g.

    ./tests/docker/bin/results2junit --name centos-7-2-ADEPT-ci-58 results/default

(unless there's a way for the script to determine the 'centos-7-2-etc' string, please let me know).

It's perl. I know. That's because I'm in a big hurry, may be taking tomorrow off, and wanted speed of development. All I want to know is if my schema for class and test names will work.

cevich:
> Is there any way you can run ADEPT 

Possibly.  I could add a check if ``results2junit`` exists (no need for bin), and if so, and it's executable, use it instead of the one in client/tools...here

https://github.com/cevich/autotest-docker-enabled-product-testing/blob/master/roles/autotested/tasks/produce_results.yml

Yes, I can pass the job name into it...

cevich:
@edsantiago does your script send to stdout just like ``results2junit.py``?

edsantiago:
Yes, it's almost identical to the other. It doesn't emit timestamp in the testsuite element, and doesn't emit useful error messages, but right now I don't have time to work on those. All I want to know is if the classname and name structure makes reasonable sense, and if a failure in one run will be seen by Jenkins. And OBTW I don't treat TEST_NA as failure.

cevich:
okay.  https://github.com/cevich/autotest-docker-enabled-product-testing/pull/67

cevich:
There we go, got it now.  One more run to confirm then I'll merge.

cevich:
jenkins, test this please.

cevich:
okay, maybe it will work, err, ADEPT may still be having trouble with rebooting and enabling inconsequential services (like docker)...but...it'll run your script if it's in ``autotest/client/tests/docker/`` and executable.  :grinning:

edsantiago:
As in, I should mv it from run to the base directory?

cevich:
> As in, I should mv it from run to the base directory?

Ya.  There's still a problem with the non-atomic VMs not rebooting before services are setup.  Though you should (hopefully) still get your script to run on the Atomic Hosts.  In any case, we'll get back on 'er come Monday.  Havagoodweekend.

cevich:
Mmmm, thought prob. is, I dunno if we have perl on atomic host.  Going to guess not :frowning: 

sigh

never easy.

cevich:
@edsantiago I don't grok perl so you're on your own as far as debugging.  I don't care that it's in perl though, beyond maintaining consistency, it shouldn't matter much in the long run.

Once you get it working, it might be nice to re-write it in python and/or update ``.travis.yml`` so some kind of testing can happen on it.  I have no idea what's available for perl but there is a travis [guide here](https://docs.travis-ci.com/user/languages/perl/) and a [perl6 one here](https://docs.travis-ci.com/user/languages/perl6/).

edsantiago:
Yeah, python rewrite will take me a week or two (I suck at python). This is just a proof of concept to see if the results are useful enough to spend that time.

cevich:
> I suck at python

Ummmm, gunna toss a big fat pile of BS on that, in terms of the history of this project's contributors, you're up near the top of the python-foo charts in my book :grinning:

I merged https://github.com/cevich/autotest-docker-enabled-product-testing/pull/69 with your fix and my ``list-units`` -> ``list-unit-files`` un-goof :blush:  up the stack, so you should start getting full runs on all VMs (instead of so many outright failures b/c docker isn't running).

cevich:
The last CI run was started before I merged your ADEPT fix, so `"results2junit: missing RESULTS-DIR argument; try results2junit --help"``.  is expected.  Anyone on the 'approved' list can issue the special re-test magic incantation...

cevich:
jenkins, test this please.

cevich:
...and it'll do as it's told.  Should be more useful output this time.

edsantiago:
>The last CI run was started before I merged your ADEPT fix

Nerts. Is that because of devel vs master? I pushed my changes because I saw your adept push.

cevich:
> Nerts. Is that because of devel vs master? I pushed my changes because I saw your adept push.

Exactly.  ADEPT CI uses the ADEPT PR commit + Autotest-docker master.  The Docker-autotest CI uses the PR code + ADEPT master.

The only way (currently) to use some custom mashup of the different repos/branches is to setup your own test Jenkins server, manipulate the ``.gitmodules`` to the main JJB repo. (internal), cuss and swear a lot, patch up the JJB's to use different github credentials, curse jenkins some more, press the 'build' button and pray it works (with more silently belittling).

That said, if you think it would be helpful to have additional jobs that run on alternate branches combinations, it can be done.  There's also this card-  https://trello.com/c/xRqq6XzW/280-cn-docker-autotest-ci-optimization  

I was thinking of just scraping a CSV list out of the PR description, but the features is not written yet, so there's time to design the format to meet our needs. If you think that includes specifying alternate commit-ids, or selecting a sub-set of VM-types, or whatever else, it's possible to do.

edsantiago:
> if you think it would be helpful to have [yet more combinatorial explosions]

/me faints

cevich:
/me faints

lol.  Well, the results look better :+1:

edsantiago:
They do? I've been looking at them and am confused in a few places, e.g. why some results still show "autotest.docker". I also think I didn't implement TEST_NA properly.

cevich:
> They do?

Yeah, there's not a bunch of missing results and it's easier to tell exactly what failed on which VMs.

This will cleanup the tags back to just having the sequence numbers-

```diff
diff --git a/control b/control
index 96b5d60..4fdf794 100644
--- a/control
+++ b/control
@@ -15,11 +15,8 @@ import os.path
 import logging
 import collections
 import ConfigParser
-from binascii import crc32
 from socket import gethostname
 
-MYHOSTNAME = gethostname()
-
 def log_list(method, msg, lst):
     """
     Call method for msg, then every item in lst
@@ -494,7 +491,7 @@ class Step(collections.Callable):
         if advance:
             self.context.index += 1
         # Long names make aweful looking results
-        self.tag = "%d_%d" % (self.context.index, abs(crc32(MYHOSTNAME)))
+        self.tag = str(self.context.index)
 
     def __call__(self):
         self._mangle_syspath()
```

cevich:
Ahh, I C a problem with ADEPT passing in the entire ``ansible_hostname`` - it includes the job number, which will be different for every job (in theory).  Let me see about whittling that down a bit...

results2junit
```diff
@@ -242,7 +242,7 @@ sub dump_junit {
             $test_suite_name, $class_name;
         printf $out_fh " name=\"%s\"", $test_name;
         printf $out_fh " time=\"%d\"",  $s->{run_time};
-        if ($s->{status} eq 'GOOD') {
+        if ($s->{status} eq 'GOOD' || $s->{status} eq 'TEST_NA') {
             print $out_fh "/>\n";
```
cevich: dunno that we want TEST_NA == PASS.  It will confuse the idea of 'test shouldn't run here' with 'test was successful'.  In other words, if the test really starts running (and it shouldn't be) we won't notice (from the high-level results view)

edsantiago: Hm, but then are we really going to want to see test failures each time because CentOS is running a rhel-only plugin, or docker-1.10 is running a 1.12-only test? Are there different levels of messages other than <error> and <failure> ?

edsantiago: Found <skipped/>. Testing now.

cevich:
Great!  Glad it was just a 'typo'.  I'm still working on the naming thing, it's a challenging problem.  The names must be unique between hosts for the run, but consistent across runs.  Unf. the VM hostnames are different, they must be universally unique, yet somewhat human readable.  I'll figure something out.

edsantiago:
jenkins, test this please.

cevich:
``16:34:36 FAILED - RETRYING: TASK: openstack : peon is in status active (1 retries left).``

I've seen that before, no idea why it happens.  For some reason the VM never boots up, it just stays powered off.  If you catch it in time, you can hit the power-on button and it'll continue.

cevich:
Hmmm, looks like it's tracking skips now which is good, but it's still picking up the old format junits....OIC...it's because on the atomic hosts, there's no perl:

```
23:15:44 TASK [autotested : Attempt to run alternate results2junit] *********************

23:15:45 changed: [centos-latest-7-2-ADEPT-dockerautotest-59-317e56fb5dab]

23:15:45 changed: [centos-7-2-ADEPT-dockerautotest-59-5e5d415ed705]

23:15:45 fatal: [centosatomic-latest-7-2-ADEPT-dockerautotest-59-c8d109e1b430]: FAILED! => {"changed": true, "cmd": "/var/lib/autotest/client/tests/docker/results2junit --name centosatomic-latest-7-2-ADEPT-dockerautotest-59-c8d109e1b430 results/default > /var/lib/autotest/client/results/default/results.junit", "delta": "0:00:00.003278", "end": "2016-11-14 23:15:44.470310", "failed": true, "rc": 126, "start": "2016-11-14 23:15:44.467032", "stderr": "/bin/sh: /var/lib/autotest/client/tests/docker/results2junit: /usr/bin/perl: bad interpreter: No such file or directory", "stdout": "", "stdout_lines": [], "warnings": []}

23:15:45 ...ignoring

23:15:45 fatal: [centosatomic-7-2-ADEPT-dockerautotest-59-46fcc2baf196]: FAILED! => {"changed": true, "cmd": "/var/lib/autotest/client/tests/docker/results2junit --name centosatomic-7-2-ADEPT-dockerautotest-59-46fcc2baf196 results/default > /var/lib/autotest/client/results/default/results.junit", "delta": "0:00:00.003403", "end": "2016-11-14 23:15:45.325573", "failed": true, "rc": 126, "start": "2016-11-14 23:15:45.322170", "stderr": "/bin/sh: /var/lib/autotest/client/tests/docker/results2junit: /usr/bin/perl: bad interpreter: No such file or directory", "stdout": "", "stdout_lines": [], "warnings": []}

23:15:45 ...ignoring
```

The results of your script are getting mixed, so that's actually okay.  When this is re-written in simple python, it'll work fine.  I just checked, and ``xml.etree.ElementTree`` and ``xml.dom`` are both available.  IMHO, ElementTree is the easiest to work with inside simple loops.

edsantiago:
Perl on Atomic: yeah, I know. I'm having to just ignore those for now. Rewrite: so, is this worth pursuing?

cevich:
Yeah, I think so.  A simple custom script like this is FAR simpler, more reliable, and mutable than the autotest one*.  I'm almost done with ADEPT changes that will expose peon metadata as ``host_vars``.  That way, the hosts in ``autotested`` role will have access to their creation details.  That means they can retrieve their definition names instead of compound job-names.  This will be portable across all jobs and test-harnesses (if there ever are any others).

*fun fact: results2junit was written someone at cononnical who's an an ex-redhatter and openly cautioned anyone against ever actually looking at the code :grinning:

cevich:
[(?) [CI/CN] SPIKE: HOWTO: De-fubar jUnit results](https://trello.com/c/c9wTyspx/323-ci-cn-spike-howto-de-fubar-junit-results)

cevich:
```
             <failure message="Sub-subtest failures: set(['git_path'])">11/21 14:52:18 ...
```

I think you need to also escape all ``<`` and ``>`` characters w/in the message (``&lt;`` and ``&gt;``).  Though maybe there's an easier way.

edsantiago:
I'm not sure what that's in reference to. Did you miss `xml_escape()`?

cevich:
Oops, misread that results.junit line, it's fine and I see you're escaping quotes.  Anyway.  FWIW.

cevich:
The test results displayed in Jenkins are a bit confusing because it's trying to show a comparison of current vs previous runs (which won't line up).  Though the new results look **much** better with your script.  When this is ready to merge (or whenever you like), I can do a "clean-room" run of it to show what the results really look like w/o the historical garbage.

results2junit
```diff
@@ -0,0 +1,289 @@
...
+    """
+    Set of test results, initialized from a list read by blah blah
+    """
+
+    def __init__(self, name, results):
+        self.input_name = name
+        # Strip gobbledygook from rhel-latest-7-3-ADEPT-full-26-a451b130bf40
+        self.name = re.sub(r'-\d+-[0-9a-f]{11,}$', '', name)
+        self.results = results
+        self.count = {'tests': 0, 'failures': 0, 'errors': 0, 'skipped': 0}
```
cevich: It's presumptuous at this level to try and sanitize the name based on any kind of format not tied to this code or project.  At best It's  obfuscating a legitimate name (erasing clues to what may have gone wrong).  At worst, it causes results to clash because the name becomes common vs some other host.  Best to just take what the caller provided. 

NOTE: [I've fixed the long ugly names being passed by ADEPT](https://github.com/cevich/autotest-docker-enabled-product-testing/commit/d31ad6875a50f614ff22efc6f8b1c6214bc26573)

results2junit
```diff
@@ -0,0 +1,289 @@
...
+        Add one testcase to our list. Update relevant counts (total number
+        of tests, number of failures/errors/skipped).
+        """
+        self.testcases.append(testcase)
+
+        self.count['tests'] += 1
+        if testcase.category in self.count:
+            self.count[testcase.category] += 1
+        elif testcase.category:
+            print "********* WEIRD %s" % testcase.category
```
cevich: hehe, probably better just toss an exception here.

results2junit
```diff
@@ -0,0 +1,289 @@
...
+            if tc.name == 'garbage_check':
+                tc.classname = self.testcases[i-1].classname
+                tc.name = self.testcases[i-1].name + '--garbage-check'
+
+    def write_xml(self, outfile):
+        """
+        Writes a junit XML file.
+        Yes, this is hand-created XML and it's nauseating. That's how
+        the existing autotest JUnit_api.py does it, and that's the quickest
+        way for me to do it right now.
```
cevich: ``ElementTree`` is super easy to work with, especially with all the looping you've got going on here.  It also takes care of all the escaping and encoding stuff for you.  Here's a cheap example:

```python
from xml.etree import ElementTree as ET

tree = ET.ElementTree(ET.Element("testsuites"))
sweet = ET.Element("testsuite", dict(name="foo", timestamp="now", tests="1", errors="2", failures="3", skipped="4"))

props = ET.Element("properties")
for _dict in [dict(foo="bar"), dict(sna="foo")]:
    props.append(ET.Element("property", **_dict))

tc_elements = [ET.Element("testcase", _dict)
               for _dict in [dict(classname="a", name="<b>"),
                             dict(classname="c", name="\"d\"")]]
sweet.append(props)
sweet.extend(tc_elements)
tree.getroot().append(sweet)
# tree.write(filepath, encoding="UTF-8")
ET.dump(tree)
```

The other way to do it is with the ``SubElement()`` method.  Check out the python docs for it, it's ***way*** simpler than what you probably imagine.

results2junit
```diff
@@ -0,0 +1,289 @@
...
+                                 'run_time': timestamp - self.start_times.pop(),
+                                 'result': self.messages.pop()})
+            return
+
+        if parts[5]:
+            self.messages.append(parts[5])
+
+    def pop(self):
+        """
+        Allows treating this obj as a list. Returns the last element.
```
cevich: Why not just subclass ``list`` or ``collections.MutableSequence``?  Then you're guaranteed the python list-API is present and correct.

results2junit
```diff
@@ -0,0 +1,289 @@
...
+
+    parser.add_argument('autotest_results_dir',
+                        help='path to directory containing "status" file;' +
+                        ' this is also where we write results.junit')
+    return parser.parse_args()
+
+
+def main(argv=None):
+    """
+    Entry point. FIXME
```
cevich: Naa, this is totally fine to do.

results2junit
edsantiago: I guess I don't understand this comment. I will try to explain the code, and if that doesn't address your concern, please let me know.

The `status` file contains the following status strings: GOOD, TEST_NA, ERROR, and FAIL. `results.junit`, however, needs to have different strings: Jenkins expects `skipped` (for TEST_NA), `errors` (plural) for ERROR, and `failures` (not just fail) for FAIL. I need to convert the `status` strings to the form expected by Jenkins. Does that address your question?

I've refactored that code to reduce some duplication and, I hope, make it easier to understand. Coming in next commit.

results2junit
edsantiago: I simply removed the message. If this situation ever happens, it'll be a silent failure.

results2junit
edsantiago: `xml.etree` doesn't have prettyprinting and [looks like it never will](https://bugs.python.org/issue14465). To me that shows bad judgment, which makes me question what else is wrong with the module. 
Hand-formatting XML is pretty bad judgment on my part, too, but IMO the advantage of readable XML (for debugging) outweighs the ugliness of this code.

results2junit
```diff
@@ -0,0 +1,289 @@
...
+            xml += "/>\n"
+        return xml
+
+    @property
+    def stacktrace(self):
+        """
+        Returns the full docker-autotest stacktrace for a given testcase,
+        as read from the .ERROR file in this test's subdirectory.
+        """
+        # Given 'docker/subtests/docker_cli/run_volumes.46_756384709',
```
edsantiago: This code will handle both.

results2junit
edsantiago: Because all I need is `pop`, and because if I subclass `list` I'm not sure how to store other state in `self`. I'm running into diminishing returns here, and don't want to spend a week polishing this up to perfection.

results2junit
cevich: Noooooo, ``xml.etree`` is perfect, right along with all the rest of python 2.7, no bugs whatsoever, none! :grinning: 

Generally XML stuff isn't intended for human consumption, just like JSON.  It's not uncommon for it's parsers to omit the pretty printing part.  Properly checking the content of XML is done by a specialized tooing with defined schema (probably already written for jUnit).  In any case, ``etree`` is virtually guaranteed not to make formatting/syntax mistakes.  Where as a myriad of string-formatting ops can easily mess that up.  If you really need pretty-printing, perhaps one of the other xml modules could be used for that part?

I guess my main argument is that using ET will make _this_ code simpler, (and therefor easier to read and maintain).  Once the output is good and working, nobody's going to look at it by eye again, ever.  If needed, there are plenty of XML/HTML pretty-formatting tools out there for one-off use.  The long-term maintenance and clarity of _this_ code is much more important IMHO.

cevich:
I think I can compromise on leaving ET to Spielberg, if you can drop the ``name`` sanitizing.  We have bigger lakes-full-of-fish that need frying.  Deal?

edsantiago:
Thanks, that helps a great deal. I hope to have a new push, with tests, this afternoon.

cevich:
Results look great, and somehow all the historical failures went *poof*.  odd.  Going to do one final peek-through then will merge this.  Nice quick work fixing a nasty nasty nasty problem.

PR #661, opened 2016-11-16
New test: docker-1.12 regression in cp

edsantiago:
https://github.com/docker/docker/issues/27773


edsantiago:
Test passes on docker-1.10.3-57.el7, fails on docker-1.12.3-4.el7

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+
+    See: https://github.com/docker/docker/issues/27773
+    """
+
+    def initialize(self):
+        super(volume_mount, self).initialize()
+
+        vol = 'v_' + utils.generate_random_string(8)
+        subargs = ['create', '--name', vol]
+        mustpass(DockerCmd(self, 'volume', subargs).execute())
```
cevich: (didn't read bug)  If you don't pass ``--name`` docker will create a non-clashing name automatically and splat it to stdout.

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+        self.sub_stuff['volume_name'] = vol
+
+        # First container: bind-mounts new volume, then unpacks two
+        # tarballs into it. I don't know why it has to be two tarballs
+        # and not one, but the tar thing (vs plain copy) has to do
+        # with pivot_root. Whatever that is.
+        c1 = 'c1_' + utils.generate_random_string(8)
```
cevich: ``dockertest.containers.DockerContainers.get_unique_name()`` makes sure name's characters conform and that it doesn't already "randomly" exist.

cevich: Also, pylint/travis is likely to be unhappy with a two-character variable name.

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+        # Path is not configurable because it's hardcoded in the tarballs
+        vol_binding = vol + ':/.imagebuilder-transient-mount'
+        subargs = ['--name', c1, '-v', vol_binding, 'busybox']
+        mustpass(DockerCmd(self, 'create', subargs).execute())
+        self.sub_stuff['container1'] = c1
+        for ab in ['a', 'b']:
+            tar_file = 'cp_volume_mount_data_%s.tar' % ab
+            tar_path = os.path.join(self.parent_subtest.bindir, tar_file)
+            # Failure mode only happens if cp is via redirected stdin.
```
cevich: redirected ``stdin`` == ``AsyncDockerCmd.execute(_int_)`` where_int_ is the ``r`` side from ``os.pipe()``.  Docker will block until you start filling the ``w`` side, and it will complete when you close it....

...Yeah, ``utils.run()`` is probably a whole lot easier.  

You should use ``self.config[docker_path]`` though, b/c at some point we may be testing multiple clients :D

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+                   '-v', mp + '/1:/mountfile',
+                   'busybox', 'cat', '/mountfile']
+        mustpass(DockerCmd(self, 'create', subargs).execute())
+        self.sub_stuff['container2'] = c2
+
+    def run_once(self):
+        super(volume_mount, self).run_once()
+        subargs = ['-a', self.sub_stuff['container2']]
+        result = mustpass(DockerCmd(self, 'start', subargs).execute())
+        # This command fails on docker-1.12
```
cevich: maybe ``self.logdebug("This command fails on docker-1.12")`` :grin:

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+        self.sub_stuff['start_result'] = result
+
+    def postprocess(self):
+        super(volume_mount, self).postprocess()
+        self.failif_not_in('value3b', self.sub_stuff['start_result'].stdout)
+
+    def cleanup(self):
+        super(volume_mount, self).cleanup()
+        for which in ['1', '2']:
```
cevich: How about ``for key in ['container1', 'container2']:``?

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+            key = 'container' + which
+            if key in self.sub_stuff:
+                subargs = ['-f', self.sub_stuff[key]]
+                mustpass(DockerCmd(self, 'rm', subargs).execute())
```
cevich: If this fails, then a volume (below) will be leaked.

cevich: Also, if it fails on ``container1``, it could leak ``container2`` ***and*** the volume.

subtests/docker_cli/cp/cp.py
```diff
@@ -207,3 +208,72 @@ def verify_files_number(self, copied_number, expected_number):
...
+
+        if 'volume_name' in self.sub_stuff:
+            vol = self.sub_stuff['volume_name']
+            mustpass(DockerCmd(self, 'volume', ['rm', vol]).execute())
```
cevich: Probably best just not use ``mustpass()``.  Fire-and-forget (or at least log any error) is normally a better strategy in ``cleanup()``.

subtests/docker_cli/cp/cp.py
edsantiago: If I have to debug using logs, I'm going to find `v_AXwjOgbV` easier to interpret as "the volume in question" than `0ac7fea86ed4f93a028005c657369107e3346630623ab93b32ef824bde21055b`.

subtests/docker_cli/cp/cp.py
edsantiago: Fixed (using docker_path, not the horrible redirection). Thanks.

subtests/docker_cli/cp/cp.py
edsantiago: Well, it shouldn't. And on today's -8 build it fails, but differently. Sigh. Changing this to `# This is the command that fails in broken docker-1.12 builds`

subtests/docker_cli/cp/cp.py
edsantiago: I can live with that. Fixed.

subtests/docker_cli/cp/cp.py
edsantiago: Fixed (removed `mustpass`). Good catch.

subtests/docker_cli/cp/cp.py
cevich: Sure, but then validating the name is acceptable and doesn't clash becomes a pain.  By the time anyone investigates, all the other volumes are gone, so PD is impossible.  Maybe a "cheap" compromise could be to just ``logdebug`` the output from ``docker volume ls``?

cevich:
There's a problem in CI.  The centosatomic_latest is running docker 1.10 instead of 1.12.  Investigating...

cevich:
...okay, this is a bug in ADEPT, not the test.  Working on a fix.

PR #662, opened 2016-11-18
Followup to new volume_mount test: SELinux fix

edsantiago:
docker-1.12.3-8.el7 brings in the fix for #27773 [1] but
introduced a new failure, an AVC when starting the container
using a bind mount directly on the filesystem. Per Dan Walsh
the AVC is correct behavior, and docker-1.10 passing is a
bug. The correct way to bind mount is to include :Z - see
docker-run(1) for details.

 [1] https://github.com/docker/docker/issues/27773


cevich:
Ahh, that must be related to this problem too then: https://url.corp.redhat.com/450f665

edsantiago:
No idea if they're related. Guess I'll have to take a look at that failure as a new one...

PR #663, opened 2016-11-28
Experiment: Customize log file/command sysinfos

cevich:
The defaults for both the job-level and test-level are hard-coded behind
a private attribute w/in autotest's job class.  Start out by completely
wiping out all entries.  Then, walk static mappings of keyword
arguments.  Each is passed to the proper API call for adding per-job
or per-test files/commands to be collected or run.

### DIFF
--- a/control
+++ b/control
@@ -621,19 +621,6 @@ class StepInit(Context, collections.Callable):
                      job.sysinfo.boot_loggables,
                      job.sysinfo.test_loggables]:
             item.clear()
-        for _type, items in SYSINFOS.items():
-            # Avoid calling private job._add_sysinfo_loggable()
-            if _type == 'job':
-                on_every_test = False
-            else:
-                on_every_test = True
-            for dargs in items.get('commands', []):
-                dargs.update({"on_every_test": on_every_test})
-                job.add_sysinfo_command(**dargs)
-            for _string in items.get('logfiles', []):
-                dargs = {"file": _string,
-                         "on_every_test": on_every_test}
-                job.add_sysinfo_logfile(**dargs)
 
     def __call__(self):
         """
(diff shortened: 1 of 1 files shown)


cevich:
***N/B:*** Totally untested, experimental, and likely radioactive.

cevich:
[(1)(CI) Collect set of files between every test](https://trello.com/c/JduUdbRn/335-1-ci-collect-set-of-files-between-every-test)

cevich:
jenkins, test this please.

control
```diff
@@ -582,6 +595,26 @@ class StepInit(Context, collections.Callable):
...
+        for _type, items in SYSINFOS.items():
+            # Avoid calling private job._add_sysinfo_loggable()
+            if _type == 'job':
+                on_every_test = False
+            else:
+                on_every_test = True
+            for dargs in items.get('commands', []):
+                # Easier than dealing with parameter conflicts
+                dargs.update(dict(on_every_test=on_every_test))
+                job.add_sysinfo_commnad(**dargs)
```
edsantiago: This looks like it's misspelled (command), but in my testing I don't see any errors logged. (I also don't see the files getting created)

control
```diff
@@ -582,6 +595,26 @@ class StepInit(Context, collections.Callable):
...
+            for dargs in items.get('logfiles', []):
+                dargs.update(dict(on_every_test=on_every_test))
+                job.add_sysinfo_logfile(**dargs)
```
edsantiago: Fails because dargs is a string ('logfiles' is a list of strings, not list of dicts). This loop needs to be rethought. E.g.:

    for logfile in items.get('logfiles', []):
                job.add_sysinfo_logfile(logfile, on_every_test=on_every_test)

That doesn't crash, but again, I still don't see the sysinfo files getting created.

cevich:
There, fixed the title.  Ya, don't spend time on this @edsantiago it's one of those "suppose to be easy" things that's really not.  I had something similar working before, so I know it's possible, just need to play some more.

cevich:
correction, remembering now, I almost had something working before, but gave up and took your approach.  Then later I ditched that, and just stuck it into the docker version test :D

cevich:
jenkins, test this please

cevich:
I think I prefer having this in a test.  The control file is big and complex enough, it doesn't need more added.

PR #665, opened 2016-11-29
Add output.newlines module

cevich:
This module provides several classes needed for monitoring and testing
stdout/stderr output from asynchronous commands.  Output is fully
buffered, and peek-able.  There are classes to support both positive and
negative matches within specific timeout limits.

### DIFF
--- a/dockertest/output/__init__.py
+++ b/dockertest/output/__init__.py
@@ -9,4 +9,5 @@ from . dockerversion import DockerVersion
 from . texttable import TextTable, ColumnRanges
 from . validate import OutputGood, OutputGoodBase, OutputNotBad
 from . validate import wait_for_output, mustpass, mustfail
-from . newlines import NewLines, NewlineMatchTimeout, NewlineMatch, NewlineMatchPeek, NoNewlineMatch
+from . unseenlines import UnseenLines, UnseenlineMatchTimeout, UnseenlineMatch
+from . unseenlines import UnseenlineMatchPeek, NoUnseenlineMatch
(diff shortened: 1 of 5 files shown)


cevich:
**N/B:** ***This is very old code, extracted from an internal test.  It's _mostly_ ready for merging with some minor polishing.  The internal-only test is in much worse shape, so is not included here.***

edsantiago:
Where is this intended to be used? And, is it too late to come up with a different name? 'Newline' is pretty closely associated with `^J`; I find myself constantly trying to remind myself "no, here newline means something else".

cevich:
> Where is this intended to be used?

I cc'd you an e-mail about this.

> And, is it too late

Absolutely not too late, only my lack of imagination stands in the way :grinning:

cevich:
**``groklines``** ?

edsantiago:
`tail`?

cevich:
**``SeizeLines``** ?

cevich:
**``SenseLines``**

**``CatchLines``**

cevich:
**``DashedLines``** ?

cevich:
**``TailHarvest``**?

edsantiago:
Changing to a verb (grok, sense, seize, harvest) makes me uncomfortable because the class was originally written as a noun. How about `UnseenLines`, or perhaps just `Unseen` (since it's a subclass of `output`)? Alternatively, `NewOutput`?

cevich:
Hmmmm, I like "Unseen"... but it would be nice to include a verb ("new" can be an adverb, no?).   Something will strike us that works, as soon as we don't think about it :grinning:

cevich:
I like ``UnseenLines``, let's go with that.

PR #666, opened 2016-12-01
results2junit: handle repeated test calls

edsantiago:
Chris pointed out that some ADEPT runs will invoke a particular
subtest multiple times, e.g. when testing docker vs docker-latest
or perhaps when running stress tests.

Differentiate these by keeping track of test names, and when
we encounter one we've seen before, append '(N)' to the test
name to make it unique. (Except for garbage_check).

### DIFF
--- a/results2junit
+++ b/results2junit
@@ -70,6 +70,7 @@ class TestSuite(object):
             testcase = TestCase(result, self.name)
             self.add_testcase(testcase)
 
+        self._indicate_multi_test_runs()
         self._consolidate_garbage_checks()
 
     def add_testcase(self, testcase):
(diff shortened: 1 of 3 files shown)


results2junit
```diff
@@ -78,15 +80,32 @@ class TestSuite(object):
...
+        """
+        Append '(N)' to all tests that are run more than once
+        """
+        for name, count in self.testname_count.items():
+            if count > 1 and name != 'garbage_check':
+                seq = 1
+                for tc in self.testcases:
+                    if tc.name == name:
+                        tc.name += '(%d)' % seq
+                        seq += 1
```
cevich: Wouldn't it be easier to just accept the autotest `name.tag` but strip out the '.' character?  The `.tag` thing is more-or-less hard-coded into Autotest itself, it's never going to go away or change. That will work with your ``garbage_check`` magic and should fix the Jenkins shenanigans no?

results2junit
```diff
@@ -164,6 +183,12 @@ class TestCase(object):
         self.time = result['run_time']
         self.set_status(result)
 
+        # *Sob*. Jenkins intercepts URLs ending in '/run', redirecting them
+        # to a non-useful page. This makes it impossible to view failures
+        # in the run test. Workaround: use a different name in junit. Ugh.
+        if self.name == 'run':
+            self.name = 'docker_run'
+
```
cevich: This will confuse the hell outta anyone that goes looking for ``subtests/docker_cli/docker_run``.  It also won't match with the ``job_report.html`` which could be confusing.  TBH, it may be easier to just rename the test.  We did the same thing in other cases where there were conflicts (i.e. w/ python reserved words).

cevich:
LGTM, just two minor concerns that I can live with if my comments are wrong / unfounded.

results2junit
cevich: yep, nothing wrong with that at all, I do it all the time :grinning:

results2junit
cevich: oh my bad, I thought your code was stripping out the tag number, then adding another number back onto the name.  Nevermind then.

cevich:
jenkins, test this please.

cevich:
@edsantiago 

New idea to address this, #666 #679 and #695 all in one go:

Rewrite ``results2junit`` with this sequence:

1. Read, parse, buffer ``known_failures.csv``.
1. Backup all original ``status*`` file**s**.
1. Read ``status.json``, filter known failures, output new ``status``,  ``status.json``, and ``results.junit``.
1. Call ``$AUTOTEST_PATH/client/shared/report.py"

Benefits:

* No ``dockertest`` code needs to be modified
* Test results/reporting are not tightly-coupled to ``subtest`` module (or classes) behavior
* Existing unittests can be re-used
* Known failures are filtered from results
* PR #666 can be addressed
* PR #695 can be closed
* A window is opened to (between 3 and 4) to affect the ``HTML`` rendering (modify templates and/or patch module code).
* BZ usage in control file can be killed
* All results and ``HTML`` report contents match

PR #667, opened 2016-12-05
Remove individual subtest timeouts

cevich:
It's not possible to have a variable timeout across steps, they all need
to be defined up-front before anything runs.  This meant quick tests
disproportionately utilize their time vs long tests.  Instead, remove
the individual timeouts and fudge-factor.  Autotest will then impose the
global (``TIMEOUT``) value.  This could cause a cascade failure of all
subsequent tests, if one consumes too much time.  This could make
results hard to debug in the worst-case, but the best-case (no timeouts)
should make the run-to-run results more consistent.


edsantiago:
LGTM. My test run finished just fine.

PR #668, opened 2016-12-06
Log docker rpm version

edsantiago:
Write a new sysinfo/docker_rpm file containing
the output of 'rpm -q docker[*]' (* - docker or
docker-latest). This requires exporting which_docker()
from dockertest.docker_daemon.

Purpose: post-run analysis can look at failures,
then cross-reference against a TBD table of known
subtest failures in specific docker builds.

### DIFF
--- a/pretests/log_versions/log_versions.py
+++ b/pretests/log_versions/log_versions.py
@@ -1,0 +1,50 @@
+r"""
+Summary
+-------
+
+Preserve docker versions in state files.
+
+Operational Summary
+-------------------
+
+#. Run 'docker version'; use DockerVersion module to parse output
+#. Run 'rpm -q docker|docker-latest' (whichever one is in use)
+#. Preserve output in sysinfo files
+
+"""
+
+import os.path
+from autotest.client import utils
+from dockertest import subtest
+from dockertest.dockercmd import DockerCmd
+from dockertest.output import DockerVersion, mustpass
+from dockertest.docker_daemon import which_docker
+
+
+class log_versions(subtest.Subtest):
+
+    def run_once(self):
+        """
+        Determine the installed version of docker, and preserve it
+        in a sysinfo file.
+        """
+        super(log_versions, self).run_once()
+        cmdresult = mustpass(DockerCmd(self, "version").execute())
+        docker_version = DockerVersion(cmdresult.stdout)
+        info = ("docker version client: %s server %s"
+                % (docker_version.client, docker_version.server))
+        self.loginfo("Found %s", info)
+        self.write_sysinfo('docker_version', info + "\n")
+        self.write_sysinfo('docker_rpm', self._docker_rpm())
+
+    def write_sysinfo(self, filename, content):
+        """
+        Write the given content to sysinfodir/filename
+        """
+        path = os.path.join(self.job.sysinfo.sysinfodir, filename)
+        with open(path, 'wb') as outfile:
+            outfile.write(content)
+
+    @staticmethod
+    def _docker_rpm():
+        return utils.run("rpm -q %s" % which_docker()).stdout
--- a/pretests/log_versions/log_versions.py
+++ b/pretests/log_versions/log_versions.py
@@ -35,7 +35,9 @@ class log_versions(subtest.Subtest):
                 % (docker_version.client, docker_version.server))
         self.loginfo("Found %s", info)
         self.write_sysinfo('docker_version', info + "\n")
-        self.write_sysinfo('docker_rpm', self._docker_rpm())
+        self.write_sysinfo('docker_rpm_active', self._rpmq(which_docker()))
+        self.write_sysinfo('docker_rpm_current', self._rpmq('docker'))
+        self.write_sysinfo('docker_rpm_latest', self._rpmq('docker-latest'))
 
     def write_sysinfo(self, filename, content):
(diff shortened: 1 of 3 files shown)


pretests/log_versions/log_versions.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Write the given content to sysinfodir/filename
+        """
+        path = os.path.join(self.job.sysinfo.sysinfodir, filename)
+        with open(path, 'wb') as outfile:
+            outfile.write(content)
+
+    @staticmethod
+    def _docker_rpm():
+        return utils.run("rpm -q %s" % which_docker()).stdout
```
cevich: This could fail for any number of reasons, best set ``ignore_errors=True``.

subtests/docker_cli/version/version.py
```diff
@@ -42,12 +41,6 @@ def postprocess(self):
         # Raise exception on Go Panic or usage help message
         outputgood = OutputGood(self.stuff['cmdresult'])
         docker_version = DockerVersion(outputgood.stdout_strip)
-        info = ("docker version client: %s server %s"
-                % (docker_version.client, docker_version.server))
-        self.loginfo("Found %s", info)
```
cevich: Fine to strip out the file-writing, but no harm in leaving the ``loginfo()`` in there, no?  If anything it continuously verifies the ``docker_version`` class is working.  It's failed one time that I remember, when docker's output format changed.

pretests/log_versions/log_versions.py
```diff
@@ -0,0 +1,50 @@
...
+Summary
+-------
+
+Preserve docker versions in state files.
+
+Operational Summary
+-------------------
+
+#. Run 'docker version'; use DockerVersion module to parse output
+#. Run 'rpm -q docker|docker-latest' (whichever one is in use)
```
cevich: Why limit it just to docker?  Howabout have the test accept a config. option with a list of rpm's to query?  e.g. docker relies on runc and there are other OCI nubbins we're interested in (generally).  Would be handy to have that data recorded also.

pretests/log_versions/log_versions.py
```diff
@@ -0,0 +1,50 @@
...
+        in a sysinfo file.
+        """
+        super(log_versions, self).run_once()
+        cmdresult = mustpass(DockerCmd(self, "version").execute())
+        docker_version = DockerVersion(cmdresult.stdout)
+        info = ("docker version client: %s server %s"
+                % (docker_version.client, docker_version.server))
+        self.loginfo("Found %s", info)
+        self.write_sysinfo('docker_version', info + "\n")
+        self.write_sysinfo('docker_rpm', self._docker_rpm())
```
cevich: Might be nice to have a separate file for each RPM name.  e.g. ``docker_rpm``,  ``docker-latest_rpm``, etc.  That way, anything that might need to parse it can just read the entire contents in.

cevich:
'nother way to do this: https://github.com/autotest/autotest-docker/pull/663

pretests/log_versions/log_versions.py
edsantiago: My immediate goal is to have my jenkins-analyze script cross-check a list of known failures in known docker/docker-latest builds. For that, it needs to check the rpm currently in use. Fetching `docker_rpm` is an easy way to do that.

Logging all rpm versions (docker, docker-latest, runc, plugins, selinux) is preferable but that wasn't my immediate goal; and it actually blocks my immediate goal because I then need to devise a mechanism to log whether docker or docker-latest is in use. Any suggestion on how to solve that?

pretests/log_versions/log_versions.py
edsantiago: Urp. I meant to do that. Thanks.

subtests/docker_cli/version/version.py
edsantiago: Added back. Thanks.

pretests/log_versions/log_versions.py
cevich: Based on dealing with it in ADEPT, IMHO the server (daemon) versions is the most important.  You can determine which is being used by a zero exit status from ``systemctl --quiet is-active docker`` and ``systemctl --quiet is-active docker-latest``.  I'd suggest treating them equally, on the off-chance they're both enabled (a problem we need to know about).

In fact, maybe it would be good to just dump all the systemd unit details.  ``systemctl --plain --no-legend --no-pager list-unit-files`` will show everything along with it's enabled/disabled state.  ``systemctl --type=service --state=active --plain --no-legend --no-pager list-units | awk -e '{print $1" "$4}'``  will show the running/exited state of all active (started) services.

Hmmm, maybe this is too bug for just one subtest, maybe have separate sub-subtests for rpm vs systemd?

edsantiago: Since this is not relevant to my current purpose, I would prefer to deal with this in a separate PR.

edsantiago:
@cevich do you think we could get this in before your PTO? Unfortunately I can't test it because of the OSP7 outage.

cevich:
@edsantiago Okay, this works as expected and LGTM.  Shall I merge?

edsantiago:
Sure; it can still be useful for postprocessing. Thanks.

PR #669, opened 2016-12-08
Configured test images exist between every test

cevich:
It's possible that one test deletes one of the test images.  This could
leave the next test's docker command performing an implied pull.
However, some number of tests have hard-coded timeouts for specific
docker commands (like list).  This means the implied pull can easily
cause timeout errors where pulling the image is otherwise
inconsequential.  Fix this by symlinking ``docker_test_images`` from
pretests into intratests.  This way it will run before all and in-between
every test.

### DIFF
--- a/intratests/docker_test_images
+++ b/intratests/docker_test_images
@@ -1,0 +1,1 @@
+../pretests/docker_test_images


cevich:
@edsantiago @afomm Simple solution to some other timeout errors (see description).

edsantiago:
Lots of errors in my testing: 

    Found leftover images from prior test: [DockerImage(full_name:registry.access.redhat.com/rhel7/rhel LONG_ID:sha256:e8e3aaf82af53563382e4079d833ebeda12f65a7afb9ab98345decbceece7e20 CREATED:12 days ago SIZE:192.5 MB)]

cevich:
> Lots of errors in my testing:

hmmm, blah, that sux.  Okay, maybe it's not the right fix or not a problem worth tinkering with.

PR #670, opened 2016-12-08
Strip out nonprintable characters from xml

edsantiago:
Some test somewhere is generating an error message
with ANSI codes: ^[[91mpermission denied. This is
making Jenkins choke with invalid XML.

Workaround: strip off chars not in string.printable.

A better solution, suggested by Chris, would be to
use autotest.client.utils.strip_console_codes().
Unfortunately that requires having autotest in
our include path, and we can't assume that when
we run.


edsantiago:
I really don't want to copy-paste and duplicate that code.

If our goal is to strip ANSI sequences, it can be done with a regex on `'\e\[[^m]*m'`. That strips noise characters, at the cost of possibly missing other situations we haven't thought of yet.

If the goal is to give jenkins some usable xml, we can use this crappy code and accept the cost that a human will have to deal with `[1mpermission denied`.

Life sucks. We have to make some tradeoffs.

cevich:
Why re-invent the wheel?  If you can't justify duplicating, then why not including autotest?  At least for CI  purposes, this script is never going to run where autotest isn't.  I can even make sure ``AUTOTEST_PATH`` and ``PYTHON_PATH`` are set correctly when it's called.

cevich:
Opened https://github.com/cevich/autotest-docker-enabled-product-testing/pull/83

cevich:
Hehe, complicated monster could just be another word for "collaborative development" 😀

It's only our inner-perfectionist that's thumbing-nose and crying "Thou Shalt Not Duplicate!".  In reality, code is duplicated all the time, you just don't notice!  Some say there is no original code, all of it was copy-pasted (in some part) from elsewhere.  Okay, I said that, but s/code/ideas/g and you get the same thing.

So, why kludge when solid, tested, time-proven, and never-changing code is readily available?  It's much cheaper in the long-run (where maintenance cost weighs heavy) to create the code w/ as few kludges as possible.

PR #672, opened 2016-12-09
Noprinting

cevich:


cevich:
N/B: Haven't fixed up the unittests yet.

cevich:
😢 travis is reporing a false-positive

edsantiago:
Something got lost here: `nonprintables_check()` is not being invoked

cevich:
> Something got lost here: nonprintables_check() is not being invoked

Assuming u mean the build test.  It's a bug in the test, just fixed.  Otherwise, where are you seeing it not being invoked?

There is a problem with it catching crap on many ``docker run`` calls though.  Where we probably want it to skip that check.  Considering fixing each one, or making this ``nonprinting_check()`` a special-case.  Another idea is we always send ``-e TERM=dumb`` for most ``docker run calls``.

edsantiago:
The string 'nonp' only appears once in the diffs. The string 's_check' only appears once in the entire source tree.

edsantiago:
I think you have diffs that you haven't pushed to github.

cevich:
> The string 'nonp' only appears once in the diffs.

That class just calls all of it's methods that end in ``_check``.  If that's what you're worried about.

> I think you have diffs that you haven't pushed to github.

Yes I did, was missing your results choocher update.

Going to leave this one sit for now, not sure how to handle the ``docker run`` case.  Leaning toward making the ``nonprintables_check()`` an opt-in item instead of default-for-all.

cevich:
jenkins, test this please.

cevich:
jenkins, test this please.

cevich:
Tested this manually, results look pretty good now.  Only three failing tests, will investigate causes next...

cevich:
Found + Fixed why non-printing check was not running.  I see it working now.  More testing...

cevich:
jenkins, test this please.

cevich:
Bah, okay, CI seems "normal" now and I'm just adding fixes for other non-related stuff.  Need to stop doing that.

One more go around with CI, then this one is ready to merge IMHO.

@edsantiago @afomm Could you do a final review please?

edsantiago:
I see much there to like, but it's hard to review: the diffs I see include a hodgepodge of unrelated changes. Would it make sense to split those off into separate, smaller PRs?

cevich:
> Would it make sense to split those off into separate, smaller PRs?

Yes, I started adding on "this will only take 5 minutes".  Good idea, I'll see about splitting.

cevich:
force-pushed just the bear necessities (rawr!)

edsantiago:
Two comments overall; neither a blocker, and I have no good suggestions for fixing:

  1) With the true-false form, a fail in `nonprintables_check()` causes `OutputGood` to emit the failing stdout/stderr -- including, of course, the nonprintables. Solutions are hard: changing the callback mechanism to a filter, or adding (duplicate) nonprintable-cleaning code to OutputGood. I think best ROI is simply to document it as a warning comment in OutputGood.

  2) Why are there so many exclusions? Is docker really emitting so many nonprintables as a regular part of its output? If so, is this test really adding value?

cevich:
> Why are there so many exclusions? Is docker really emitting so many nonprintables as a regular part of its output? If so, is this test really adding value?

There are simply many times where docker needs to just pass through data unmodified.  For example, docker "run", "log", "attach", "exec", and even some particular invocations of build, it's the contained-process that's producing the output.  Docker's just the middle man.  This non-printables checker is only intended for checking output generated by docker itself, and only when stdout/stderr are NOT going to a tty/pty (the vast majority of ours).

I looked into doing this checker both ways (enabled by default or disabled).  It wound up having it disabled by default meant updating many more calls.

PR #673, opened 2016-12-12
Pr653 cleanup

edsantiago:
Remove some duplication; remove unnecessary intermediate steps
(i.e. use same variable names in .ini file and in source files;
don't edit files in-place, just transform as we copy); expand
brvtd variable names into legible ones; remove trailing whitespace;
add a missing super()

### DIFF
--- a/config_defaults/subtests/docker_cli/systemd.ini
+++ b/config_defaults/subtests/docker_cli/systemd.ini
@@ -1,8 +1,30 @@
 [docker_cli/systemd]
-docker_timeout = 300
-#: Systemd unit file
+subsubtests = systemd_pull, systemd_build, systemd_run
+
+[docker_cli/systemd/systemd_pull]
+#: systemd unit file
+sysd_unit_file = docker-pull.service
+#: name of an image to pull
+image_name = centos
+
+[docker_cli/systemd/systemd_build]
+#: systemd unit file
+sysd_unit_file = docker-build.service
+#: name of an image to build
+image_name = centos_p4321
+#: build options when building an image
+unit_opts = --force-rm
+#: for Dockerfile FROM variable 
+img_frm = centos:latest
+
+[docker_cli/systemd/systemd_run]
+#: systemd unit file
 sysd_unit_file = p4321.service
-#: Systemd unit file destination
-sysd_unitf_dest = /etc/systemd/system/p4321.service
-#: build image name
-name = p4321
+#: name of an image build and run
+image_name = centos_p4321
+#: unit file build options when building an image
+unit_opts = --force-rm
+#: build options 
+build_opt = %(unit_opts)s -t %(image_name)s
+#: for Dockerfile FROM variable 
+img_frm = centos:latest
--- a/config_defaults/subtests/docker_cli/systemd.ini
+++ b/config_defaults/subtests/docker_cli/systemd.ini
@@ -1,30 +1,31 @@
 [docker_cli/systemd]
 subsubtests = systemd_pull, systemd_build, systemd_run
 
+#: base image for Dockerfile. Not used in every test.
+dockerfile_base_image = centos:latest
+#: systemd unit file (basename only). Each subsubtest has a file by this
+#: name in our source dir; we copy it into /etc/systemd/system, replacing
+#: a given string `foo` inside curly braces with the corresponding `foo`
+#: config setting defined below.
+sysd_unit_file = FIXME
+
 [docker_cli/systemd/systemd_pull]
-#: systemd unit file
 sysd_unit_file = docker-pull.service
 #: name of an image to pull
 image_name = centos
 
 [docker_cli/systemd/systemd_build]
-#: systemd unit file
 sysd_unit_file = docker-build.service
 #: name of an image to build
 image_name = centos_p4321
 #: build options when building an image
 unit_opts = --force-rm
-#: for Dockerfile FROM variable 
-img_frm = centos:latest
 
 [docker_cli/systemd/systemd_run]
-#: systemd unit file
 sysd_unit_file = p4321.service
 #: name of an image build and run
 image_name = centos_p4321
 #: unit file build options when building an image
 unit_opts = --force-rm
-#: build options 
+#: build options
 build_opt = %(unit_opts)s -t %(image_name)s
-#: for Dockerfile FROM variable 
-img_frm = centos:latest
--- a/subtests/docker_cli/systemd/systemd.py
+++ b/subtests/docker_cli/systemd/systemd.py
@@ -1,86 +1,160 @@
-"""
+r"""
 Summary
 -------
-Testing basic container as systemd service
+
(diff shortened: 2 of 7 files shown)


edsantiago:
I don't think this worked properly. It's showing a huge set of changes, not just mine.

Anyway, I'm heading out for a run while I've got time. I'll check back in after scrum.

cevich:
> It's showing a huge set of changes, not just mine.

Looks fine here.  Shows Afom's two commits and your 'cleanup' commit.  That's what it should look like.  When it's good-to-go, just do one final ``git rebase -i --autosquash upstream/master``, marking his 'fixup' as a 'fixup'.  Finally force-push, and we should have two commits showing here, one of his and one of yours.

cevich:
Looking into the travis sphinx barf...

config_defaults/subtests/docker_cli/systemd.ini
```diff
@@ -1,8 +1,30 @@
 [docker_cli/systemd]
-docker_timeout = 300
-#: Systemd unit file
+subsubtests = systemd_pull, systemd_build, systemd_run
+
+#: base image for Dockerfile. Not used in every test.
+dockerfile_base_image = centos:latest
+#: systemd unit file (basename only). Each subsubtest has a file by this
+#: name in our source dir; we copy it into /etc/systemd/system, replacing
+#: any '{foo}' strings in it with the corresponding 'foo' config setting.
```
afomm: @edsantiago - thanks for cleaning it up. It is much cleaner & shorter than what I had.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,140 @@
...
+        # Copy unit file into place, with translations.
+        bindir = self.parent_subtest.bindir
+        unitfile = self.config['sysd_unit_file']
+        unitfile_src = join(bindir, unitfile)
+        self.sub_stuff['unitfile_dst'] = join('/etc/systemd/system', unitfile)
+        self.sed_file(unitfile_src, self.sub_stuff['unitfile_dst'])
+
+        # Dockerfile & requirements to tmp workdir (not all tests use this)
+        for f in ['Dockerfile', 'p4321-server.py']:
+            self.sed_file(join(bindir, f), join(self.tmpdir, f))
```
cevich: Travis/pylint may get mad by the ``f``` being too short.  It will probably let you use ``_`` though if you prefer.

edsantiago:
Force-pushed. Travis clean. Thanks for the feedback and for catching that weird brace error.

subtests/docker_cli/systemd/systemd.py
```diff
@@ -1,86 +1,140 @@
...
-        self.failif(not time_from_socket.isdigit(),
-                    "Data received from container is non-numeric: '%s'"
-                    % time_from_socket)
+        for act in 'start', 'status':
+            self.sysd_action(act, self.config['sysd_unit_file'])
+
+    def postprocess(self):
+        super(systemd_base, self).postprocess()
+        imgs_fulname = DockerImages(self).list_imgs_full_name()
+        names_lst = [x.split(':')[0].split('/', -1)[-1] for x in imgs_fulname]
```
cevich: Best keep the parsing centralized.  How about:

```python
    image_name = self.config['image_name']
    img_objs = DockerImages(self).list_imgs()
    # True if only repo (aka name) matches, with "latest" tag.  Ignore ``repo_addr`` and ``user`` components.
    name_matches = [img_obj.cmp_greedy(repo=image_name, tag="latest") for img_obj in img_objs]
    # At least one must match, don't care if more than one.
    self.failif(not max(name_matches), "Image name %s not found in %s" % (image_name, img_objs))
```

subtests/docker_cli/systemd/systemd.py
cevich: Seems a little dangerous to modify these, but as long as there's always/ever a 1:1 config-option token per file, it's okay.  If there's ever more than one token, and the test runs multiple times w/ different config, it's possible the files won't be modified correctly.  The "Proper" (pedantic) way of doing this is to copy everything to a tmpdir, modify and build from there.  However, this is fine for now.

edsantiago: pylint, pep8, Travis, none of them have complained. IMO (apparently shared by pylint) there are times when `f` and `i` are perfectly fine variables.

edsantiago: Can you explain your concern? I don't understand what you find dangerous. Your comment about "copy to a tmpdir" (which is exactly what this does) makes me wonder if you missed my rewrite of `sed_file`.

subtests/docker_cli/systemd/systemd.py
edsantiago: That code makes no sense to me (I have no idea what `cmp_greedy` is, or why the `max`. I find it less readable, but, shrug, I'll make the change. Testing now.

subtests/docker_cli/systemd/systemd.py
cevich: Okay, good by me then.

cevich: ...ahh, okay IC now, it builds from the tmpdir.  Good.

```
+        dkrcmd = DockerCmd(self, 'build', [self.config['build_opt'],
+                                           self.tmpdir])
```

subtests/docker_cli/systemd/systemd.py
cevich: I agree the "greedy" name sucks, wasn't my invention.  What it does is, compare only the non-None parameters.  This is important b/c sometimes docker likes to rename your images for you.  e.g. docker pull 'centos' might actually get you 'docker.io/centos'.  Since the test only deals with the image name, using the greedy compare ensures any extra added name components are ignored.

The ``max()`` does just what the comment says.  Yeah it's not that readable.  Maybe this way would be clearer:

```python
    found_it = False
    for img_obj in DockerImages(self).list_imgs():
        if img_obj.cmp_greedy(repo=image_name, tag="latest"):
            found_it = True
            break
    failif(not found_it, ...)
```

cevich:
CI was stuck setting up another job.  I applied a squirt of oil, it's running now.

edsantiago:
Code barfing now because `img_objs` is gone. Once again, a simple fix becomes a nightmare :-(. I'll check back in again in the morning and see if we can come up with something clean.

cevich:
Yes, clean is good.  I think my main concern is if ever the image name format or docker's behavior changes.  It's much easier to fix it in one central place than go hunting through tests one by one.  There are many ways to skin that cat, and it's likely I haven't thought up the simplest and/or it's not a problem that needs solving now. 

Hmmm.  What if we just had a config. option "expected_built_name", then ``self.failif(DockerImages(self).list_imgs_with_full_name(expected_built_name) == [], ...)`` ?

cevich:
Jenkins, test this please.

PR #674, opened 2016-12-13
Clean up leftover <none> images

edsantiago:
Once or twice or many times a month, a run fails with
repeated failures in each garbage_check step: "found
leftover <none>". I've just been ignoring those, manually
deleting the <none> images, and restarting tests.

Apparently garbage_check was supposed to deal with
those in a more dignified manner. This PR fixes the
bugs that were preventing it from doing so.


cevich:
fantastic!

PR #677, opened 2017-01-05
Addendum to #672

cevich:
Fixes several test failures on the latest/greatest RHEL w/ docker-1.12.5.

The 'display failing output check in build test' commit depends on #672 being merged first.

### DIFF
--- a/config_defaults/subtests/docker_cli/negativeusage.ini
+++ b/config_defaults/subtests/docker_cli/negativeusage.ini
@@ -64,6 +64,7 @@ subcmd = run
 subarg = -p,192.168.9.1:9000,%%(FQIN)s,/bin/true
 stderr = Invalid hostPort
 
+# Note: This exposes RH BZ 1393572 and will need updating if/when it's status changes.
 [docker_cli/negativeusage/iv4]
 subcmd = run
 subarg = -e,PATH=/tmp,%%(FQIN)s,true
--- a/subtests/docker_cli/build/build.py
+++ b/subtests/docker_cli/build/build.py
@@ -29,6 +29,7 @@ from dockertest.dockercmd import DockerCmd
 from dockertest.images import DockerImage
 from dockertest.images import DockerImages
 from dockertest.output import OutputGood
+from dockertest.output import OutputNotBad
 from dockertest.config import get_as_list
 from dockertest.xceptions import DockerTestFail
 from dockertest.xceptions import DockerTestError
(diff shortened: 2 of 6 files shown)


config_defaults/subtests/docker_cli/negativeusage.ini
```diff
@@ -66,8 +66,8 @@ stderr = Invalid hostPort
 
 [docker_cli/negativeusage/iv4]
 subcmd = run
-subarg = -e,PATH=/tmp,%%(FQIN)s,true
-stderr = exec\: \"true\"\: executable file not found in \$PATH
+subarg = -e,PATH=/tmp,%%(FQIN)s,tRuE
+stderr = .*tRuE.*not found
```
edsantiago: This masks bz1393572: we will never know if/when it is fixed, nor if it breaks again. Is this your intention?

subtests/docker_cli/build/build.py
```diff
@@ -208,22 +209,26 @@ def parse_postprocess_commands(self, build_def):
...
-            # Verify zero exit status + healthy output
-            if OutputGood(build_def.dockercmd.cmdresult,
-                          ignore_error=True):
+            # Verify zero exit status and healthy output
+            opg = OutputGood(build_def.dockercmd.cmdresult,
+                             ignore_error=True)
+            if opg:
                 return build_def.dockercmd.cmdresult.exit_status == 0
             else:
+                self.loginfo(str(opg))
```
edsantiago: Should this be `warn` (vs info)?

Also, given that there's an identical log below, would you consider adding a differentiator to each one? E.g.

    self.logwarning("OutputGood failed: %s" % opg)
....
    self.logwarning("OutputNotBad failed: %s" % opb)

That makes it easier for a human to determine which of those two logs triggered, and continue debugging from there.

subtests/docker_cli/build/build.py
```diff
@@ -208,22 +209,26 @@ def parse_postprocess_commands(self, build_def):
...
                 return False
         elif command == 'negative':
-            # Verify non-zero exit status + healthy output
-            if OutputGood(build_def.dockercmd.cmdresult,
-                          ignore_error=True,
-                          skip=['error_check']):
+            # Verify non-zero exit status and no panics
+            opb = OutputNotBad(build_def.dockercmd.cmdresult,
```
edsantiago: I'm really uncomfortable with the name `opb`, especially when contrasted to `opg`. I read opb as 'output bad', not 'output NOT bad'. A casual reader may make the same assumption. How about just renaming both opg and opb to `ok`?

config_defaults/subtests/docker_cli/negativeusage.ini
cevich: No it wasn't.  I had a feeling there was something more to the breakage here.  Thanks for the ref. BZ.  I think a better "fix" would be for me to just stick that BZ in a comment and drop all the other changes.

subtests/docker_cli/build/build.py
cevich: good idea.

subtests/docker_cli/build/build.py
cevich: %s/opb/opng/g

cevich:
@edsantiago Thanks.  All addressed and force-pushed.

subtests/docker_cli/build/build.py
```diff
@@ -208,22 +209,28 @@ def parse_postprocess_commands(self, build_def):
...
+                self.logwarn('Positive output expected but check failed: %s',
+                             str(opg))
                 return False
         elif command == 'negative':
-            # Verify non-zero exit status + healthy output
-            if OutputGood(build_def.dockercmd.cmdresult,
-                          ignore_error=True,
-                          skip=['error_check']):
+            # Verify non-zero exit status and no panics
+            opng = OutputNotBad(build_def.dockercmd.cmdresult,
```
edsantiago: That's still confusing! If the function is *NotBad*, I find a variable named *ng* misleading. Did you mean to make it *nb* ?

cevich: > Did you mean to make it nb ?

Uggg, yes...The number of letters has no impact on my ability to spell correctly.  Thankyou edautospellchecker 😀 

(really, I did mean ``opnb`` but my head is elsewhere)

cevich: oops, there's a bug in the printing part too (wrong variable)

PR #679, opened 2017-01-05
Experimental: ignore failures in known-problem subtests

edsantiago:
https://trello.com/c/qqlVHKRX/353-5-silence-known-ci-failures

A common pattern seems to be:

  1) New docker introduces a regression
  2) Wait months for docker to fix
  3) Wait more months for new RPM build

It is frustrating to see those failures over and over on each
test run.

This commit introduces a mechanism for associating a docker build
(exact NVR), a subtest, and a comment describing the known failure.
If we see this failure, we treat it as a skipped test. Bonus: if
we see a failure in a subtest, and there's a known-failure for
a different NVR but same NV (e.g. docker-1.12.5-6 but not -7),
emit a warning near the failure message; this can help a
maintainer update the known-failures list.

Problems this solves:

  * constant failures in ADEPT nightly runs with same NVR
  * constant failures in ADEPT runs on github PRs

Problems this does not solve:

  * ADEPT runs on new builds which do not address known failures
  * ADEPT runs on docker PRs which "  "    " "    " "    " "
    (theoretical concern only: that is still just a pipe dream)

### DIFF
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -19,14 +19,56 @@ import copy
 from ConfigParser import Error
 from autotest.client.shared.error import TestError, TestNAError
 from autotest.client.shared.version import get_version
-from autotest.client import test
+from autotest.client import test, utils
 import version
 import config
 import subtestbase
+from docker_daemon import which_docker
 from xceptions import DockerTestFail
 from xceptions import DockerTestNAError
 from xceptions import DockerTestError
 from xceptions import DockerSubSubtestNAError
+
+
+MEMO_CACHE = {}
+
+
+def docker_rpm():
+    """
+    Returns the full NVRA of the currently-installed docker or docker-latest
+    """
+    if 'docker_rpm' not in MEMO_CACHE:
+        rpm = utils.run("rpm -q %s" % which_docker()).stdout.strip()
+        MEMO_CACHE['docker_rpm'] = rpm
+    return MEMO_CACHE['docker_rpm']
+
+
+def known_failures():
+    """
+    Returns a dict containing known test failures. Primary key is
+    subtest name (e.g. docker_cli/foo/bar), value is another dict
+    whose key is docker NVRA (e.g. docker-1.12.5-8.el7.x86_64),
+    value of that is a string description of the problem (e.g.
+    a bz number and comment).
+
+    FIXME: file path is hardcoded for development purposes; it
+    should be centrally accessible somewhere, possibly provided
+    at setup time by ADEPT.
+    """
+    if 'known_failures' not in MEMO_CACHE:
+        known = {}
+        import csv
+        # FIXME: how to determine path to top-level directory?
+        path_known = '/var/lib/autotest/client/tests/docker/known_failures.csv'
+        with open(path_known, 'rb') as csv_fh:
+            csv_reader = csv.reader(csv_fh)
+            for row in csv_reader:
+                if row[1] not in known:
+                    known[row[1]] = {}
+                # Each row is: NVR, subtest, description
+                known[row[1]][row[0]] = row[2]
+        MEMO_CACHE['known_failures'] = known
+    return MEMO_CACHE['known_failures']
 
 
 class Subtest(subtestbase.SubBase, test.test):
@@ -429,12 +471,44 @@ class SubSubtestCaller(Subtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
+            # Ignore known problems in specific docker builds
+            if self.is_known_failure(name):
+                self.final_subsubtests.add(name)
             # Treat N/A as passed, even though this may hide obscure failures
             if isinstance(detail, TestNAError):
                 self.logwarning("Treating TestNAError as PASS")
                 self.final_subsubtests.add(name)
             # cleanup() will still be called before this propagates
             raise exc_info[0], exc_info[1], exc_info[2]
+
+    def is_known_failure(self, name):
+        """
+        Do we have a registered known failure in this subtest when running
+        on the currently-installed docker? Return True if so.
+        Side effect: log warning messages to help human debuggers.
+        """
+        # e.g. docker_cli/subtest/subsubtest
+        fullname = os.path.join(self.config_section, name)
+        known = known_failures()
+        if fullname not in known:
+            return False
+        docker_nvr = docker_rpm()
+        if docker_nvr in known[fullname]:
+            why = known[fullname][docker_nvr]
+            self.logwarning("Known test failure on %s: %s", docker_nvr, why)
+            return True
+        # This exact NVR is not known to fail, but perhaps this is just a
+        # new build that doesn't fix a previously-known problem. Look for
+        # known failures in other builds of this N-V (name, version) and
+        # issue a heads-up if appropriate.
+
+        def _nv(nvr):
+            return nvr[:nvr.rfind('-')]
+        docker_nv = _nv(docker_nvr)
+        if docker_nv in [_nv(x) for x in known[fullname]]:
+            self.logwarning("This test is known to fail in other %s builds",
+                            docker_nv)
+        return False
 
     def run_all_stages(self, name, subsubtest):
         """
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -484,14 +484,19 @@ class SubSubtestCaller(Subtest):
             why = known[fullname][docker_nvr]
             self.logwarning("Known test failure on %s: %s", docker_nvr, why)
             return True
-        # This exact NVR is not known to fail, but perhaps this is just a
-        # new build that doesn't fix a previously-known problem. Look for
-        # known failures in other builds of this N-V (name, version) and
-        # issue a heads-up if appropriate.
 
+        # This exact NVR is not known to fail. What about NV?
         def _nv(nvr):
             return nvr[:nvr.rfind('-')]
         docker_nv = _nv(docker_nvr)
+
+        if docker_nv in known[fullname]:
+            why = known[fullname][docker_nv]
+            self.logwarning("Test expected to fail on all builds of %s: %s",
+                            docker_nv, why)
+            return True
+
+        # No known failures for NVR or NV. What about other builds of same NV?
         if docker_nv in [_nv(x) for x in known[fullname]]:
             self.logwarning("This test is known to fail in other %s builds",
                             docker_nv)
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -468,21 +468,22 @@ class SubSubtestCaller(Subtest):
             # cleanup() will still be called before this propagates
             raise exc_info[0], exc_info[1], exc_info[2]
 
-    def is_known_failure(self, name):
+    def is_known_failure(self, subsubtestname):
         """
         Do we have a registered known failure in this subtest when running
         on the currently-installed docker? Return True if so.
         Side effect: log warning messages to help human debuggers.
         """
         # e.g. docker_cli/subtest/subsubtest
-        fullname = os.path.join(self.config_section, name)
+        fullname = os.path.join(self.config_section, subsubtestname)
         known = known_failures()
         if fullname not in known:
             return False
         docker_nvr = docker_rpm()
         if docker_nvr in known[fullname]:
             why = known[fullname][docker_nvr]
-            self.logwarning("Known test failure on %s: %s", docker_nvr, why)
+            self.logwarning("%s: Known failure on %s: %s",
+                            subsubtestname, docker_nvr, why)
             return True
 
         # This exact NVR is not known to fail. What about NV?
@@ -492,14 +493,25 @@ class SubSubtestCaller(Subtest):
 
         if docker_nv in known[fullname]:
             why = known[fullname][docker_nv]
-            self.logwarning("Test expected to fail on all builds of %s: %s",
-                            docker_nv, why)
+            self.logwarning("%s expected to fail on all builds of %s: %s",
+                            subsubtestname, docker_nv, why)
             return True
 
-        # No known failures for NVR or NV. What about other builds of same NV?
+        # No known failures for NVR or NV. What about other builds of same NV
+        # or a related one? These messages are informational only, intended
+        # as hints for a test engineer trying to understand new failures.
         if docker_nv in [_nv(x) for x in known[fullname]]:
-            self.logwarning("This test is known to fail in other %s builds",
-                            docker_nv)
+            # e.g. docker is 1.12.5-6, we have an exception for 1.12.5-5
+            self.logwarning("%s is known to fail in other %s builds",
+                            subsubtestname, docker_nv)
+        elif docker_nv.count('.') > 1:
+            def _nv_base(nv_orig):
+                return nv_orig[:nv_orig.rfind('.')]
+            docker_nv_base = _nv_base(docker_nv)
+            if docker_nv_base in [_nv_base(_nv(x)) for x in known[fullname]]:
+                # e.g. docker is 1.12.6-1, we have an exception for 1.12.5-5
+                self.logwarning("%s is known to fail in other %s.x builds",
+                                subsubtestname, docker_nv_base)
         return False
 
     def run_all_stages(self, name, subsubtest):
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -30,17 +30,11 @@ from xceptions import DockerTestError
 from xceptions import DockerSubSubtestNAError
 
 
-MEMO_CACHE = {}
-
-
 def docker_rpm():
     """
     Returns the full NVRA of the currently-installed docker or docker-latest
     """
-    if 'docker_rpm' not in MEMO_CACHE:
-        rpm = utils.run("rpm -q %s" % which_docker()).stdout.strip()
-        MEMO_CACHE['docker_rpm'] = rpm
-    return MEMO_CACHE['docker_rpm']
+    return utils.run("rpm -q %s" % which_docker()).stdout.strip()
 
 
 def known_failures():
@@ -50,25 +44,18 @@ def known_failures():
     whose key is docker NVRA (e.g. docker-1.12.5-8.el7.x86_64),
     value of that is a string description of the problem (e.g.
     a bz number and comment).
-
-    FIXME: file path is hardcoded for development purposes; it
-    should be centrally accessible somewhere, possibly provided
-    at setup time by ADEPT.
     """
-    if 'known_failures' not in MEMO_CACHE:
-        known = {}
-        import csv
-        # FIXME: how to determine path to top-level directory?
-        path_known = '/var/lib/autotest/client/tests/docker/known_failures.csv'
-        with open(path_known, 'rb') as csv_fh:
-            csv_reader = csv.reader(csv_fh)
-            for row in csv_reader:
-                if row[1] not in known:
-                    known[row[1]] = {}
-                # Each row is: NVR, subtest, description
-                known[row[1]][row[0]] = row[2]
-        MEMO_CACHE['known_failures'] = known
-    return MEMO_CACHE['known_failures']
+    known = {}
+    import csv
+    path_known = os.path.join(config.CONFIGCUSTOMS, 'known_failures.csv')
+    with open(path_known, 'rb') as csv_fh:
+        csv_reader = csv.reader(csv_fh)
+        for row in csv_reader:
+            if row[1] not in known:
+                known[row[1]] = {}
+            # Each row is: NVR, subtest, description
+            known[row[1]][row[0]] = row[2]
+    return known
 
 
 class Subtest(subtestbase.SubBase, test.test):
--- a/dockertest/subtest.py
+++ b/dockertest/subtest.py
@@ -460,6 +460,7 @@ class SubSubtestCaller(Subtest):
             exc_info = self.exception_info["exc_info"]
             # Ignore known problems in specific docker builds
             if self.is_known_failure(name):
+                self.logwarning("Treating %s subsubtest failure as PASS", name)
                 self.final_subsubtests.add(name)
             # Treat N/A as passed, even though this may hide obscure failures
             if isinstance(detail, TestNAError):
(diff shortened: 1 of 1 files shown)


edsantiago:
DO NOT MERGE. Proof-of-concept only. Still needs a defined mechanism for finding the known-failures file

dockertest/subtest.py
```diff
@@ -19,16 +19,58 @@
...
+
+    FIXME: file path is hardcoded for development purposes; it
+    should be centrally accessible somewhere, possibly provided
+    at setup time by ADEPT.
+    """
+    if 'known_failures' not in MEMO_CACHE:
+        known = {}
+        import csv
+        # FIXME: how to determine path to top-level directory?
+        path_known = '/var/lib/autotest/client/tests/docker/known_failures.csv'
```
cevich: How about ``config.CONFIGCUSTOMS``?  The config module ignores anything that doesn't end in ``.ini``.   It's handy to keep all the runtime configuration in one place for collection/archiving/reporting purposes.

dockertest/subtest.py
```diff
@@ -19,16 +19,58 @@
...
 import config
 import subtestbase
+from docker_daemon import which_docker
 from xceptions import DockerTestFail
 from xceptions import DockerTestNAError
 from xceptions import DockerTestError
 from xceptions import DockerSubSubtestNAError
 
 
+MEMO_CACHE = {}
```
cevich: Constants are okay, but global variables in modules can lead to complicated synchronization and unittesting problems down the line.

This is fine for a POC.  But consider if there's a better way/place to store this temporary stuff.    If the data can't be encoded in an instance somewhere (singleton pattern comes to mind), the easiest future-proof solution is to make it private (single leading underscore) and write a function that returns it when called.

dockertest/subtest.py
```diff
@@ -19,16 +19,58 @@
...
+            csv_reader = csv.reader(csv_fh)
+            for row in csv_reader:
+                if row[1] not in known:
+                    known[row[1]] = {}
+                # Each row is: NVR, subtest, description
+                known[row[1]][row[0]] = row[2]
+        MEMO_CACHE['known_failures'] = known
+    return MEMO_CACHE['known_failures']
+
+
```
cevich: Ahh okay.  Another idea:  Write ``known_failures()`` and ``MEMO_CACHE`` into an immutable singleton class in ``subtestbase.py`` (e.g. see ``config.Config.__new__()`` - minus all the dictionary copying crap).  Then have ``SubBase`` instantiate it as a class-attribute.  e.g.

```python
    ...
    #: Path to file indicating which Red Hat release this is
    redhat_release_filepath = "/etc/redhat-release"

    known_failures = KnownFailures()
```

Finally, stick ``is_known_failure()`` into ``SubBase`` where all descendants can call it. 

This way the data file only gets parsed once, and all subtests and sub-subtests can access it through API calls (like ``self.is_known_failure()``).

dockertest/subtest.py
```diff
@@ -429,13 +471,45 @@ def try_all_stages(self, name, subsubtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
+            # Ignore known problems in specific docker builds
+            if self.is_known_failure(name):
+                self.final_subsubtests.add(name)
```
cevich: I forgot if we discussed this or not...but if it's a known failure, why run the test at all?

Oh, also, what about standalone subtests (i.e. not sub-subtests), don't they get to eat cake too?

There may be a hook for the ``Subtest`` class in it's ``execute()`` method.  IIRC, that's an intermediate call by used by autotest itself to then call ``setup(), initialize(), run_once(), ...``.  Perhaps it could intercept calls to ``postprocess()``, and if ``is_known_failure()``, wrap it up into a ``try...except...`` where it can convert an actual faulure into a ``TestNAError``.  Just an idea.

edsantiago:
Whew! Resurfacing. New push coming, the TL;DR is:

  - globals : removed
  - singleton : abandoned. rathole.
  - subtest (vs subsubtest) : abandoned. rathole. Too hard, and not worth it for just two subtests.
  - new functionality: compare NV in addition to NVR. For those cases where, sigh, no fix is imminent.

In response to "why run the test at all": because the test may pass, especially with the NV exclusion (a new build *might* pick up a fix. Unlikely but not impossible).

cevich:
Great, I'm glad to see progress on this.

> subtest (vs subsubtest) : abandoned. rathole. Too hard, and not worth it for just two subtests.

You cannot predict what new tests will look like.  At some point in the (hopefully) near future, we'll have a more steady stream of them coming in.  It would be nice if the filtering could cope with any test result, not just the current majority.

I'm also leery about complicating the ``subtest`` further with this kind of thing.  It's been working/stable as-is for a really long time.  I'm not saying "no" absolutely, though, what's wrong with making this filtering part of your ``results2junit`` script?

In that script, you're already looping over the results anyway, excluding / skipping some at that level would be simpler.  The unittests for ``results2junit`` are much more results-focused (than ``subtests``) and you've already mocked up most of what's needed to support results-testing there.

edsantiago:
> You cannot predict what new tests will look like. At some point in the (hopefully) near future, we'll have a more steady stream of them coming in. It would be nice if the filtering could cope with any test result, not just the current majority.

Understood, but I have a simple clean 95%-good solution today. Making it 99% will take many more days of effort and increase risk. I choose the Good-Enough compromise.

> what's wrong with making this filtering part of your results2junit script?

That would simplify things, but unfortunately it's not helpful: results2junit does not alter `job_report.html`, it only alters what jenkins sees. What humans overwhelmingly see and use is `job_report.html`, especially the green/red GOOD/FAIL. To me, the overwhelmingly critical reason for this entire exercise is to see all-green in `job_report.html`. 

Maybe it's time to take a quick step back for reality check. Two questions: (1) is there something else that can alter `job_result.html`, more cleanly? (2) Is there something better than `job_report.html`, something perhaps generated from the junit file? (Keeping in mind that none of my test setups use jenkins). I keep wondering if there's something I've missed.

cevich:
> Understood, but I have a simple clean 95%-good solution today. Making it 99% will take many more days of effort and increase risk. I choose the Good-Enough compromise.

After todays "it's  a POC" discussion, this is perfectly fine.  I thought this was the final-product.  My bad.  Yep, good-enough is good-enough.

> That would simplify things, but unfortunately it's not helpful: results2junit does not alter job_report.html, it only alters what jenkins sees. What humans overwhelmingly see and use is job_report.html, especially the green/red GOOD/FAIL. To me, the overwhelmingly critical reason for this entire exercise is to see all-green in job_report.html.

The (internal) mandated format recording of results is and has always been: jUnit results are the final word and accepted format.  So anything that filters within/before we post up the jUnit results is totally cool.   I hate ``job_report.html`` **AND** we can't turn it off.  It's always run by Autotest no matter what 😱 and it's targeted at kernel-testing.  So IMHO, even the janky-Jenkins test-results view is *far* better / more relevant.

...tl;dr...

If/when we want/need "pretty" clicky-clicky reports, we're better off generating those ourselves where we can include only exactly the details/links we need instead of using the stoopid, not-even-generic  ``job_report.html`` it's worse than useless. 

Have I mentioned that I don't much like ``job_report.html``?

> (2) Is there something better than job_report.html, something perhaps generated from the junit file? (Keeping in mind that none of my test setups use jenkins). 

jUnit != Jenkins.  It's a format originaly tied to testing java code.  it's got nothing to do with Jenkins really.  It's just a standard test-results data format.  I haven't looked, but I'm sure there are TONS of tools out there for viewing/manipulating/massaging jUnit based data into pretty graphs and clicky-clicky HTML.  Likely none is going to 60% match what we want.  The tool / script we write ourselves can do way better than 60%, even if it's totally fugly.

> I keep wondering if there's something I've missed.

Don't worry, this is a much better problem to have than the alternatives.   It's easy to question/explain, but very difficult to un-know things 😀

edsantiago:
OK, so you don't like `job_report.html`. To me that implies there is something better. What do you use?

cevich:
> OK, so you don't like job_report.html. To me that implies there is something better. What do you use?

ATM, I use the "Test Report" as generated by Jenkins (from your filtered jUnits).  To dig deeper from there, I go directly to the archived Docker Auotest results and look at ``sysinfo``, then either the main ``client.INGO`` or the ``testname.INFO``.  If I still need more I go after the ``DEBUG`` versions of those files.  The exception is if I need details on how the test VM was setup/configured, then I have to go view the log from ADEPT.

edsantiago:
jenkins, test this please

edsantiago:
Jenkins, test this please.

cevich:
Hrmmm, not sure how/why it broke (yet) but working on a possible fix in https://github.com/cevich/autotest-docker-enabled-product-testing/pull/95 commit "Fix implied common-role dependency",  bb5353cc8e65328

edsantiago:
Whew, thanks. I spent a lot of time trying to find any form of error indication, got nowhere. I feel better knowing it isn't a trivial one to diagnose.

cevich:
>  I feel better knowing it isn't a trivial one to diagnose.

Aye, the snag is: I haven't (knowingly) touched anything to do with role dependencies, and it was working before (and is working for some jobs).  The fix makes the dependency explicit instead of implicit, but I'm not sure if it will fix this (and some other jobs too).

edsantiago:
Jenkins, test this please.

cevich:
Okay, I see ``config_custom/known_failures.csv`` in at least one of the results directories, so it's there for sure now 😄

cevich:
looking...

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
 
+    def is_known_failure(self, subsubtestname):
+        """
+        Do we have a registered known failure in this subtest when running
+        on the currently-installed docker? Return True if so.
+        Side effect: log warning messages to help human debuggers.
+        """
+        # e.g. docker_cli/subtest/subsubtest
+        fullname = os.path.join(self.config_section, subsubtestname)
+        known = known_failures()
```
cevich: Perhaps it's premature optimization, but e.g. the "help" test has MANY subsubtests.  As the CSV file grows larger (and/or system is under memory pressure) performance will take a big hit waiting for the disk inside a loops of loops.  One idea could be to cache the known_failures, something like:

```python
class SubSubtest(subtestbase.SubBase):
    ...
    #: For internal-use only, don't touch it!
    _kf_cache = None
    ...
    def is_known_failure(self, subsubtestname):
        ...
        if self._kt_cache is None:
            self.kt_cache = known_failures()
        known = self.kt_cache
        ...
```

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
+        if fullname not in known:
+            return False
+        docker_nvr = docker_rpm()
```
cevich: Same here, this value doesn't need to be retrieved more than once.  In fact, if one of the yum-related cron-jobs happens to be running, this could block for a really long time (waiting for an rpm DB lock).  Only an observation, it's not critical, and this is just a POC.

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
+            # Ignore known problems in specific docker builds
+            if self.is_known_failure(name):
```
cevich: Maybe add here: ``self.logwarning("Treating failing subsubtest %s as passing", name)``

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
+        docker_nvr = docker_rpm()
+        if docker_nvr in known[fullname]:
+            why = known[fullname][docker_nvr]
+            self.logwarning("%s: Known failure on %s: %s",
+                            subsubtestname, docker_nvr, why)
+            return True
+
+        # This exact NVR is not known to fail. What about NV?
+        def _nv(nvr):
+            return nvr[:nvr.rfind('-')]
```
cevich: This would be better expressed as a lambda:

```python
        _nv = lambda nvr: nvr[:nvr.rfind('-')]
```

Python nerds love them some lambdas!

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
+        docker_nv = _nv(docker_nvr)
+
+        if docker_nv in known[fullname]:
+            why = known[fullname][docker_nv]
+            self.logwarning("%s expected to fail on all builds of %s: %s",
+                            subsubtestname, docker_nv, why)
+            return True
```
cevich: This and the ``if docker_nvr in known[fullname]:`` (above) conditions could be tightened up a bit by turning it into a function/method.  Then calling that in turn with ``docker_nvr`` then ``docker_nv``.  More modular code tends to make unittesting easier too.  It's only repeated twice though, and this is a POC, so probably not a big deal.

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
+
+        # No known failures for NVR or NV. What about other builds of same NV
+        # or a related one? These messages are informational only, intended
+        # as hints for a test engineer trying to understand new failures.
+        if docker_nv in [_nv(x) for x in known[fullname]]:
+            # e.g. docker is 1.12.5-6, we have an exception for 1.12.5-5
+            self.logwarning("%s is known to fail in other %s builds",
+                            subsubtestname, docker_nv)
```
cevich: Eek, a third (and maybe a forth below).  Would be nice to see if these could be wrapped up in a more modular way.  It'll def. make unit-testing a lot easier.  Maybe a method like (possibly ill-conceived):

 ``def known_failure_action(self, needle, advisory=False, needle_dot=False)``
(where ``known = self.kt_cache``)

I think the main duplication is the messages are very similar, except in the advisory case.  Anyhoo, it's just an idea of a way so ``is_known_failure(..)`` could call ``known_failure_action(*kwargs)`` inside a loop over ``kwargs`` as ``[(docker_nvr,), (docker_nv,), (docker_nv,  True), (docker_nv,  True, True)]`` (or something to that effect).

Anyway...just an idea...send it to ``/den/null`` if you like, there's always more ``/dev/random`` (where it came from) 😀

cevich:
[[CI] Version-dependent testing](https://trello.com/c/tdAhquxo/316-ci-version-dependent-testing)

dockertest/subtest.py
```diff
@@ -429,13 +458,62 @@ def try_all_stages(self, name, subsubtest):
...
+        elif docker_nv.count('.') > 1:
```
cevich: Oh, one more quick thing.  This ``elif`` I didn't quite understand, would be good to have a comment about why having a ``.`` is a special case.

dockertest/subtest.py
```diff
@@ -19,16 +19,45 @@
...
+    a bz number and comment).
+    """
+    known = {}
+    import csv
+    path_known = os.path.join(config.CONFIGCUSTOMS, 'known_failures.csv')
+    with open(path_known, 'rb') as csv_fh:
+        csv_reader = csv.reader(csv_fh)
+        for row in csv_reader:
+            if row[1] not in known:
+                known[row[1]] = {}
```
edsantiago: > I think you need an else: raise(..) here. Without one, this will silently ignore duplicate entries where the docker NVRA happens to clash.

I'm a little confused; I'll try a brief run-through of how the code works, and let's see if we can get on common ground.

The known_failures.csv file is laid out with NVR first. That's because I, as maintainer of this file, want to see docker versions first. It's the most important for me, it lets me quickly see if there are records for the version(s) of docker that I'm testing in a given moment; it lets me add them quickly in one go; it gives me an at-a-glance insight into how buggy a given version may be. The subsubtest is second, then a reason string.

For purposes of the code, the most important field is the subsubtest. That lets the postprocessing code instantly check: "are there known problems with this subsubtest? No? Good, return." Then, if there are known problems, check NVR.

In diagram form:

    docker-1.12.5-*,docker_cli/negativeusage/iv4,"bz1393572: expect fix in docker-1.13?"

becomes:

  known['docker_cli/negativeusage/iv4'] = { 'docker-1.12.5-*': "bzetcetc" }

Does that help? If your comment still applies after this, could you rephrase it or add clarification? Thanks.

dockertest/subtest.py
edsantiago: Good idea, so done. (There's already a warning issued in `is_known_failure()` but this removes ambiguity, and I'm OK with having two contiguous warnings. Thanks.

dockertest/subtest.py
edsantiago: Clarification: `is_known_failure` only gets called on failure. There is no file reading, no performance penalty, on successful completion of a test.

That said, I spent far too long playing with caching (which you didn't like because there was a global) and singletons (which never actually cached anything) and Borg (likewise). I think optimizations will have to wait until there's an actual problem, and will have to be done by a more python-adept person than I.

dockertest/subtest.py
edsantiago: Same as above. I couldn't get caching working without using a global.

dockertest/subtest.py
edsantiago: Done. Thanks!

dockertest/subtest.py
edsantiago: You know how much I despise duplication. So as you might surmise, I already tried variations of that. (Not with a `self.kt_cache` because, as mentioned above, I couldn't get caches working). It was unbearably clumsy. That doesn't mean there isn't a clean solution, just that I wasn't able to find one. And before you try, note that each warning message is subtly but importantly different. Those differences are meant to help the engineer reviewing error results.

dockertest/subtest.py
edsantiago: There is, just four lines below. It doesn't directly address the dot question (which is, basically, "deal with docker-1.12.x-rrr but not docker-1.12-rrr") but I thought the example in the comment clarified the intention.

dockertest/subtest.py
cevich: Oooohhhh, this is my misunderstanding then, thanks for the 'diagram' form.   I had originally read 'docker NVR is the primary-key' (or mis-read).  So I thought it was ending up like:

``known['docker-1.12.5-*'] = { 'docker_cli/negativeusage/iv4': "bzetcetc" }``

okay, we're on the same page now and I agree with the structure.

dockertest/subtest.py
cevich: > Clarification: is_known_failure only gets called on failure. There is no file reading, no performance penalty, on successful completion of a test.

Oooohhhh, okay, my bad again.  I thought it was being called after every single one.  Errr, wait..no, it's github's fault!  I just looked, it did some code/diff-folding that totally covered up:

```python
 except Exception, detail:
    ...
```

right before ``if self.is_known_failure(name):``.  Though I suppose it's my fault for assuming the scope.  In that case, what you have here is fine.  No caching needed.
```

File: dockertest/subtest.py (hunk 6 above)
Comment (cevich) [inline dockertest/subtest.py]: FWIW: Should have been able to dangle it off of ``self`` (the subsubcaller instance) with no prob.  Just in case you do need a wider-scope variable.  That's the place for it.

File: dockertest/subtest.py (hunk 10 above)
Comment (cevich) [inline dockertest/subtest.py]: >  So as you might surmise, I already tried variations of that.

Surmise I did not.  No prob. then since you already tried.  I was wishfully hoping that some careful word-smithing of the messages could let them be combined somehow.  It is very particular though, and certainly can be left up to some future brilliant idea someone wakes up with.  Or not.

File: dockertest/subtest.py (hunk 11 above)
Comment (cevich) [inline dockertest/subtest.py]: Hmm, the example didn't dawn on me then.  Working in declarative languages like C has my brain stuck on "scroll up if you want to know what something is".  Anyway, if you think of it, probably wouldn't hurt to add something in case someone else is infected with declaritive-language-lookup-syndrome.  Maybe even like: ``elif docker_nv.count('.') > 1:  # See example below``

**Comment by cevich** [comment]
eek, you've posted ' 	review feedback, and wildcard fix'  and I just missed it.  Ewwww, that diff looks horrible (The "view changes one").  I really need to just look at the whole file, the diffs and the folding/hiding github does is really hurting my brain.  Will add that to my list for Monday morning.

File: dockertest/subtest.py [hunk 13]
```diff
@@ -429,13 +458,61 @@ def try_all_stages(self, name, subsubtest):
                               self.exception_info["error_source"],
                               detail)
             exc_info = self.exception_info["exc_info"]
+            # Ignore known problems in specific docker builds
+            if self.is_known_failure(name):
+                self.logwarning("Treating %s subsubtest failure as PASS", name)
+                self.final_subsubtests.add(name)
             # Treat N/A as passed, even though this may hide obscure failures
             if isinstance(detail, TestNAError):
```
cevich: Not sure this is a problem or not, just looking at the condition:  If a sub-subtest throws a ``TestNAError`` and that test also happens to satisfy ``is_known_failure()``.  The code will:

*  Log the traceback from the test
* Log it's a known failure, "treating it as a pass"
* Log, "Treating TestNAError as a PASS".

Seeing that, one question could be- Is this test Not applicable (to this environment / setup) or is it expected to fail?

Wondering if maybe the "TestNAError" should take priority...like:

```python
    if isinstance(detail, TestNAError):
        ...
    elif if self.is_known_failure(name):
        ...
    else:
        raise exc_info[0], exc_info[1], exc_info[2]
```

???
```

edsantiago: I don't know what possible circumstance could trigger this, because the workflow I envision is:

1. test fails in ADEPT
2. human looks at it, files a bz (or recognizes an existing one), adds exception to known-b0rked file

Raising `TestNAError` from within a subtest is when we have situations that are broadly tuned to a docker major version (1.12, 1.13, etc); exceptions are for specific builds and, at times, minor versions (1.12.5).

That said, let's pretend that we do get a `TestNAError` that also matches a known exception. In that case, the outcome is that there will be two warning messages. The outcome is the same (test is marked as passed). I can live with duplicate warnings because this is a situation that just shouldn't happen.

Does that address your question?

cevich:
okay.  As a PoC I think this is complete, and the concept has been proven (moving trello card).  For the [real-deal trello card](https://trello.com/c/tdAhquxo) maybe let's start a fresh PR.  ATM, I'm thinking "good enough" could just be to get this working for subtests (``if 'subsubtests' not in self.config``).  The rest seems to be about as refined as we can get it, and (clearly) it's working.  Maybe some things could/should move to ``subtestbase.py``, that's up to you.  Finally, we'll need at least some minimal unittest (don't go crazy, just very basic is fine).   Does that sound reasonable?

dockertest/subtest.py
cevich: > two warning messages.

That's what I was wondering about, yes the outcome is the same, except for maybe a confused developer at the other end.

Let's just leave this way for now, it's really not a big deal.

PR #680, opened 2017-01-06
Fix run_volumes_selinux depending on ls command

cevich:
This test will fail if the --scontext option happens to not exist.
In any case, options to the 'ls' command could be different across
distros / versions or change unpredictably (WRT docker autotest).
Instead, add an API call to retrieve the context directly from the
extended attribute value.  This works on all filesystems and whether or
not SELinux is enabled.

### DIFF
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -7,9 +7,15 @@ Low-level/standalone host-environment handling/checking utilities/classes/data
        in autotest!
 """
 
+import errno
 import subprocess
+# This always fails on travis with: no get in xattr (E0611)
+# dispite successful 'pip install pyxattr==0.5.1'.  A test
+# for this name has been added to .travis.yml directly.
+from xattr import get as getxattr  # pylint: disable=E0611
 
 
+# FIXME: pwd is misleading, it should be 'path'
 def set_selinux_context(pwd, context=None, recursive=True):
     """
     When selinux is enabled it sets the context by chcon -t ...
@@ -34,3 +40,19 @@ def set_selinux_context(pwd, context=None, recursive=True):
         raise OSError("Fail to set selinux context by '%s' (%s):\nSTDOUT:\n%s"
                       "\nSTDERR:\n%s" % (_cmd, cmd.poll(), cmd.stdout.read(),
                                          cmd.stderr.read()))
+
+
+def get_selinux_context(path):
+    """
+    When selinux is enabled, return the context of ``path``
+    :param path: Full or relative path to a file or directory
+    :return: SELinux context as a string or None if not set
+    """
+    try:
+        # may include null-termination in string
+        return getxattr(path, 'security.selinux').replace('\000', '')
+    except IOError, xcpt:
+        if xcpt.errno == errno.EOPNOTSUPP:
+            return None
+        else:
+            raise
--- a/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
+++ b/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
@@ -2,19 +2,18 @@
 Summary
 -------
 
-Tests the --volume ::Z feature (automatic selinux context setting).
+Tests the --volume ::Z and ::z feature (automatic selinux context setting).
 
 
 Operational Summary
 -------------------
 
-1.  Start container using volume with z/Z set
+1.  Start container using volume with z or Z set
 2.  Check context and file creation within the container
-3.  Start another container --volumes-from $cont1
-4.  Check context and file creation
-5.  Try writing using $cont1
-6.  Verify (only) all created files are present
+3.  Start another container --volumes-from $first_container
+4.  Verify context and file creation from first, second, and host
 """
+
 import os
 import shutil
 import tempfile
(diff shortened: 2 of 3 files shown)


cevich:
***Note:***Just the beginning of a fix.  Currently not 100% working.

cevich:
Okay, this is done, should be fixed for all distros and all filesystems also.

dockertest/environment.py
```diff
@@ -34,3 +35,18 @@ def set_selinux_context(pwd, context=None, recursive=True):
...
+
+def get_selinux_context(pwd):
+    """
+    When selinux is enabled, return the context of ``pwd``
+    :param pwd: target file or directory
+    :return: SELinux context as a string or None if not set
+    """
+    try:
+        # may include null-termination in string
+        return getxattr(pwd, 'security.selinux').replace('\000', '')
```
cevich: Probably a better way to do this.  Afraid some newer version of ``xattr`` will strip-null automatically, so didn't want to blindly discard the last character or assume it will be at the end.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -149,7 +136,8 @@ def touch_and_check(self, cont, volume, filename, context_pre,
                       get_selinux_context(volume))
         self.logdebug("Touching /tmp/test/%s in container"
                       % filename)
-        os.write(cont.stdin, "touch /tmp/test/%s\necho RET: $?\n" % filename)
+        # Don't just update directory entries with touch, add content too
```
edsantiago: Why? Since I don't understand why this is important, maybe the comment should include a reason? Also, the debug log above should change to 'Creating'.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -174,12 +162,14 @@ def check_all_files(self, volume, exp_files):
...
         """
+        # Make sure disk matches pagecache
+        utils.run("sync", 10)
         exp_files = set('%s/%s' % (volume, _) for _ in exp_files)
         act_files = set()
         for pwd, _, filenames in os.walk(volume):
             for filename in filenames:
                 act_files.add('%s/%s' % (pwd, filename))
-        self.failif(exp_files.symmetric_difference(act_files), "Not all files"
+        self.failif(exp_files.symmetric_difference(act_files), "Not all files "
```
edsantiago: Very minor nit: my habit is to put the space on the continuation line, not the continued line. This makes it very obvious to the reader that the string is continued from the previous line. Again, minor.

dockertest/environment.py
```diff
@@ -34,3 +35,18 @@ def set_selinux_context(pwd, context=None, recursive=True):
         raise OSError("Fail to set selinux context by '%s' (%s):\nSTDOUT:\n%s"
                       "\nSTDERR:\n%s" % (_cmd, cmd.poll(), cmd.stdout.read(),
                                          cmd.stderr.read()))
+
+def get_selinux_context(pwd):
```
cevich: IKR, who let that slip through! 😬   Probably, ya, I can fix.

dockertest/environment.py
```diff
@@ -34,3 +35,18 @@ def set_selinux_context(pwd, context=None, recursive=True):
...
+    """
+    When selinux is enabled, return the context of ``pwd``
+    :param pwd: target file or directory
+    :return: SELinux context as a string or None if not set
+    """
+    try:
+        # may include null-termination in string
+        return getxattr(pwd, 'security.selinux').replace('\000', '')
+    except IOError, xcpt:
+        if xcpt.errno & errno.EOPNOTSUPP:
```
cevich: sry brain-dead-friday 😢

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -24,26 +24,13 @@
 from dockertest import subtest
 from dockertest import dockercmd
 from dockertest import environment
```
cevich: probably not

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
cevich: ahh good catch.  okay, ya, I'll add comment about why:

Creating a file is a directory (file) update: a separate cache for some filesystems.
Writing to a file (that doesn't exist) touches all the caches (fs directory and pagecache).

This guarantees that when you 'sync', the file is there.  Some filesystems ignore the sync for files with only directory entry changes (until some timeout period happens).

cevich:
Thx Ed, I'll work on this more.

cevich:
*%&$*$(W*THN@!@@#@NB$#UBUNTU!!!!

cevich:
%&$$(W*THN@!@rrhg!#4453%/]+_B$#***UBUNTU!!!!***

dockertest/environment.py
```diff
@@ -34,3 +37,19 @@ def set_selinux_context(pwd, context=None, recursive=True):
         raise OSError("Fail to set selinux context by '%s' (%s):\nSTDOUT:\n%s"
                       "\nSTDERR:\n%s" % (_cmd, cmd.poll(), cmd.stdout.read(),
                                          cmd.stderr.read()))
+
+
+def get_selinux_context(path):
+    """
+    When selinux is enabled, return the context of ``pwd``
```
edsantiago: ``path``

cevich:
no idea how to satisfy that build dependency on ubuntu or why it won't just use the apt package 😖

dockertest/environment.py
cevich: oh thanx.

cevich:
Travis, please show me what names are in the module `xattr`: ...getxattr...get...
chug
chug
chug
```
(pylint) E0611: 12,0: : No name 'get' in module 'xattr'
```

PR #683, opened 2017-01-16
Fix missing pyxattr module in Fedora Atomic

cevich:
Also update documentation indicating hat libselinux-python or pyxattr
are required dependencies.  However, SELinux is not native under Ubuntu
(which travis CI runs) so some extra work will be needed to 100% convert
to using the preferred libselinux-python.  Added a TODO to this effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ install:
     - pip install Autotest==0.16.2
     - pip install inspektor==0.2.0
     - pip install unittest2==0.8.0
+    # TODO: Use libselinux-python equivilent instead
     - pip install pyxattr==0.5.1
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,25 +14,25 @@ matrix:
 git:
     submodules: false
 
+# Allow importing modules that can't be installed with pip
+virtualenv:
+    system_site_packages: true
+
 addons:
   apt:
     packages:
-        - python-pyxattr
-        - libattr1-dev
+        - python-selinux
 
 install:
     - pip install Pylint==1.5.5 Pep8==1.6.2 Sphinx==1.2.2
     - pip install Autotest==0.16.2
     - pip install inspektor==0.2.0
     - pip install unittest2==0.8.0
-    # TODO: Use libselinux-python equivilent instead
-    - pip install pyxattr==0.5.1
 
 script:
     - SPHINXOPTS="-W" make
     - ./run_checkdocs.py
     - ./run_unittests.sh
-    - python -c "import xattr; exit(int('get' not in dir(xattr)))"
     - ./run_pylint.sh --FF
     - inspekt style --disable E501,E265,W601,E402,E731,E221
 
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -7,7 +7,6 @@ Low-level/standalone host-environment handling/checking utilities/classes/data
        in autotest!
 """
 
-import errno
 import subprocess
 import selinux
 
@@ -31,6 +30,7 @@ def set_selinux_context(pwd, context=None, recursive=True):
     _cmd = ("type -P selinuxenabled || exit 0 ; "
             "selinuxenabled || exit 0 ; "
             "chcon -%st %s %s" % (flags, context, pwd))
+    # FIXME: Should use selinux.setfilecon()
     cmd = subprocess.Popen(_cmd, stdout=subprocess.PIPE,
                            stderr=subprocess.PIPE, shell=True)
     if cmd.wait() != 0:
@@ -44,12 +44,7 @@ def get_selinux_context(path):
     When selinux is enabled, return the context of ``path``
     :param path: Full or relative path to a file or directory
     :return: SELinux context as a string or None if not set
+    :raises IOError: When file does not exist.
     """
-    try:
-        # First list item is null-terminated string length
-        return selinux.getfilecon(path)[1]
-    except IOError, xcpt:
-        if xcpt.errno == errno.ENOENT:
-            return None
-        else:
-            raise
+    # First list item is null-terminated string length
+    return selinux.getfilecon(path)[1]
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -7,19 +7,32 @@ Low-level/standalone host-environment handling/checking utilities/classes/data
        in autotest!
 """
 
+import warnings
 import subprocess
 import selinux
 
 
-# FIXME: pwd is misleading, it should be 'path'
-def set_selinux_context(pwd, context=None, recursive=True):
+def set_selinux_context(path=None, context=None, recursive=True, pwd=None):
     """
     When selinux is enabled it sets the context by chcon -t ...
-    :param pwd: target directory
+    :param path: Relative/absolute path to file/directory to modify
     :param context: desired context (svirt_sandbox_file_t by default)
     :param recursive: set context recursively (-R)
+    :param pwd: target path (deprecated, was first argument name)
     :raise OSError: In case of failure
     """
+    # This is here to catch any callers using pwd as a keyword argument.
+    # Moving it to the end of the list is safe because there only ever
+    # was a single positional argument in this function (preserves API)
+    if pwd is not None:
+        warnings.warn("Use of the pwd argument is deprecated,"
+                      " use path (first positional) instead",
+                      DeprecationWarning)
+        path = pwd
+    # Catch callers that only use ketword arguments but fail to pass
+    # either ``pwd`` or ``path``.
+    if path is None:
+        raise TypeError("The path argument is required")
     if context is None:
         context = "svirt_sandbox_file_t"
     if recursive:
@@ -43,8 +56,9 @@ def get_selinux_context(path):
     """
     When selinux is enabled, return the context of ``path``
     :param path: Full or relative path to a file or directory
-    :return: SELinux context as a string or None if not set
-    :raises IOError: When file does not exist.
+    :return: SELinux context as a string
+    :raises IOError: As per usual.  Documented here as it's
+    a behavior difference from ``set_selinux_context()``.
     """
     # First list item is null-terminated string length
     return selinux.getfilecon(path)[1]
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -7,17 +7,9 @@ Low-level/standalone host-environment handling/checking utilities/classes/data
        in autotest!
 """
 
-import sys
 import errno
 import subprocess
-try:
-    # TODO: Get python-selinux module working in travis (ubuntu)
-    import selinux
-except ImportError:
-    # This always fails on travis with: no get in xattr (E0611)
-    # dispite successful 'pip install pyxattr==0.5.1'.  A test
-    # for this name has been added to .travis.yml directly.
-    from xattr import get as getxattr  # pylint: disable=E0611
+import selinux
 
 
 # FIXME: pwd is misleading, it should be 'path'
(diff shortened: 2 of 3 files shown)


dockertest/environment.py
```diff
@@ -49,8 +56,12 @@ def get_selinux_context(path):
     :return: SELinux context as a string or None if not set
     """
     try:
-        # may include null-termination in string
-        return getxattr(path, 'security.selinux').replace('\000', '')
+        if USEGETFILECON:
```
edsantiago: I thought you didn't like globals? Why not

    if 'selinux' in sys.modules:

dockertest/environment.py
```diff
@@ -49,8 +56,12 @@ def get_selinux_context(path):
...
+            # First list item is null-terminated string length
+            return selinux.getfilecon('/etc')[1]
```
edsantiago: `/etc` ?

dockertest/environment.py
cevich: global constants are fine, just global *variables* are problematic.  But, I like your solution even better.  Thanks!

dockertest/environment.py
cevich: hehe, wups, left my test in there...

edsantiago:
LGTM, but tot ake a step back... why keep the xattr stuff? The only reason for the hoop-jumping last week was to get Travis to pass, right? You don't actually need any xattr code anywhere for running this code on RHEL, Fedora, CentOS, or their Atomic variants. Why not just a simple 'try import selinux' and leave it at that?

cevich:
> Why not just a simple 'try import selinux' and leave it at that?

Because I can't be bothered to look Travis in the eye this week (ssshhhhhh, don't tell him, I secretly covered his toilet with cellophane) 😀 

I've got a *TODO* on my list to investigate making libselinux-python work on Ubuntu.  With that done, then I can safely nuke all the xattr stuff.  It's not that this is a lot of work...maybe an afternoon...it really is just that I'm still mad at Travis.  I'll get over it though.

edsantiago:
I may have misspoke: it doesn't need to run on Ubuntu! The only Ubuntu anywhere is for running pylint. The actual code doesn't run.

cevich:
> I may have misspoke: it doesn't need to run on Ubuntu! The only Ubuntu anywhere is for running pylint. The actual code doesn't run.

You're right it doesn't need to run, but...all three (pylint, pep, and inspekt) import the module, and last I looked Ubuntu still uses some "AppArmor" silliness and doesn't ship selinux bits by default.  So the module does need to load in order to be checked.  That flows downwards through the stack as well, so in order for a test that uses the ``environment`` module to be checked, it has to load and it's dependent modules must also load.  It's annoying, but that's how it works.

cevich:
Okay, I'm going to merge this piece of crap, and I've got a TODO for fixing it properly.  This is just temporary to get things un-stuck on Fedora.  In fact...maybe I should stick that in the commit message.

edsantiago:
I still don't think I'm expressing myself correctly. What I mean is: you can remove the whole xattr stuff, including the PEP and travis stuff. It doesn't need to exist.

dockertest/environment.py
```diff
@@ -7,12 +7,17 @@
...
 import errno
 import subprocess
-# This always fails on travis with: no get in xattr (E0611)
-# dispite successful 'pip install pyxattr==0.5.1'.  A test
-# for this name has been added to .travis.yml directly.
-from xattr import get as getxattr  # pylint: disable=E0611
+try:
+    # TODO: Get python-selinux module working in travis (ubuntu)
+    import selinux
+except ImportError:
```
edsantiago: IOW: I don't think you need an exception handler. You can just try to import selinux, and not care if it fails, because:

dockertest/environment.py
```diff
@@ -49,8 +54,12 @@ def get_selinux_context(path):
     :return: SELinux context as a string or None if not set
     """
     try:
-        # may include null-termination in string
-        return getxattr(path, 'security.selinux').replace('\000', '')
+        if 'selinux' in sys.modules:
+            # First list item is null-terminated string length
+            return selinux.getfilecon(path)[1]
```
edsantiago: ...because you have the selinux code in a `try`. So Ubuntu shouldn't care. I don't think you even need the conditional, you could probably just `try return selinux.etc()`. You don't need any of the `xattr` stuff, because this isn't actually running in Ubuntu. Does that make sense?

cevich:
Ugg, I didn't wanna get into this but...okay let's see what there is to get into (new commit added)...

cevich:
See, and I even forgot anotherone: Sphinx needs to import it for generating documentation.  So it has to be importable, and as you can see, it's not provided by default.

dockertest/environment.py
```diff
@@ -9,10 +9,7 @@
 
 import errno
 import subprocess
-# This always fails on travis with: no get in xattr (E0611)
-# dispite successful 'pip install pyxattr==0.5.1'.  A test
-# for this name has been added to .travis.yml directly.
-from xattr import get as getxattr  # pylint: disable=E0611
+import selinux
```
edsantiago: You lost the `try`, that's why it broke. What I was suggesting was: keep the `try`, just ditch all the xattr crap in the exception handler

edsantiago:
I'm sorry, I just don't seem to be communicating clearly today. Here's another rewrite of what I've been trying to say:

  * Ubuntu doesn't have selinux. We don't care.
  * The `try` + `import` is good. It works for both xattr and selinux.
  * But now that we've discovered `import selinux`, there's no actual need for fallback `xattr` code. You can just wrap the `import selinux` in a try, and also all code that references `selinux.foo()`, and Ubuntu won't care because it doesn't run that code. 

Sorry for all the confusion.

cevich:
> I'm sorry, I just don't seem to be communicating clearly today.

Takes two to tango...it's gotta be at least 50% my fault too 😀 

> But now that we've discovered import selinux, there's no actual need for fallback xattr code. You can just wrap the import selinux in a try, and also all code that references selinux.foo(), and Ubuntu won't care because it doesn't run that code.

OIC, so that would be **implicitly** un-supporting running Docker Autotest on Ubuntu.  I have mixed feelings on that because I have been tasked with making that work in the past, as well as some other popular distros.  Remember nearly ***all*** upstream docker development is Ubuntu-based.  Though, be that as it...was(?),  I don't know of any actual usage (today). So perhaps making it **explicitly** unsupported by documenting the fact, is the way to go.

The alternative is to spend a few hours tinkering to see about getting that python module in travis, then no documentation changes are needed.

How about a compromise: I'll set a count-down timer for two hours.  If I can't get it working in travis in that time, I'll take your approach and un-support Ubuntu.

edsantiago:
Oh yeesh. I can't imagine this working on Ubuntu, period! Paths, SELinux, rpm. Two hours seems like a lot of time to spend when there are pressing issues, such as the huge list of ADEPT failures on rhel-7-3. I'm making no progress understanding those.

cevich:
okay, how about 1 hour 😀

cevich:
Boom! Done.  19 minutes and 40 seconds 😀

cevich:
> I can't imagine this working on Ubuntu,

No reason why it can't, It is a supported platform for Autotest.  Though I've not tested it, there are probably some tweaks needed for some tests/configuration.  I've always assumed that since I was asked to make it work at one point, that day could come again.  Might as well at least leave the door cracked open.

.travis.yml
```diff
@@ -14,24 +14,26 @@ matrix:
...
 
 install:
     - pip install Pylint==1.5.5 Pep8==1.6.2 Sphinx==1.2.2
     - pip install Autotest==0.16.2
     - pip install inspektor==0.2.0
     - pip install unittest2==0.8.0
-    - pip install pyxattr==0.5.1
 
 script:
+    - python -c 'import selinux; print dir(selinux)'
```
edsantiago: This looks like a debugging holdover...?

cevich: oh right, thanks.

dockertest/environment.py
```diff
@@ -49,10 +46,10 @@ def get_selinux_context(path):
     :return: SELinux context as a string or None if not set
     """
     try:
-        # may include null-termination in string
-        return getxattr(path, 'security.selinux').replace('\000', '')
+        # First list item is null-terminated string length
+        return selinux.getfilecon(path)[1]
     except IOError, xcpt:
-        if xcpt.errno == errno.EOPNOTSUPP:
+        if xcpt.errno == errno.ENOENT:
```
edsantiago: What's this change for? IIUC the EOPNOTSUPP was something xattr-specific. Now, using the selinux module, does this exception make sense? ISTM that feeding this function a bad path (ENOENT) should not be handled cleanly, it should be raised.

cevich: libselinux raises a different exception.  The bigger thing was the previous behavior of this function when it was embedded in the test-code.  I was trying to preserve that instead of messing with the test itself too much.  Though I suppose I could try fixing the test, it's just out of scope for what I was intending here.  I'll take a look though, maybe it'll only take 19 minutes or less 😀

edsantiago: > libselinux raises a different exception.

When? What is causing that exception? In reading that, what I see is "if file doesn't exist, i.e. caller has fed us garbage, return None". That doesn't seem meaningful. Is selinux raising ENOENT on some other condition?

cevich:
Blah...while I'm into this module anyway, might as well resolve another fixme...

cevich:
@edsantiago still LGTY?  Travis is happy, internal CI was happy (not running now b/c an internal repo. is down ATM).

dockertest/environment.py
```diff
@@ -7,23 +7,30 @@
...
+    :param path: Relative/absolute path to file/directory to modify
     :param context: desired context (svirt_sandbox_file_t by default)
     :param recursive: set context recursively (-R)
+    :param pwd: target path (deprecated, was first parameter name)
     :raise OSError: In case of failure
     """
+    # This is here to catch any callers using pwd as a keyword argument.
+    # Moving it to the end of the list is safe because there only ever
+    # was a single positional argument in this function (preserves API)
+    if pwd is not None:
```
edsantiago: Good thinking!

dockertest/environment.py
```diff
@@ -47,12 +55,7 @@ def get_selinux_context(path):
     When selinux is enabled, return the context of ``path``
     :param path: Full or relative path to a file or directory
     :return: SELinux context as a string or None if not set
+    :raises IOError: When file does not exist.
```
edsantiago: Why single this one out for mentioning? We can't know what other exceptions selinux might raise; it's not really our job here to list them...

cevich: It's nice when ``get_*()`` and ``set_*()`` match behavior closely, but this is a very specific difference in this case.  So I felt it would be nice to at least document it.  Are you afraid this implies other exceptions are not raised?

cevich:
(force-push s/parameter/argument for consistency + add comment.  No code changes)

dockertest/environment.py
edsantiago: If I see `raises` in the doc string, I expect to see `raises` in the code. I also wonder if there are other circumstances in which selinux might raise IOError, not just ENOENT, making the comment misleading. This isn't a critical issue, just not something I'd like to see as a habit.

cevich: >  also wonder if there are other circumstances in which selinux might raise IOError

I'm sure there are.  Hmmm, okay, if you're wondering about this stuff, it's probable someone else will too.  That can all be addressed by docs and comments...

cevich:
...third try for the fourteenth time 😀 ...

PR #684, opened 2017-01-18
Oops. Missed one instance in arg name change

edsantiago:


cevich:
Ahh great, thanks!

PR #685, opened 2017-02-10
syslog test: try reading more syslog lines

edsantiago:
i.e. bump up seek from -4k to -8k. Also, improve the
diagnostics:

  - test failed
  - Did not find expected message 'foo' in last N lines of syslog

...so if/when we see this again the poor test engineer has
some chance of understanding what happened and then investigating.

### DIFF
--- a/subtests/docker_cli/syslog/syslog.py
+++ b/subtests/docker_cli/syslog/syslog.py
@@ -53,6 +53,10 @@ class syslog(Subtest):
                    self.stuff['fin'],
                    self.stuff['testcmds']]
 
+        # New syslog output must be after the current syslog file end
+        self.stuff['syslog_fh'] = open(self.config['syslogfile'])
+        self.stuff['syslog_fh'].seek(0, os.SEEK_END)
+
         nfdc = DockerCmd(self, "run", subargs)
         self.stuff['cmdresults'] = mustpass(nfdc.execute())
 
@@ -65,18 +69,14 @@ class syslog(Subtest):
         self.verify_message_logged()
 
     def verify_message_logged(self):
-        linecount = 0
-        with open(self.config['syslogfile']) as f:
-            f.seek(-8192, os.SEEK_END)
-            for line in f:
-                linecount += 1
-                if line.strip().endswith(self.stuff["msg"]):
-                    self.loginfo(line.strip())
-                    return
+        for line in self.stuff['syslog_fh']:
+            if line.strip().endswith(self.stuff["msg"]):
+                self.loginfo("Found in syslog: %s" % line.strip())
+                self.stuff['syslog_fh'].close()
+                return
         raise DockerTestFail("Did not find expected message '%s'"
-                             " in last %d lines of syslog file %s" %
-                             (self.stuff["msg"], linecount,
-                              self.config['syslogfile']))
+                             " in syslog file %s" %
+                             (self.stuff["msg"], self.config['syslogfile']))
 
     def cleanup(self):
         super(syslog, self).cleanup()
--- a/subtests/docker_cli/syslog/syslog.py
+++ b/subtests/docker_cli/syslog/syslog.py
@@ -21,6 +21,7 @@ Prerequisites
 import os
 import os.path
 import datetime
+from autotest.client.shared import utils
 from dockertest.subtest import Subtest
 from dockertest.output import mustpass
 from dockertest.dockercmd import DockerCmd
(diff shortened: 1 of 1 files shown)


cevich:
Fix LGTM.  My only concern here is that the OS could be logging 8k worth of crap for the duration of this test?   Geeze that seems like a lot of noise, but that's not on us.  This would be a good test to update to use that ``dockertest.output.unseen_lines.UnseenlineMatch()`` with a suitably large timeout.

cevich:
Hmmm, I do see this in the CI results for ``centosatomic.subtests.docker_cli.syslog``

(from centosatomic) Failing for the past 5 builds (Since Failed#173 )

Error Message

Did not find expected message 'SYSLOGTEST at time: 2017-02-10 17:13:52' in last 47 lines of syslog file /var/log/messages

I wonder if this problem is caused because atomic needs that [``rsyslog`` container](http://www.centosabc.com/archives/1565) running?  (I dunno, just guessing).

edsantiago:
Past 5 builds? Which ADEPT is that? ("full" only goes up to 108, not 173). My reason for submitting this PR is that on full I see this test flip-flopping between pass and fail.

Actually, on second thought, the whole seek thing is crazy. The test should instead open the syslog file before running docker. Let me see if I can figure out the pythony way to do that.

cevich:
> the whole seek thing is crazy. The test should instead open the syslog file before running docker. Let me see if I can figure out the pythony way to do that.

If it's flip-flopping, that could be due syslog buffering.  i.e. sometimes the test hits before the flush and sometimes after.  That ``unseen_lines`` module will should care of this, as long as the timeout is > the dirty-flush interval (30 or 60s by default IIRC).  Create the instance during ``initialize()``, store it and check it in ``post_process()``.  It should sit there (until timeout) and wait for the line to appear.

The centos atomic failure may be a special case.  There's an ``rsyslog`` container (see link) that's used for proper logging/journaling.  I just can't remember if it's required or not for simple cases like this.

edsantiago:
Ah, "173" must refer to `dockerautotest_github`. Yeah, the centosatomic there is hosed, failing lots of tests. I don't bother looking at it.

cevich:
e.g. I think we could do something like this (untested):

```python
from dockertest.output.unseenlines import UnseenLines, UnseenlineMatch
    ...
    def initialize(self):
        ...
        self.stuff[syslog_ul] = UnseenLines(self.config['syslogfile'], log_fn=self.logdebug)

    def run_once(self):
        ... # logger SYSLOGTEST blah
    
    def post_process(self):
        # Throws exception if no match w/in timeout.
        UnseenlineMatch(r'.*SYSLOGTEST.*', self.stuff[syslog_ul], self.config.docker_timeout)
```

edsantiago:
I've got it, thanks. Testing in progress, looking good. Will submit PR soon.

edsantiago:
I'm not using UnSeenLines because I don't understand it. I'm using stuff I understand. It seems to be working fine under `run_until_error` (i.e. it hasn't failed yet).

cevich:
> I'm not using UnSeenLines because I don't understand it.

I don't understand everything in my cars computer, yet I still drive my car 😀    In any case, as long as your solution accounts for our (by default) async-writes to syslog, it's fine by me.

What's this ``dockerautotest_github`` thing that keeps failing?

subtests/docker_cli/syslog/syslog.py
```diff
@@ -62,16 +66,17 @@ def postprocess(self):
         self.loginfo("Commands: %s" % _command)
         _status = self.stuff['cmdresults'].exit_status
         self.failif(_status, str(self.stuff['cmdresults'].stderr))
-        _check = self.verify_message_logged()
-        self.failif(not _check, "syslog test failed")
+        self.verify_message_logged()
```
cevich: ``/etc/rsyslog.conf``:

```
...
# File syncing capability is disabled by default. This feature is usually not required,
# not useful and an extreme performance hit
#$ActionFileEnableSync on
...
```

subtests/docker_cli/syslog/syslog.py
```diff
@@ -62,16 +66,17 @@ def postprocess(self):
...
 
     def verify_message_logged(self):
-        with open(self.config['syslogfile']) as f:
-            f.seek(-4096, os.SEEK_END)
-            for line in f:
-                if line.strip().endswith(self.stuff["msg"]):
-                    self.loginfo(line.strip())
-                    return True
```
cevich: ``utils.run('journalctl --flush')`` right before this loop logically negates any possibility of journald/rsyslog buffering having any effect.

edsantiago: hm, there's no "f" in the new code. I think you're looking at the older commit.

cevich:
jenkins, test this please.

cevich:
CI results are wonky.  I watched one of the centosatomic VMs fail on 'atomic host upgrade', suspecting networking/timeout issues are involved.  https://github.com/cevich/autotest-docker-enabled-product-testing/pull/107

PR #686, opened 2017-02-14
Fix pull timeout

cevich:
there's an inf. loop in ``while not dockercmd.done``

### DIFF
--- a/dockertest/dockercmd.py
+++ b/dockertest/dockercmd.py
@@ -11,6 +11,7 @@ from autotest.client import utils
 from subtestbase import SubBase
 from xceptions import DockerNotImplementedError
 from xceptions import DockerExecError, DockerTestError
+from xceptions import DockerCommandError
 
 
 class DockerCmdBase(object):
@@ -398,14 +399,20 @@ class AsyncDockerCmd(DockerCmdBase):
     @property
     def done(self):
         """
-        Return True if processes has ended
+        Return True if processes has ended (successful or not)
 
-        :raises DockerTestError: on incorrect usage
+        :raises DockerTestError: execute() was not called first
+        :raises DockerCommandError: If timeout exceeded
         """
 
         if self._async_job is None:
             raise DockerTestError("Attempted to wait for done before execute()"
                                   " called.")
+        if self.duration >= self.timeout:
+            # Exception takes care of logging the command
+            raise DockerCommandError("Timed out after %0.2f seconds"
+                                     % (float(self.timeout)),
+                                     self.cmdresult)
         return self._async_job.sp.poll() is not None
 
     @property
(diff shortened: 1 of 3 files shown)


dockertest/dockercmd.py
```diff
@@ -398,15 +398,21 @@ def wait(self, timeout=None):
...
 
         if self._async_job is None:
             raise DockerTestError("Attempted to wait for done before execute()"
                                   " called.")
-        return self._async_job.sp.poll() is not None
+        has_exited = self._async_job.sp.poll() is not None
+        timed_out = self.duration >= self.timeout
+        # Don't assume this condition is logged elsewhere
+        if timed_out:
+            self.subtest.logwarning("Async. docker command has timed out"
```
edsantiago: Shouldn't a timeout raise an exception, or otherwise be treated as an error?

cevich: That was my original implementation, however I'm afraid it alters the API behavior too drastically.  Also because I'm lazy and didn't want to go chasing down every test that waits on done, and wrap it up in a ``try...except``.

cevich: ...though.  I wonder if not raising an exception could cause a false-positive situation...hmmmm, maybe you're right.

cevich:
(I aborted first CI run)

dockertest/dockercmd.py
```diff
@@ -398,21 +399,20 @@ def wait(self, timeout=None):
...
                                   " called.")
-        has_exited = self._async_job.sp.poll() is not None
-        timed_out = self.duration >= self.timeout
-        # Don't assume this condition is logged elsewhere
-        if timed_out:
-            self.subtest.logwarning("Async. docker command has timed out"
-                                    " after %0.2f", float(self.timeout))
-        return has_exited or timed_out
+        if self.duration >= self.timeout:
+            raise DockerCommandError("Async. docker command has timed out"
```
edsantiago: Future debuggers might appreciate a `str(self)` indicating what the command was...

cevich: https://github.com/cevich/autotest-docker/blob/59c8c615d21aca43f1c0bc78fdab71d80091d24c/dockertest/dockercmd.py#L333

An exception is raised if ``execute()`` wasn't called before ``wait``/``done``.

edsantiago: I'm not sure what I'm looking at. To me it looks like you highlight a `logdebug` wrapped inside `if verbose`. What I was suggesting is that the error message in the above exception might be more helpful if it included the actual docker command that timed out. That seems like it would help a future person investigating a failure.

cevich: aye,  This property requires calling ``execute`` first where if verbosity is not important it's explicitly set false.  Otherwise the full details are logged for debugging.  In other words, if there's a timeout error, the command is w/in the debug log, just scroll up.

It's a small change though, and I need to fixup the unittests anyway...

cevich:
(waiting on CI run then I'l --force push)

cevich:
Silly me, totally forgot that exception does the logging on its own.

PR #687, opened 2017-02-16
Add intratest to log recent AVC audit messages

cevich:
super-duper quick-and-dirty so we can get info. ASAP from CI.

### DIFF
--- a/control
+++ b/control
@@ -691,6 +691,14 @@ class StepInit(Context, collections.Callable):
         return [subthing for subthing in subthings
                 if subthing in subtest_modules]
 
+# Grab a compressed copy of the audit log after every test
+# N/B: Autotest silently ignores any failures with this.
+audit_log = '/var/log/audit/audit.log'
+if os.path.isfile(audit_log):
+    command = 'gzip -9c %s' % audit_log
+    job.add_sysinfo_command(command, "audit.log.gz", True)
+else:
+    logging.warning("Can't record %s because it doesn't exist", audit_log)
 
 # Entry point into step-engine, job searches for this callable
 step_init = StepInit()


cevich:
> LGTM. All files generated on my system are empty;

Now that may be a problem.  Investigating...

>  Also, this runs after every garbage_check as well as every test. I assume that, too, is expected.

Unexpected, but understandable.  Yeah, that's maybe just a down-side to live with.

cevich:
Files all empty in CI run as well.  This is one of the down-sides to using this facility, it totally ignores any errors and doesn't tell you what went wrong.  My best guess here is that the command isn't available on atomic.  Let's try with it just grabbing the audit log file itself...

cevich:
confirmed: There's no ``ausearch`` available on Atomic.

cevich:
uggg, so ``add_sysinfo_logfile`` doesn't seem to do the business, I get nothing out the other end and no debugging assistance either 😖    I think this might be why I abandoned that other PR.  We may need to go with the intra-test approach for sanity's sake.  Maybe ``results2junit`` can just filter out all the intratests which aren't ``garbage_check``.  Not sure how to proceed (and there's a big list of other commands/data that would be nice to have).

edsantiago:
>so add_sysinfo_logfile doesn't seem to do the business

Odd. It's working for me (RHEL test; still ongoing). What symptom are you seeing?

edsantiago:
FWIW this revision seems to be working as well. Actually, exactly as well as the previous one (without the `file=`). I see lots of `audit.log` files, all of them growing in size with each subtest. I don't understand the purpose of adding `file=` but will not spend any time investigating.

cevich:
> FWIW this revision seems to be working as well

Huh? What? Really?  WTH?  I looked at the CI results and saw exactly ---zero--- audit.log files.

> but will not spend any time investigating.

``file`` is a python builtin (alias for ``open()`` basically), somebody decided to use it as a parameter name in the ``add_sysinfo_logfile()`` method 😢   Since I didn't see any ``audit.log`` files, I thought maybe it was because of that (the autotest method is a wrapper around  ``something(file, other_param)``.

I'll look again, if it is working as you say, I'll kill that last commit of mine.  Thanks for saying something...

cevich:
OIC...it's there for ``centos`` but totally missing for ``centosatomic``.  I must have only looked at the atomic results previously.  Huh.  Anyway, I guess solving this for atomic can be left up to a future PR.

cevich:
Okay, this seems to be working except for centosatomic.  The place where we need this urgently is not an atomic host anyway, so  I'm willing to live with that for now.  @edsantiago okay if I merge?

control
```diff
@@ -691,6 +691,14 @@ class StepInit(Context, collections.Callable):
         return [subthing for subthing in subthings
                 if subthing in subtest_modules]
 
+# Grab a compressed copy of the audit log after every test
+# N/B: Autotest silently ignores any failures with this.
+audit_log = '/var/log/audit/audit.log'
+if os.path.isfile(audit_log):
+    command = 'gzip -9c %s' % audit_log
```
cevich: Note: Compressing these goes from ~1MB per test down to ~60KB, so it's a significant reduction (times hundreds / thousands of tests).

control
```diff
@@ -691,6 +691,14 @@ class StepInit(Context, collections.Callable):
...
+    job.add_sysinfo_command(command, "audit.log.gz", True)
```
edsantiago: Shouldn't that be audit_log.gz (underscore, not dot)?

edsantiago: no, never mind.

edsantiago:
compressing seems prudent, those are big files. It seems to be working on my test setup. Run still in progress, but I see no reason not to merge.

cevich:
@edsantiago thanks for the double-check.  Merging.

PR #688, opened 2017-02-21
Workaround RH BZ 1393572 & docker_cli/negativeusage/iv4

cevich:
I realize this test failure is subject of some debate.  However that's orthagonal to getting some useful runtime out of it, while we wait (maybe indefinitely) for a bug fix.   The original reproducer and a comment about the BZ is preserved in the configuration.

### DIFF
--- a/config_defaults/subtests/docker_cli/negativeusage.ini
+++ b/config_defaults/subtests/docker_cli/negativeusage.ini
@@ -64,11 +64,13 @@ subcmd = run
 subarg = -p,192.168.9.1:9000,%%(FQIN)s,/bin/true
 stderr = Invalid hostPort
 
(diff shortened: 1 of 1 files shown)


cevich:
jenkins, test this please.

edsantiago:
This BZ was pretty much *THE* test case for my ignore-known-failures code.

cevich:
> This BZ was pretty much THE test case for my ignore-known-failures code.

Making tests fail is easy, especially on purpose 😀 

Maybe just toss a **TESTING: DO NOT MERGE** commit into whichever PR which causes some breakage.  There are many other options too, ping me if/when/where needed and we'll work something out.

PR #692, opened 2017-03-15
docker 1.13 changes a word in an error message

edsantiago:
in docker logout:

  - Remove login credentials for
  + Removing login credentials for

Deal with both.


cevich:
Fix LGTM, seems like CI is borked (or docker.io registry is not reachable), please go ahead and merge assuming travis is happy.

cevich:
Hmmm, seems travis is also borked.  It's using docker containers also, so maybe ``docker.io`` has some outage that's affecting it and (internal) CI.

PR #693, opened 2017-03-15
Improve run volumes selinux

cevich:
The main change here is the commit ``run_volumes_selinux: Seperate + Improve start detection``.  The rest are mostly fluff.  For whatever reason, I wasn't able to install pylint to check any of this, so I'm relying on CI to do that.

### DIFF
--- a/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
+++ b/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
@@ -98,12 +98,10 @@ class selinux_base(subtest.SubSubtest):
         """ Check all files in given $path have the context $context """
         for pwd, _, filenames in os.walk(path):
             for filename in filenames:
-                act = get_selinux_context("%s/%s" % (pwd, filename))
-                self.logdebug("File %s Context: %s",
-                              os.path.join(pwd, filename),
-                              act)
-                self.failif_ne(act, context, "Context of file %s/%s"
-                               % (pwd, filename))
+                full_filepath = os.path.join(pwd, filename)
+                actual = get_selinux_context(full_filepath)
+                self.failif_ne(actual, context,
+                               "SELinux context of file %s" % full_filepath)
 
     def init_volume(self, context):
         """
--- a/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
+++ b/subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
@@ -38,6 +38,8 @@ class selinux_base(subtest.SubSubtest):
 
     """ Base class """
 
+    OCTAL_READY = r'\0122\0105\0101\0104\0131'
+
     def initialize(self):
         """
         Runs one container
@@ -62,7 +64,8 @@ class selinux_base(subtest.SubSubtest):
             subargs.append("--volume %s" % volume)
         if volumes_from:
             subargs.append("--volumes-from %s" % volumes_from)
-        name = self.sub_stuff['dc'].get_unique_name()
+        dc = self.sub_stuff['dc']
+        name = dc.get_unique_name()
         self.sub_stuff['containers'].append(name)
         subargs.append("--name %s" % name)
         fin = DockerImage.full_name_from_defaults(self.config)
@@ -72,15 +75,23 @@ class selinux_base(subtest.SubSubtest):
         self.sub_stuff['fds'].append(write_fd)
         self.sub_stuff['fds'].append(read_fd)
         dkrcmd = dockercmd.AsyncDockerCmd(self, 'run', subargs)
-        # TODO: Fix use of dkrcmd.stdin when appropriate class mech. available
         dkrcmd.execute(read_fd)
         dkrcmd.stdin = write_fd
         os.close(read_fd)  # no longer needed
-        os.write(write_fd, 'echo "Started"\n')
-        self.failif(not wait_for_output(lambda: dkrcmd.stdout,
-                                        'Started',
-                                        timeout=20),
-                    "Error starting container %s: %s" % (name, dkrcmd))
+
+        # This could take a while if image needs pulling
+        find_name = lambda: str(dc.list_containers_with_name(name)) != []
+        self.failif_ne(utils.wait_for(find_name,
+                                      timeout=self.config['docker_timeout']),
+                       True,  # None == timeout
+                       "Container %s not in 'docker ps' output within"
+                       " timeout: %s" % (name, dc.list_containers()))
+
+        # Don't match echo of the 'echo' command itself
+        self.logdebug('Confirming %s is responding', name)
+        os.write(write_fd, "echo -e '%s'\n" % self.OCTAL_READY)
+        dkrcmd.wait_for_ready()
+        self.logdebug('Confirmed')
         return dkrcmd
 
     def check_context_recursive(self, path, context):
--- a/config_defaults/subtests/docker_cli/run_volumes_selinux.ini
+++ b/config_defaults/subtests/docker_cli/run_volumes_selinux.ini
@@ -5,6 +5,6 @@ run_options_csv = -i
 #: Set yes to force use of system temp directory.  This may be
 #: necessary if selinux is unable to relabel files in the autotest
 #: temporary directory.
-use_system_tmp = yes
+use_system_tmp = no
 #: SELinux context type to set on volumes source directory
 selinux_host = public_content_rw_t
--- a/control
+++ b/control
@@ -700,7 +700,8 @@ class StepInit(Context, collections.Callable):
 audit_log = '/var/log/audit/audit.log'
 if os.path.isfile(audit_log):
     command = 'gzip -9c %s' % audit_log
-    job.add_sysinfo_command(command, "audit.log.gz", True)
+    # on_every_test=False has already run (before control file loaded)
+    job.add_sysinfo_command(command, "audit.log.gz", on_every_test=True)
 else:
     logging.warning("Can't record %s because it doesn't exist", audit_log)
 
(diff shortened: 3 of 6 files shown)


cevich:
[(3) Fix docker_cli.run_volumes_selinux system integration test](https://trello.com/c/HfliW6Fb/438-3-fix-docker-cli-run-volumes-selinux-system-integration-test)

cevich:
Something is terribly wrong with the CentOS-latest VMs setup (probably in ADEPT).  Their setup is almost identical to what's used for RHEL.  But the /etc/sysconfig/docker* files simply don't look right.  Need to investigate deeper so these results don't predict future RHEL results.

cevich:
(waiting to merge travis fixes until investigation into possible ADEPT VM provisioning/setup problems addressed https://github.com/cevich/autotest-docker-enabled-product-testing/pull/122)

cevich:
Closed other PR - no change in broken CentOS behavior.  I think our energies are better spent focusing on RHEL and Fedora for Docker Autotest.  I'll take up that change in another PR.

While this PR doesn't fix or reveal the problem I was hoping it would, it's still worth pursuing a merge.  The commits here do reveal a lot more details than were available before.

cevich:
Example of what this change looks like (internal CI):
Before: https://hurl.corp.redhat.com/290cb22
After: https://hurl.corp.redhat.com/a72ddde

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -72,27 +75,32 @@ def init_container(self, volume=None, volumes_from=None):
...
             for filename in filenames:
-                act = get_selinux_context("%s/%s" % (pwd, filename))
-                self.logdebug("File %s Context: %s",
-                              os.path.join(pwd, filename),
-                              act)
-                self.failif_ne(act, context, "Context of file %s/%s"
-                               % (pwd, filename))
+                actual = get_selinux_context("%s/%s" % (pwd, filename))
+                msg = ("File %s context: %s expecting %s"
+                       % (os.path.join(pwd, filename), actual, context))
```
edsantiago: You don't need to include actual/expected: `failif_ne` does that automatically. You can just say "selinux context of <filepath>"

cevich: Hmm, do we even need to log here at all then?  If we get automatic details out of ``failif_ne`` on breakage, can't the test just be quiet if everything is okay?

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -72,27 +75,32 @@ def init_container(self, volume=None, volumes_from=None):
...
-        os.write(write_fd, 'echo "Started"\n')
-        self.failif(not wait_for_output(lambda: dkrcmd.stdout,
-                                        'Started',
-                                        timeout=20),
-                    "Error starting container %s: %s" % (name, dkrcmd))
+
+        # This could take a while if image needs pulling
+        find_name = lambda: str(dc.list_containers_with_name(name))
+        is_running = wait_for_output(find_name, name,
+                                     timeout=self.config['docker_timeout'])
```
edsantiago: I'm having a tough time with this. First, understanding it (but I finally did). Then, making sure it does what it's supposed to (which I still haven't confirmed). The unnecessary duplicate use of `name` troubles me. Is there any reason not to just use

    find_name = lambda: dc.list_containers_with_name(name) != []
    utils.wait_for(find_name, timeout=...) ?

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
edsantiago: I'm confused. Where do you see logging? I see logging removed.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
cevich: Ahh, okay, yes, that would be a simpler way.  In fact, that could be folded into ``failif_ne()`` to be even simpler...

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
cevich: oic, nvm.

cevich:
@edsantiago okay, both items fixed.

cevich:
--force push again, commit spelling fix.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -72,27 +75,33 @@ def init_container(self, volume=None, volumes_from=None):
...
         os.close(read_fd)  # no longer needed
-        os.write(write_fd, 'echo "Started"\n')
-        self.failif(not wait_for_output(lambda: dkrcmd.stdout,
-                                        'Started',
-                                        timeout=20),
-                    "Error starting container %s: %s" % (name, dkrcmd))
+
+        # This could take a while if image needs pulling
+        find_name = lambda: str(dc.list_containers_with_name(name)) != []
+        self.failif_ne(wait_for_output(find_name,
```
edsantiago: Now I'm really confused. How does `wait_for_output` work without a `pattern` arg?

cevich: Nope, the confusion is all mine.  I intended to use your suggestion of ``utils.wait_for()`` but a bird suddenly flew past the window.  Out of the corner of my eye, I swore it had my toothbrush in it's beak.  Unfortunately, between getting up to check, and sitting back down, my cat attacked me.  Then while getting a new pair of pants, I found the pack of dental-floss that went missing last week.   Unfortunately, justice had already (mistakenly) been delivered to that particular clepto-bird.  RIP.

cevich:
rebase and --force-push, ``wait_for_output()`` -> ``utils.wait_for()``

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -72,27 +75,33 @@ def init_container(self, volume=None, volumes_from=None):
...
         for pwd, _, filenames in os.walk(path):
             for filename in filenames:
-                act = get_selinux_context("%s/%s" % (pwd, filename))
-                self.logdebug("File %s Context: %s",
-                              os.path.join(pwd, filename),
-                              act)
-                self.failif_ne(act, context, "Context of file %s/%s"
-                               % (pwd, filename))
+                actual = get_selinux_context("%s/%s" % (pwd, filename))
+                msg = ("File %s context: %s"
```
edsantiago: Why the context? Again, `failif_ne` gives a helpful message. Duplicating it is not necessary.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
```diff
@@ -72,27 +75,33 @@ def init_container(self, volume=None, volumes_from=None):
...
+        find_name = lambda: str(dc.list_containers_with_name(name)) != []
+        self.failif_ne(utils.wait_for(find_name,
+                                      timeout=self.config['docker_timeout']),
+                       True,  # None == timeout
+                       "Container %s not in 'docker ps' output within"
+                       " timeout: %s" % (name, dc.list_containers()))
+
+        # Don't match echo of the 'echo' command itself
+        self.logdebug('Confirming %s is responding', name)
+        os.write(write_fd, "echo -e '%s'\n" % self.OCTAL_READY)
```
edsantiago: Out of curiosity, why the convolutions? Why not just echo READY?

cevich:
Note: ADEPT CI passed 100% since I updated the centos image to 7.3.  The two 'atomic' failures were due to what looks like networking problems during the 'atomic host upgrade' process.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
cevich: Uggg, sry, I'm missing something then.  Don't I have to pass it a ``msg`` parameter?  Are you suggesting I just leave that blank?

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
cevich: Awww, didn't notice my nice commit message answering that?

```
Finally, the shell being
run under docker, has command-echo turned on.  Therefor it's possible
to get a false match, searching output for 'READY' after sending an
'echo READY' as input.  (i.e. it could match feedback of the echo command
itself instead of it's output.)
```

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
edsantiago: `failif_ne` already says "expected THIS, got THAT". All you need to say is *WHAT* the thingy is, and in this case that should be "SELinux context for [PATH]". See `dockertest/subtestbase.py:failif_ne()`

cevich: Oooohhh that part.  Thanks, sorry, failed my Cptn. Obvious. Test today...twice... 😢

cevich:
(I aborted CI run)  --force push w/ rebase.  Should be all set now, oh! look, another bird!  Hey! That's my wallet!  Oh wait, nope, just a leaf.

subtests/docker_cli/run_volumes_selinux/run_volumes_selinux.py
edsantiago: Ah, got it. And that explains the comment as well. Thanks.

cevich:
--force push, again, for the forth time.  Pre-approved via IRC 😀

PR #695, opened 2017-03-17
Fixup ignore known failures

cevich:
Based on #679, remove all "experimental", apply some minor polish, add some basic unittests.

### DIFF
--- a/dockertest/subtestbase.py
+++ b/dockertest/subtestbase.py
@@ -7,11 +7,34 @@ modules.
 # Pylint runs from a different directory, it's fine to import this way
 # pylint: disable=W0403
 
+import os.path
+import csv
 import logging
 import traceback
+from environment import docker_rpm
 from xceptions import DockerTestFail
 from xceptions import DockerTestNAError
-from config import get_as_list
+from config import get_as_list, CONFIGCUSTOMS
+
+
+def known_failures():
+    """
+    Returns a dict containing known test failures. Primary key is
+    subtest name (e.g. docker_cli/foo/bar), value is another dict
+    whose key is docker NVRA (e.g. docker-1.12.5-8.el7.x86_64),
+    value of that is a string description of the problem (e.g.
+    a bz number and comment).
+    """
+    known = {}
+    path_known = os.path.join(CONFIGCUSTOMS, 'known_failures.csv')
+    with open(path_known, 'rb') as csv_fh:
+        csv_reader = csv.reader(csv_fh)
+        for row in csv_reader:
+            if row[1] not in known:
+                known[row[1]] = {}
+            # Each row is: NVR, subtest, description
+            known[row[1]][row[0]] = row[2]
+    return known
 
 
 class SubBase(object):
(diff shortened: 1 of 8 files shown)


dockertest/subtestbase.py
```diff
@@ -100,6 +124,56 @@ def log_step_msg(self, stepname):
...
+        Do we have a registered known failure in this subtest/sub-subtest
+        when running on the currently-installed docker? Return True if so.
+
+        :param subthingname: Subtest/sub-subtest name to search for
+        :returns: True/False if a match to active docker NVRA found
+        :SideEffect: log warning messages to help human debuggers.
+        """
+        if self.__class__.__name__ in ['Subtest', 'SubSubtest']:
+            raise DockerTestError('%s.is_known_failure() not yet supported',
+                                  self.__class__.__name__)
```
edsantiago: I don't get this. I've tried. What purpose does it serve?

cevich: This is here just in case somebody decides to try and call ``is_known_failure()`` from those two classes.  Basically an goof-plug, until we can support using it for subtests.  Though...now that you point it out, maybe it would be useful for some reason, to call this from w/in a test.  I can kill it, not married to it at all.

dockertest/subtestbase_unittests.py
```diff
@@ -98,41 +85,87 @@ def test_failif_redhat(self):
...
+            self.tmpfile.write(contents)
+            self.tmpfile.seek(0,0)
+
+    def test_empty_known_failures(self):
+        self.reset_tmpfile()
+        known = self.subtestbase.known_failures()
+        self.assertEqual(known, {})
+
+    def test_known_failures(self):
+        self.reset_tmpfile('"one","two","three"\n')
```
cevich: ya, let's do that.   it could use some more tests as well, but that was the best I could do at the time.

dockertest/subtestbase_unittests.py
```diff
@@ -98,41 +85,87 @@ def test_failif_redhat(self):
...
+        self.fake_open = lambda *args, **dargs: self.tmpfile
+        self.subtestbase.open = self.fake_open
+
+        self.docker_rpm = 'docker'
+        self.subtestbase.docker_rpm = lambda: self.docker_rpm
+
+    def tearDown(self):
+        # Don't let mocks or state leak anywhere
+        del self.subtestbase.open
+        del self.subtestbase.docker_rpm
```
cevich: hrmmmm 🤷‍

dockertest/subtestbase_unittests.py
```diff
@@ -48,44 +48,31 @@ class DockerTestFail(Exception):
...
         self.stbsb = self.subtestbase.SubBase
+
+
+class TestNotRedHat(SubtestBaseTestBase):
+
+    def setUp(self):
+        super(TestNotRedHat, self).setUp()
         pfx = 'subtestbase_unitest'
         (fd,
          self.stbsb.redhat_release_filepath) = mkstemp(prefix=pfx)
```
cevich: can do.

dockertest/subtestbase_unittests.py
```diff
@@ -48,44 +48,31 @@ class DockerTestFail(Exception):
 # END   boilerplate crap needed for subtests, which should be refactored
```
cevich: I did, and they all passed...unless maybe I screwed something up after and forgot to run them again.

edsantiago: I get `ImportError: cannot import name DockerTestError` and so does Travis

cevich:
> Thanks for your patience. It looks mostly fine, but the big issue still remains: there's no way to get this to work with Subtests

Thx for the review, PR has been around a while anyway and I'm not in a hurry.  

I was really hung-up on the whole "POC" and "Experimental" labels and our misunderstanding on that.  Now I'm changing my position, because I'm reminded that none of this is permanent.  I think it's a good 80:80 solution now, and we should try to get some benefit from it.   Also, I'm pretty sure we can get it working for ``subtests`` by adding some clever exception catching like this:

```python
class Subtest(subtestbase.SubBase, test.test):
    ...
    def execute(self, *args, **dargs):
        """**Do not override**, needed to pull data from super class"""
        try:
            super(Subtest, self).execute(iterations=self.iterations,
                                         *args, **dargs)
        except ...
            if self.config.get('subsubtests'):
                raise
            else:
                # Do some known-failure checks
```

I'll come back and carry on here, probably starting next week.  Though if you have a mind for it, feel free to cherry-pick into your PR, or add to this one, or whatever.  No reason to wait on my lagging-posterior 😀

dockertest/subtestbase_unittests.py
cevich: Dang, I wonder if maybe I lost some code somewhere or forgot to push...

cevich: ...oops:
```
# git st
# On branch fix_run_volumes_selinux
# Your branch is ahead of 'origin/fix_run_volumes_selinux' by 4 commits.
```

unitests should work now.

PR #696, opened 2017-03-20
Switch default testing to RHEL from CentOS

cevich:
Some places, the CentOS image is still referenced because for those
specific tests, it's better to use a non-default image.  Also, in most
places, updated the CentOS tag from 7 -> latest.  Mainly for cosmetic
purposes, though it does squah any confusion that perhaps the 7-GA
version was being used.

Note: CI testing will most likely not be useful on this PR since it's setup for CentOS only (currently).


cevich:
Note: Needs coordination w/ merges in ADEPT and updates to internal dockerautotest-CI job.

PR #697, opened 2017-03-29
login test: strict SSL cert check on connect

edsantiago:
The python urllib2 currently in RHEL does not do strict
SSL cert checks. This lets us get away with sloppiness
while waiting for registry container to spin up.

Python 2.7.9 fixes urllib2 so it does strict checks.
This breaks our _wait_for_registry_ready() code because
we never make it to the expected 'Unauthorized' error.
The urlopen() fails before then.

Solution: we have a custom certificate, that's what
our registry container uses. Just import it into the
urllib2 connection.

Also: add debug messages to make it easier to track
future problems.


cevich:
Cool! I bet that was a PITA to find.

The CI job was 100% pass, except for two atomic-host systems which failed to 'atomic host update' - timeout error.  The timeout is set to 5-minutes which should be plenty long enough IMHO.  

LGTM except for one l'il pylint finding needing fixing.

I'm curious to see if after fixing that and force-pushing if we can get a green-check outta the CI job 😀

edsantiago:
>one l'il pylint finding needing fixing.

Oh, YUK! I never even looked at Travis, just assumed it would pass. Thanks for letting me know about it.

AFAICT the urllib2 on RHEL 7.3 comes standard with python 2.7.5. Should the `python` line in `.travis.yml` require 2.7.5 ?

cevich:
> AFAICT the urllib2 on RHEL 7.3 comes standard with python 2.7.5. Should the python line in .travis.yml require 2.7.5 ?

Oh yes, good catch....mmmm....I vaguely remember trying this a long time ago and it didn't work, but I could be wrong...docs say:

"""Travis CI supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5, 3.6 as well as recent development versions."""

https://docs.travis-ci.com/user/languages/python/

edsantiago:
That's not exactly helpful (the travis docs, that is). "2.7" covers a wide ground. Maybe we need to require a custom python-urllib? I'll play with it; expect a flurry of fixup commits soon

cevich:
For sure, please feel free to play with it.  They're using a python virtualenv, and I know it's possible to use a specific python version with that.  The usual way is:

```
$ virtualenv -p /usr/bin/python2.7 <some_directory>
```

Just not sure how to tell travis to do that (and it may not be possible).

cevich:
errrr...wait, that's just 2.7 also.  Mmmmm, maybe it's not possible.  Anyway, I'm curious if you can get it working.

edsantiago:
@cevich PTAL; I've squashed and force-pushed all my clutter. This now passes Travis, it's just not terribly lovely.

.travis.yml
```diff
@@ -28,6 +24,7 @@ install:
     - pip install Autotest==0.16.2
     - pip install inspektor==0.2.0
     - pip install unittest2==0.8.0
+    - export PYTHONPATH="$PYTHONPATH:/usr/lib/python2.7/dist-packages"
```
cevich: Why is this needed?  (maybe add a comment)

cevich:
Otherwise, LGTM.

PR #698, opened 2017-04-04
docker_daemon.pid() : handle dockerd-as-container

edsantiago:
When running dockerd as a system container, the PID recorded
by systemd is that of runc, not dockerd. The actual dockerd
is a direct child of runc. We want docker_daemon.pid() to
return the dockerd PID.


dockertest/docker_daemon.py
```diff
@@ -159,16 +159,32 @@ def pid():
...
-
-
-def cmdline():
+    mainpid = int(stdout[8:])
+    cmd = cmdline(mainpid)
+    if 'dockerd' in cmd[0]:
+        return mainpid
+    # As of April 2017 we may be running dockerd under runc. If so, actual
+    # dockerd process is an immediate child of the one returned by systemd.
+    for child_pid in utils.run("pgrep -P %s" % mainpid).stdout.split():
```
cevich: How about: ``for child_pid in utils.get_children(mainpid)``?  (I dunno if ``pgrep`` exists on atomic, but ``ps`` certainly does).

Also, maybe it's overkill, but maybe consider ``if not utils.pid_is_alive(mainpid): raise something()``?

dockertest/docker_daemon.py
```diff
@@ -159,16 +159,32 @@ def pid():
...
+        cmd = cmdline(child_pid)
+        if 'dockerd' in cmd[0]:
+            return int(child_pid)
+    # Urp. No dockerd process found. Cross fingers & hope systemd is right.
+    # FIXME: we should warn, but we have no self with which to logwarning()
```
cevich: Fix:  No ``self`` needed, you can just use ``logging.warning()`` directly in this context.  The only reason we have special logging methods on ``SubtestBase`` is to make the messages look pretty.

dockertest/docker_daemon.py
edsantiago: Eek. `utils.get_children()` includes threads, which I'm pretty sure is not the desired behavior. `pgrep` exists on RHEL Atomic 7.3, which is where I'm doing all this docker-as-container testing.

Under what circumstance will systemd return a PID that is not alive? I put in the 'Unexpected output' check above because I can envision systemd returning an empty string, or some sort of warning, but I can't imagine it returning MainPID=123 where 123 is invalid. What situation are you envisioning?

dockertest/docker_daemon.py
edsantiago: Thanks! Fix added and tested; will push within the minute.

dockertest/docker_daemon.py
cevich: A process can go zombie at any time, it's not going to wait for your code between getting the pid and doing something with it.  It's extremely unlikely though, so don't worry about it.

OIC, looks like both ``ps`` and ``pgrep`` come from the ``procps-ng`` package, so no worries.

PR #699, opened 2017-04-05
events: accept 'Z' for time zone

edsantiago:
Seen on RHELAH 7.3 running docker as system container:

   2017-04-04T20:51:00.104731909Z container start f04f4...
                         -------^

Up until now I had only seen numeric offsets, but Z is
perfectly valid in ISO8601. Accept it.


subtests/docker_cli/events/test_event_parsing.py
```diff
@@ -63,7 +63,7 @@ def test_1_10(self):
              "2016-04-06T09:54:07.108386649-04:00",
              "2016-04-06T09:54:07.442663135-04:00",
              "2016-04-06T09:54:07.846336619-04:00",
-             "2016-04-06T09:55:00.000000000-04:00"]
+             "2016-04-06T13:55:00.000000000Z"]
         cid = ("ae5f8ee3cdc14512fbc6f908f06cc859"
                "0b53d0eb9e36cf543165d130f1169a0e")
```
cevich: I know it's out of scope, but while you're changing this file, should the ``cid`` test include an item for ``sha256:blah``?

cevich:
Awww, finally they fixed that. I double-checked ``dockertest/output/dockertime.py`` as well, and it seems to already support an optional ``Z```.   Changes LGTM.

subtests/docker_cli/events/test_event_parsing.py
edsantiago: It already does, see `test_1_10_fqin`

cevich: Oh, sry, my bad then.  This is Good2Go then.

PR #702, opened 2017-04-11
run_flags: kernel 3.10.0-632 supports deferred_deletion

edsantiago:
Kernel 3.10.0-632.el7 (RHEL 7.4 beta) supports the desired
deferred_deletion option. We've been waiting for this.
Update the 3.999 placeholder and remove the FIXME. We now
expect docker-storage-setup to identify this feature and
enable it.

Also, add a debug log so investigators can see the full
docker command line. Possibly useful if test virt has
been wiped.

### DIFF
--- a/subtests/docker_cli/run_flags/run_flags.py
+++ b/subtests/docker_cli/run_flags/run_flags.py
@@ -26,6 +26,7 @@ class run_flags_base(subtest.SubSubtest):
 
     def has_storage_opt(self, wanted):
         docker_cmdline = dockertest.docker_daemon.cmdline()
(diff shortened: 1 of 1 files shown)


edsantiago:
Double-confirmed by editing `/etc/sysconfig/docker-storage`, removing `--storage-opt dm.use_deferred_deletion=true` from DOCKER_STORAGE_OPTIONS, restarting docker, and running test. It fails as expected with "missing option: dm.use_deferred_deletion"

cevich:
Awesome, and thanks for remembering/following up on that fixme.  Patch LGTM, waiting for CI run to complete just as a sanity-check then we can merge.

PR #703, opened 2017-04-11
docker_cli/info: Accept multiple storage driver names

cevich:
In addition to ``devicemapper``, newer versions of docker may use
``overlay2`` (or other storage drivers).  Instead of validating a single
hard-coded value, use a configurable list of values that's easy to
maintain.

N/B: This is an incomplete update to this test.  There are other info. fields
which also need updating. e.g.:

```
Storage Driver: overlay2
 Backing Filesystem: xfs
Logging Driver: journald
Cgroup Driver: systemd
```

### DIFF
--- a/config_defaults/subtests/docker_cli/info.ini
+++ b/config_defaults/subtests/docker_cli/info.ini
@@ -10,3 +10,5 @@ data_total_error = 0.25
 meta_error = 1.00
 #: Docker reported size may vary +/- by this percent from on-disk sizes
 meta_total_error = 0.25
+#: List of acceptable storage drivers, delimited by ``|`` characters
+storage_drivers = devicemapper|overlay2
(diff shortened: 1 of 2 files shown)


edsantiago:
Yep, confirmed. This test needs a lot more work to be able to handle overlay2.

cevich:
Ed, I've got a run already done on this for runcom (basis for this PR) here: [Red Hat internal link](https://hurl.corp.redhat.com/a77d4df)

cevich:
> This test needs a lot more work

Oh okay.  Well, we can call it "a start" then 😀 

I'll relabel it and update the commit message indicating that it's incomplete.

edsantiago:
In case you want to tackle it at some point, this might help:

    # docker info|grep -A3 ^Stor
    Storage Driver: overlay2
     Backing Filesystem: xfs
    Logging Driver: journald
    Cgroup Driver: systemd

PR #705, opened 2017-04-12
info subtest: handle overlay2 driver

edsantiago:
Up until now, the 'docker info' test has only had
to deal with devicemapper. Now there's overlay2,
and the docker info output for it is significantly
different. Much of the devicemapper validation is
not applicable to overlay2.

Split the storage-handler validation into individual
functions for each driver: currently devicemapper
and overlay2, each of which has different tunables.
As more back ends are added, we can add new functions
to check their settings.

Also: rewrite _build_table() so it handles sub-elements,
and add a unit test.

### DIFF
--- a/subtests/docker_cli/info/info.py
+++ b/subtests/docker_cli/info/info.py
@@ -28,6 +28,7 @@ from dockertest.output import OutputGood
 from dockertest.dockercmd import DockerCmd
 from dockertest.output import mustpass
 from dockertest.images import DockerImages
+from dockertest.xceptions import DockerTestFail
 
 
 class info(subtest.Subtest):
@@ -40,38 +41,81 @@ class info(subtest.Subtest):
 
     @staticmethod
     def _build_table(cli_output):
-        out = cli_output.split('\n')
-        out = [x.split(':', 1) for x in out if x]
-        keys = [x[0].strip() for x in out]
-        vals = []
-        for x in out:
-            try:
-                vals.append(x[1].strip())
-            except IndexError:
-                vals.append(None)
-        return dict(zip(keys, vals))
+        """
+        'docker info' returns a human-readable list of key: value pairs.
+        Some are indented by a space, indicating that these key/values
+        are subelements of a previous element. To wit:
+
+             Images: 3
+             Server Version: 1.12.6
+             Storage Driver: devicemapper
+              Pool Name: vg--docker-docker--pool
+              Pool Blocksize: 524.3 kB
+              ...
+             Logging Driver: journald
+
+        In this case 'Images', 'Server Version', and 'Logging Driver'
+        are simple tuples but 'Storage Driver' has both a value and
+        a set of further key/value pairs: Pool Name, Pool Blocksize, ...
+
+        We parse that and return a dict with the expected key/value
+        mapping *and* an extra: for all elements with subelements,
+        'element...' (element name plus three dots) is a dict containing
+        the subelements. E.g. x['Storage Driver...']['Pool Name'] = 'vg--etc'
+        """
+        out = {}
+        current_key = None
+        for line in cli_output.split('\n'):
+            # Almost every line will be Foo: Bar, but 'Insecure Registries:'
+            # is followed by a simple list of IPv4 netmasks
+            if ': ' in line or line.endswith(':'):
+                (key, value) = [e.strip() for e in line.split(':', 1)]
+            else:
+                (key, value) = (line.strip(), '')
+            if line.startswith(' '):
+                if not current_key:
+                    raise IndexError("sdfsdf")
+                if current_key not in out:
+                    out[current_key] = {}
+                out[current_key][key] = value
+            else:
+                out[key] = value
+                current_key = key + '...'
+        return out
 
     def postprocess(self):
         # Raise exception on Go Panic or usage help message
         outputgood = OutputGood(self.stuff['cmdresult'])
         info_map = self._build_table(outputgood.stdout_strip)
-        # Verify some individual items
-        self.failif_ne(info_map['Storage Driver'].lower(), 'devicemapper',
-                       'Storage Driver')
+
+        # We support multiple storage drivers. Each one has a different
+        # set of key/value settings under 'info'; so each one has a
+        # dedicated helper method for validating.
+        storage_driver = info_map['Storage Driver'].lower()
+        try:
+            handler = '_postprocess_' + storage_driver
+            getattr(self, handler)(info_map['Storage Driver...'])
+        except AttributeError:
+            raise DockerTestFail("Unknown storage driver: %s" % storage_driver)
+
+        # Count 'docker images', compare to the 'Images' info key.
+        # Yes, that's unreliable on a busy system. We're not on a busy system.
+        di = DockerImages(self)
+        di.images_args = "%s --all" % di.images_args
+        img_set = set(di.list_imgs_ids())  # don't count multi-tags
+        img_cnt = int(info_map['Images'])
+        self.failif_ne(len(img_set), img_cnt,
+                       "count of 'docker images' vs 'docker info'->Images")
+
+    def _postprocess_devicemapper(self, info_map):
+        """
+        Verify docker info settings for devicemapper storage driver.
+        """
         self.failif_ne(info_map['Data file'].lower(), '',
                        'Data file')
         self.failif_ne(info_map['Metadata file'].lower(), '',
                        'Metadata file')
-        di = DockerImages(self)
-        # Make sure nothing is 'hidden'
-        di.images_args = "%s --all" % di.images_args
-        # Possible race-condition here...
-        img_set = set(di.list_imgs_ids())  # don't count multi-tags
-        # ...with this
-        img_cnt = int(info_map['Images'].lower())
-        self.failif_ne(len(img_set), img_cnt,
-                       "More/less images %d than info reported %d"
-                       % (len(img_set), img_cnt))
+
         # verify value of elements
         self.verify_pool_name(info_map['Pool Name'])
         data_name = 'Data loop file'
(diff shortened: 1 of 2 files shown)


edsantiago:
Tested on RHEL 7.3 + devicemapper, 7.4 + overlay2

cevich:
Oh super...looking right now...

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+
+        In this case 'Images', 'Server Version', and 'Logging Driver'
+        are simple tuples but 'Storage Driver' has both a value and
+        a set of further key/value pairs: Pool Name, Pool Blocksize, ...
+
+        We parse that and return a dict with the expected key/value
+        mapping *and* an extra: for all elements with subelements,
+        'element...' (element name plus three dots) is a dict containing
+        the subelements. E.g. x['Storage Driver...']['Pool Name'] = 'vg--etc'
+        """
```
cevich: oh bless your heart, **DOCUMENTATION!**  What a treat (just don't expect me to find any speling mistakes).

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+        out = {}
+        current_key = None
+        for line in cli_output.split('\n'):
```
cevich: I think this should be ``for line in cli_output.splitlines():``

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+            # Almost every line will be Foo: Bar, but 'Insecure Registries:'
+            # is followed by a simple list of IPv4 netmasks
+            if ': ' in line or line.endswith(':'):
+                (key, value) = [e.strip() for e in line.split(':', 1)]
+            else:
+                (key, value) = (line.strip(), '')
```
cevich: This could be made slightly clearer by just spelling it out (it only adds one line)

```python
key = line.strip()
value = ''
```

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+        'element...' (element name plus three dots) is a dict containing
+        the subelements. E.g. x['Storage Driver...']['Pool Name'] = 'vg--etc'
+        """
+        out = {}
+        current_key = None
+        for line in cli_output.split('\n'):
+            # Almost every line will be Foo: Bar, but 'Insecure Registries:'
+            # is followed by a simple list of IPv4 netmasks
+            if ': ' in line or line.endswith(':'):
+                (key, value) = [e.strip() for e in line.split(':', 1)]
```
cevich: clever!

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+            else:
+                (key, value) = (line.strip(), '')
+            if line.startswith(' '):
+                if not current_key:
+                    raise IndexError("sdfsdf")
```
cevich: S.D.F.S.D.F:  Shoot down the flying saucer you damn fool!

?

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+                if current_key not in out:
+                    out[current_key] = {}
+                out[current_key][key] = value
+            else:
+                out[key] = value
+                current_key = key + '...'
```
cevich: What's the purpose of the ``...`` when you can ``if isinstance(value, dict)`` on ``value`` later?

Still reading...

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
-        self.failif_ne(info_map['Metadata file'].lower(), '',
-                       'Metadata file')
+
+        # We support multiple storage drivers. Each one has a different
+        # set of key/value settings under 'info'; so each one has a
+        # dedicated helper method for validating.
+        storage_driver = info_map['Storage Driver'].lower()
+        try:
+            handler = '_postprocess_' + storage_driver
+            getattr(self, handler)(info_map['Storage Driver...'])
```
cevich: This seems a little complex for only having two supported handlers.  How about:

```python
if "overlay2" in storage_driver:
     self._postprocess_overlay2(...)
elif "devicemapper" in storage_driver:
    self._postprocess_devicemapper(...)
else:
    self.logwarning("sdfsdf")
```

Beyond two handlers, the common way to do this is with a class attribute, mapping from driver-name to method.  Then you can call it like:  ``self.handlers[storage_driver](info_map[storage_driver)`` or something to that effect.

OTOH, what you have will work for all cases, so I'm not going to make any more fuss over it if you just ignore me 😁

subtests/docker_cli/info/info.py
```diff
@@ -40,38 +41,81 @@ def run_once(self):
...
+        except AttributeError:
+            raise DockerTestFail("Unknown storage driver: %s" % storage_driver)
```
cevich: Consider adding:

```python
            except KeyError, xcept:
                raise(DockerTestFail("Storage driver: %s is missing expected info key: %s" % (storage_driver, xcept.msg))
```

subtests/docker_cli/info/info.py
edsantiago: Fixed, thanks.

subtests/docker_cli/info/info.py
edsantiago: Fixed

subtests/docker_cli/info/info.py
edsantiago: Oops. Blush. Fixed.

subtests/docker_cli/info/info.py
edsantiago: Dealing with this:

    Storage Driver: overlay2
        Foo: bar
        Bar: baz

You can't just set `out['Storage Driver']` to a dict, because you lose the `= overlay2` value. I could set it to a dict with something like `['__value'] = overlay2` but that feels clumsy. I could create a new dict-based class that treats string compares magically, but I'm not that insane. This seemed the least evil approach.

subtests/docker_cli/info/info.py
edsantiago: My goal was, and always is, to avoid unnecessary unnecessary duplication. A secondary goal is to avoid long `if-else`s. With my scheme, to add support for a new back end, you add a new function (nothing else). With the above one, you need to (1) add a new function, (2) add a new `elif` and make sure you get the string right, (3) add a call to the new function. And you get a hard-to-read `if` block as well. I'm a big fan of table-driven programming where there's fewer things to go wrong.

I don't understand your class-attribute-handler-etc comment but will spend more time looking into it. Thanks for the suggestion.

subtests/docker_cli/info/info.py
edsantiago: Good idea, fixed.

subtests/docker_cli/info/info.py
cevich: What about ``{"blah": "foo", "Storage Driver": {"name": "overlay2", "bar":"baz", }}``

subtests/docker_cli/info/info.py
cevich: Oh nono, I mean an ``if...elif...else`` _just_ for the current, two-handler case.  Any more and I completely agree, it would be a mess.

The class attribute idea is just a lookup table for the methods, that doesn't rely on ``self.__dict__``.  It plays out better if the class ever gets split into a base + ancestor classes.  Though for here, in this context, what you have is fine, so we can just leave it as-is.

edsantiago:
Augh! Guess I shouldn't have used the word "Fixed" when I meant "Fixed but not pushed because I'm still trying to figure out the class attribute thing". Sorry. I'll submit a cleanup PR tomorrow.

subtests/docker_cli/info/info.py
edsantiago: `name` works in this case but not for anything else, e.g. `Containers` (which has a numeric value). That's why I suggested `__value` above, it's context-neutral and also unlikely to be a real key value emitted by docker info.

I still don't know what the right answer is, or if there is one. I prefer the access-time readability of:

    storage_driver = info_map['Storage Driver'].lower()
    img_cnt = int(info_map['Images'])
vs

    storage_driver = info_map['Storage Driver']['__value'].lower()
    img_cnt = int(info_map['Images']['__value'])

but I recognize that the `info_map['Storage Driver...']` form is not ideal. I think it's adequate, and I don't want to be paralyzed in search of a possibly-nonexistent perfection.

subtests/docker_cli/info/info.py
edsantiago: Thanks for your email last night (re: handlers). I still see a lot of duplication there. I still prefer a mechanism with no duplication, so I'd like to leave my lookup in place.

subtests/docker_cli/info/info.py
cevich: I gotcha.  My recommendation is based on the way YAML/JSON represent trees of values.  Basically any value more complex than a canonical string, goes into a dictionary.  It's a widely used/understood data structure/format, like what XML wishes it was.  In practice, it makes it super easy to extend/maintain code using these structures because unknown keys are ignored, and missing keys can get default values.  Same with the access patterns like ``foo = bar['parent']['child']`, they're used all over the place and very well understood.  In fact, a lot of people would see that and immediately think, where's the JSON or YAML file that this comes from.

Anyway, your right, this is just a test, no need for it to be squeaky clean or get bent up over little details.  You answered my initial question, and my "what about" is just an idea...they grow on trees, we're in the middle of a forest, and I always bring my axe :grinning:

We can roll with what you've got.

subtests/docker_cli/info/info.py
cevich: Ya, the table-of-things to call is a widely-used pattern, across many languages.  I only recommend these things (code patterns) because using them brings almost instant familiarity for anyone who knows them, even if they don't speak python well.  It's also why people have written books on the topic 😀   Anyway, yes, absolutely not critical here, what you have is fine.

 If you're even in need of bedtime reading material...I recommend [getting lost in the wikipedia pages](https://en.wikipedia.org/wiki/Software_design_pattern).  They're a little dense, but if you meditate upon them, they'll sink-in eventually.  Most/all of them transcend across most languages, so they're nearly all generally applicable (with some tweaking).

PR #707, opened 2017-04-20
Add test: Verify yum -y update works in container

cevich:
Extremely simple test, runs the ``yum -y update`` command inside a
container, verifying exit status 0.  Additionally, this test
configuration acts as a placeholder, to be overridden by a more
complex for testing other distros, such as RHEL.

In particular, RHEL maintains patches against upstream docker that
manipulate behavior in subtle ways.  For example, by sharing a
special "secrets" volume mount, to enable subscription-manager
to do it's job inside a RHEL base-image container.

This test enables both verifying the simple case, and the more complex
code paths present where additional docker behavior is involved.

### DIFF
--- a/config_defaults/subtests/docker_cli/run.ini
+++ b/config_defaults/subtests/docker_cli/run.ini
@@ -1,5 +1,5 @@
 [docker_cli/run]
-subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_tmpfs,run_passwd,run_install
+subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_tag,run_names,run_tmpfs,run_passwd,run_install,run_update
 #: The most basic sub-subtests can be generated from a generic
 #: class at runtime, set 'yes' to enable this feature.
 generate_generic = no
@@ -77,3 +77,7 @@ cmd = /bin/true
 install_cmd = yum install --disablerepo="*" --enablerepo="base" --assumeyes yajl
 #: Full command to run in commited container to verify installation
 verify_cmd = bash -c "echo '[{}]' | json_verify -q"
+
+[docker_cli/run/run_update]
+generate_generic = yes
+cmd = 'yum --assumeyes update'
--- a/config_defaults/subtests/docker_cli/run.ini
+++ b/config_defaults/subtests/docker_cli/run.ini
@@ -5,7 +5,7 @@ subsubtests = run_true,run_false,run_interactive,run_attach_stdout,run_remote_ta
 generate_generic = no
 #: CSV of options and arguments to use on docker command line,
 #: with the exception of --name, image, and contained command.
-run_options_csv = --rm,--attach=stdout,--sig-proxy=true
+run_options_csv = --rm
 #: Set 'yes' to append ``--name`` with a test-specific random name.
 run_append_name = yes
 #: CSV docker run COMMAND argument to prefix ``cmd`` option (below)


cevich:
[(5) Secrets patch with Swarm integration](https://trello.com/c/H6fFRdp9/357-5-secrets-patch-with-swarm-integration)

edsantiago:
Can I safely assume that this PR is incomplete, and that you'll fill in the missing parts later? If not, what am I missing?

cevich:
@edsantiago It's linked above by reference, but here it is in a comment:

Missing parts: https://github.com/cevich/autotest-docker-enabled-product-testing/pull/133

😀

Mmmmm...though that probably suggests my commit-comment is bad.  Fixing...

edsantiago:
I guess I meant: is there a `run_update.py`? Is there a corresponding addition of `run_update` to the `subsubtests=` line in `run.ini`?

cevich:
> I guess I meant: is there a run_update.py

Ahh, no, there doesn't need to be.  That's what ``generate_generic = yes`` does.  The run test will create the sub-subtest at runtime with a class generator (a.k.a metaclass)

edsantiago:
How do I invoke it? 

    # ./autotest-local run docker --args docker_cli/run
    ....the string "run_update" does not appear anywhere under results/default

    # ./autotest-local run docker --args docker_cli/run/run_update
    TestError: Sub-subtest(s) set(['docker_cli/run/run_update']) referenced in control include, exclude, and/or subthings but not defined in subsubtests = [.......]

edsantiago:
OK, got it. Need to add `run_update` to the `subsubtests =` line. The test now runs. I'm still not sure what it accomplishes.

cevich:
> OK, got it.

Ahh yes, good catch.  I'll fix that.

cevich:
> I'm still not sure what it accomplishes.

Ugg, what? You __STILL__ can't read my mind?!?!?!?

It's for this card: https://trello.com/c/H6fFRdp9/357-5-secrets-patch-with-swarm-integration

cevich:
@edsantiago good to merge this then https://github.com/cevich/autotest-docker-enabled-product-testing/pull/133 ?

edsantiago:
I still don't get it. I don't see this as a useful test of secrets. I can see yum update failing for many reasons that have nothing to do with secrets--eg timeout, network issues, our usual bane with the "-tus" unreachable repo--and even succeeding without testing any secrets (running a centos/fedora image instead of rhel). I can see how it's a quick-and-easy addition but I don't see it actually adding coverage.

cevich:
Unreachable repos / networking problems affect many tests, not anything we can do or control easily there.  

okayokay, more splainin' then:

For non-RHEL, rpm-based systems, this test does simply what the title says.  So it's more about testing the base-image than anything else.  However, on the RHEL-side, configuration is customized by ADEPT to only update from a specific repo.  That repo is only accessible by the rhel base-image container if secrets is working, and the host has a valid subscription.

Runcom's request in https://trello.com/c/H6fFRdp9/357-5-secrets-patch-with-swarm-integration specificly spelled out the following test-steps to be executed on RHEL and RHELAH:

```bash
$ sudo subscription-manager register --auto-attach
$ sudo yum update -y # this should work on host after subscribed
$ docker run -ti registry.access.redhat.com/rhel7 bash
# # inside the container
# yum update -y # <- this should work inside the container if host is subscribed
```

The first two steps are taken care of automatically by ADEPT (before autotest is even started).  The third step is generated by the run sub-subtest prior to running the last command (configured in ADEPT by https://github.com/cevich/autotest-docker-enabled-product-testing/pull/133.

It's not perfect, but it is quick and easy, and doesn't add to the list of to-be-written tests.  Better yet, you can see from the CI results, the test already passes 😀 

Does that help? Should I order more take out duct-tape soup 😀

edsantiago:
It's still not a test of secrets. Indirectly, it is, but it's not an actual test. All it is is quick & easy.

I can accept it, but as a compromise, could it somehow be extended so the yum upgrade stdout/stderr get logged?

cevich:
> Indirectly, it is, but it's not an actual test.

Indirect testing is what was requested, and is probably fine, though it's a pretty wide net.  So I also agree we could use more thorough testing of secrets...Hmmm....ahh, we have a card already: https://trello.com/c/xMr5wvwu/301-3-ci-redhat-secrets-test

> so the yum upgrade stdout/stderr get logged?

Ewww, now that __is__ a big problem that needs fixing...

cevich:
...okay, that should do it.  Good catch BTW, It turned out to be a double-whammy.  The test was **only** attaching to ``stdout`` _and_ never logging anything unless the test failed (exit 0 or bad output).  Waiting on travis and ci results to check if I made any typos...

edsantiago:
Sigh. I started my test, then vpn went down. Waiting...

cevich:
Ahhh n{e,o}tworking.

subtests/docker_cli/run/run.py
```diff
@@ -73,6 +73,7 @@ def run_once(self):
     def postprocess(self):
         super(run_base, self).postprocess()  # Prints out basic info
         dockercmd = self.sub_stuff['dkrcmd']
+        self.logdebug(dockercmd)
```
edsantiago: I'm having a hard time with this. My sense is that it's overkill: The docker command is already logged a few lines earlier; we don't really need to see exit status or stderr because those are already checked for and displayed on error. What I was asking for is just the stdout from 'yum update', because I keep thinking there will be cases where the test passes without actually testing secrets. Maybe I'm just being too paranoid.

The added verbosity, in an already WAY too verbose debug log (the run test), bothers me more than a merely hypothetical failure. I take back my request: let's just skip logging, and deal with problems if they arise.

cevich: okay, and remember too the test was only attached to stdout, so we probably were throwing away any errors anyway.

cevich:
CI passed, only failures are packaging problems (missing debuginfo).

PR #708, opened 2017-04-20
Grab messages in addition to audit.log

cevich:
Atomic host doesn't have ``/var/log/audit``, so all selinux stuff goes into
``/var/log/messages``.  In all cases we should be grabbing a copy of both
after every test.

### DIFF
--- a/control
+++ b/control
@@ -695,15 +695,17 @@ class StepInit(Context, collections.Callable):
         return [subthing for subthing in subthings
                 if subthing in subtest_modules]
 
-# Grab a compressed copy of the audit log after every test
+# Grab a compressed copy of the audit log and messages after every test
 # N/B: Autotest silently ignores any failures with this.
-audit_log = '/var/log/audit/audit.log'
-if os.path.isfile(audit_log):
-    command = 'gzip -9c %s' % audit_log
-    # on_every_test=False has already run (before control file loaded)
-    job.add_sysinfo_command(command, "audit.log.gz", on_every_test=True)
-else:
-    logging.warning("Can't record %s because it doesn't exist", audit_log)
+for filepath in ('/var/log/audit/audit.log', '/var/log/messages'):
+    if os.path.isfile(filepath):
+        command = 'gzip -9c %s' % filepath
+        # on_every_test=False has already run (before control file loaded)
+        job.add_sysinfo_command(command,
+                                "%s.gz" % os.path.basename(filepath),
+                                on_every_test=True)
+    else:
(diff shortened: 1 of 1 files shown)


control
```diff
@@ -695,15 +695,16 @@ class StepInit(Context, collections.Callable):
...
 # N/B: Autotest silently ignores any failures with this.
-audit_log = '/var/log/audit/audit.log'
-if os.path.isfile(audit_log):
-    command = 'gzip -9c %s' % audit_log
-    # on_every_test=False has already run (before control file loaded)
-    job.add_sysinfo_command(command, "audit.log.gz", on_every_test=True)
-else:
-    logging.warning("Can't record %s because it doesn't exist", audit_log)
+for gzname, filepath in {'audit.log.gz': '/var/log/audit/audit.log',
+                         'messages.gz': '/var/log/messages'}:
```
edsantiago: Why the duplication? Why not just enumerate paths, and use `os.path.basename(filepath)` instead of gzname?

control
```diff
@@ -695,15 +695,16 @@ class StepInit(Context, collections.Callable):
...
+    if os.path.isfile(filepath):
+        command = 'gzip -9c %s' % filepath
+        # on_every_test=False has already run (before control file loaded)
+        job.add_sysinfo_command(command, gzname, on_every_test=True)
+    else:
+        logging.warning("Can't record %s because it doesn't exist", gzname)
```
edsantiago: Someone reading this log might appreciate seeing the full filepath, not just the basename.

control
cevich: I was lazy

control
cevich: okay, working a fixup...

cevich:
> depends on whether the logs actually get preserved.

That's the sort of thing the ci job is for, let's see...hmmm, nope.  There must be some problem/bug b/c it was capturing ``audit.log`` previously.

cevich:
...Okay, here's the smoking gun.  On one of the atomic hosts, the main debug log shows:
```
04/20 20:29:08 WARNI|   control:0708| Can't record /var/log/audit/audit.log because it doesn't exist
04/20 20:29:08 WARNI|   control:0708| Can't record /var/log/messages because it doesn't exist
```

But on one of the RHEL hosts, both files are there and compressed, for example:
https://hurl.corp.redhat.com/3d56d97

So this PR is good, working as designed, the problem is with Atomic Host.  I have a feeling the solution here is to get the ``rsyslog`` container running.  I can test/tackle that via a PR to ADEPT.  For now, I think this one is good for merge.

PR #709, opened 2017-04-25
Fix run_volumes_selinux failures on Atomic Host after upgrade

cevich:


cevich:
@edsantiago holding onto fixes for travis findings until Red Hat Internal Testing CI job completes.

dockertest/environment.py
```diff
@@ -62,3 +62,24 @@ def get_selinux_context(path):
...
+    try:
+        if selinux.selinux_status_open(OLDKERNEL) == ERROR:
+            raise ValueError("Could not r/o mmap /sys/fs/selinux/status")
+        mode = selinux.selinux_status_getenforce()
+        if mode == ERROR:
+            raise ValueError("Error checking SELinux is disabled,"
+                             " permissive, or enforcing")
+    finally:
+        selinux.selinux_status_close()
+    return mode == ENFORCING
```
edsantiago: Where is this selinux module documented? In particular, where are these magic ERROR and OLDKERNEL and ENFORCING values described? `pydoc selinux` does not help in any way. I think future maintainers might greatly appreciate it if you just invoke `getenforce`.

cevich: The python module is auto-generated from the C source-code, so yes the python docs are really really ugly.  The best thing is to read the C documentation...which I can't find again at the moment.

cevich: ...still looking...

cevich: ...found it.

dockertest/environment.py
```diff
@@ -62,3 +65,13 @@ def get_selinux_context(path):
...
+
+
+def selinux_is_enforcing():
+    """
+    Return True if selinux is currently in enforcing mode.
+    """
+    # Travis complains about missing selinux.ENFORCING
+    # pylint: disable=C0103
+    ENFORCING = 1  # Travis complains about missing selinux.ENFORCING
+    return selinux.security_getenforce() == ENFORCING
```
cevich: ^^^^^ @edsantiago a bit simpler now 😸

dockertest/environment.py
```diff
@@ -9,6 +9,9 @@
 
 import warnings
 import subprocess
+# N/B: This module is automaticly generated from libselinux, so the
+# python docs are terrible.  The C library man(3) pages provided by
+# the 'libselinux-devel' RPM package (or equivilent) are much better.
```
cevich: **THAAAA** docs 😀

dockertest/environment.py
edsantiago: I'm still not sure what the purpose is of this selinux module, but yes, that's better. Thanks especially for the link to actual documentation.

And speaking of documentation, since security_getenforce(3) just says "returns 0 / 1 / -1" (not named constants), why not remove the entire ENFORCING constant, and comment, and pylint, and just make it `== 1`?

And to make it even cleaner, rename this function to `selinux_getenforce` and change the `failif` below to `failif_ne(selinux_getenforce, 1, ...)` and that way you get to differentiate between 0, 1, and the not-expected but you-never-know -1.

cevich:
/me sings the "Wee aalllll love Ubuntu" song.

dockertest/environment.py
cevich: So....you're proposing I re-write the entire thing because it's outputting enough emoji's in debug log?

cevich: ...fair point about the -1...lessee...

edsantiago: I don't think I understand your emoji comment. 

What I'm suggesting is: since the documentation uses 0 and 1, and since anyone who has ever run `setenforce` on the command line already knows about 0 and 1, I think it's safe to assume that 0/1 are known conventions and it's OK to push them up one level. Using `failif_ne` instead of `failif` has a big advantage in that the actual and expected values are displayed right there in the failure message, making it easy for a maintenance programmer to see exactly what happened.

cevich: yes...but then I need to do more typing.

How about...

cevich:
...whew! Okay! Force-pushed, maybe (we'll see) both Ed and Travis can finally be happy! 😀

edsantiago:
I'm not sure if I'm supposed to look at something here, or wait for an update. For now, waiting.

cevich:
> I'm not sure if I'm supposed to look at something here, or wait for an update. For now, waiting.

I force-pushed a change re: rewrite and rename ``selinux_is_enforcing()``.  It only does the later, because my ego got in the way of doing the former 😀 (though your ``failif_ne`` suggestion isn't lost on me, I'm still considering it).  What's there now, I'm calling a "compromise".

Still TODO: Club Travis with a rusty rail-road rail 😖

cevich:
Note: Still getting

```
************* Module dockertest.environment
(pylint) E0401: 15,0: : Unable to import 'selinux'
```

However, unit-tests pass and at least one imports virtualenv modules.  Looks like virtualenv's pylint isn't respecting the virtualenv somehow/way.  Source-code shows that it does :S

edsantiago:
This was all working. I'm really not sure what's going on here that requires such enormous changes for something so simple.

cevich:
> This was all working. I'm really not sure what's going on here that requires such enormous changes for something so simple.

The main problem I've updated in a few comments.  Basically the issue is, ``python-selinux`` is not available via pip, only as a system package.  However, all of the unittests and pylint/pep stuff runs (in theory) from the python virtual env (where pip stuff was installed).  Since you previously disabled ``use_system_site_pacakges`` as a workaround for a urllib2 bug, we must ensure that the selinux module is available from w/in the virtualenv.

I've gotten that far, however for some reason it seems pylint isn't grabbing the virtual env. selinux module.  I think this means I'm either copying it into the wrong place, or there's some other "special" problem unique to pylint (because pep8 and everything else works fine).

Three-letter-word of the day is "ick".  Taking a step back for a little while to let my unconscious mull over the bigger picture.

cevich:
[(2) [CI/CN] Fix selinux: permissive in ADEPT and Docker Autotest](https://trello.com/c/H7hlA8at/464-2-ci-cn-fix-selinux-permissive-in-adept-and-docker-autotest)

cevich:
okay, after much trial-and-error, I've confirmed this is for-sure a pylint-specific problem.  The virtual env. Travis is using works fine, ``PYTHONPATH`` shenanigans aren't needed.  It's only pylint that for whatever reason can't seem to load the selinux module even-though it exists in the right place, and python itself can load it just fine.  Whew!

So, all that's needed I think is a liberal application of duct-tape onto pylint so it can't remove it's dunce-cap.

cevich:
Garrrrrrg!  That should have fixed it.  WTF Travis?!?!?!?!

cevich:
Attempting to revert back as a sanity-check this was actually failing earlier and I'm not loosing my marbles...

cevich:
/me thinking now, f-tha selinux module, and just run the damn command-line tools.

edsantiago:
/me remembers suggesting that several months ago

cevich:
Okay, I'm not crazy.  If I roll back the pylint-fixes, we get failures in both the unittests __and__ under pylint:  https://travis-ci.org/autotest/autotest-docker/builds/226164937

cevich:
> /me remembers suggesting that several months ago

IKR...what's really bugging me is altering the code to fit Travis's (literally) unrealistic environment, is just pushing the main problem further down the road.  What happens when there's another module that fails to get updated or be properly supported, and then we're back in the same boat.

In fact, already countless hacks and hours have been wasted to junk up ``run_pylint.sh`` (and elsewhere) just to keep Travis happy.  While we actually give **zero-F*c&s** about ever running Docker Autotest under Ubuntu, inside a virtualenv, or inside a container.

In other words, this problem only exists because we're trying to squish Travis into the RHEL-like hole.  This is not problem about ``selinux`` or even the ``urlib2`` module.  I think instead of band-aid after band-aid, we should consider just dropping Travis entirely, and use out own RHEL-based CI to TCB.  I mean it's only four commands we need to run:

``
    - SPHINXOPTS="-W" make
    - ./run_checkdocs.py
    - ./run_unittests.sh
    - ./run_pylint.sh --FF
``

Better yet, there's already a mechanism for doing this under RHEL.  jlebon's ``.redhat-ci.yml``.  Let me look into getting that going here (I've already researched it for use elsewhere).  Then we solve this **and** future Travis/Ubuntu stupidity.

cevich:
Opened https://github.com/jlebon/redhat-ci/pull/34

cevich:
Change of plans, best not hold this PR hostage to broken Travis and adding RedHat-CI, let's do a separate PR for that.  Rejiggering...

cevich:
Thanks for your patients.

PR #710, opened 2017-04-26
Add docker-latest to list of BZ components to scan

cevich:

### DIFF
--- a/control
+++ b/control
@@ -57,7 +57,7 @@ def subtests_subsubtests(subthing_set, subtest_modules):
     subtest_to_subsubtest = {}
     for subthing in subthing_set:
         parent = subtest_of_subsubtest(subthing, subtest_modules)
-        if parent == None:
+        if parent is None:
             subtest = subthing
             subsubtest = set()
         else:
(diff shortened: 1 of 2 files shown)


cevich:
> LGTM

Unf. this one is also winding up to be a major PITA.  I'm having to update this stupid value all over the place in ADEPT and elsewhere.  Working on an addition that will just use the defaults if left unspecified.  Plus this helps keep my mind off of #709 😀 for a little while.

cevich:
@edsantiago Okay, that last commit passes my basic insanity-checks.  Having it means we don't need to update every bloody option in every CI job just to change a few (like BZ username/password).  It also means I don't need to track changes to the BZ components list in every copy of the file as well.

cevich:
Example of headache this removes: https://github.com/cevich/autotest-docker-enabled-product-testing/pull/135/files

control
```diff
@@ -153,64 +153,58 @@ class ControlINI(ConfigParser.SafeConfigParser, Singleton):
...
-                self.add_section(section)
-            except ConfigParser.DuplicateSectionError:
-                pass  # already existing section
-            # Empty string is the default value
-            self.set(section, option, '')
-        # These option values default to the option name, look up section
-        for key in ('pretests', 'subtests', 'intratests', 'posttests'):
-            self.set(self.OPT_SEC_MAP[key], key, key)
+
+    def defaults(self):
```
edsantiago: Is this a special Python magic? I don't see anywhere that invokes `defaults`

edsantiago: Never mind; looks like you fixed that in a force push.

I'm going to back away until this is stable.

cevich: It's called from ``__init__()``, but even if not, it's [part of the ``ConfigParser`` interface](https://docs.python.org/2.7/library/configparser.html?highlight=configparser#ConfigParser.RawConfigParser.defaults).  Though perhaps I can/should make this more clear.

cevich:
okay, --force push with some more comments about ``defaults()``

cevich:
Should be stable now.

control
```diff
@@ -153,64 +153,59 @@ class ControlINI(ConfigParser.SafeConfigParser, Singleton):
...
+        returns: List of items loaded representations
+        """
+        if filenames is None:  # Preserve ancestor interface just in case
+            filenames = []
+        filenames.append(self.control_ini_custom)
+        result = ["<Empty Defaults>"]
+        # Try load defaults via readfp as documented
+        try:
+            with open(self.control_ini_default, 'rb') as defaults_file:
+                super(ControlINI, self).read(defaults_file)
```
edsantiago: Sorry, this isn't an actual review; I'm still focusing on deferred deletion / mount spaces. But at a quick glance: is this correct? Running read in void context? Shouldn't this be what's added to `result`, instead of a filename?

cevich: oh good catch, that should be ``readfp`` as the comment says.  Thanks.

PR #711, opened 2017-04-27
Give Travis CI the boot

cevich:
Too much effort has been spent trying to squeeze it's Ubuntu-based,
container and virtualenv environment into a RHEL-work-a-like.  Drop
Travis entirely, and instead rely on a TBD system for running the four
quick-check commands:

- SPHINXOPTS="-W" make
- ./run_checkdocs.py
- ./run_unittests.sh
- ./run_pylint.sh --FF

Replace the quick-code checking with jlebon/redhat-ci

Similar idea as travis, only it runs on RHEL, CentOS, and Fedora (or all
three at the same time).  This is also the same CI framework used by
most of the projectatomic repos.

Ref: https://github.com/jlebon/redhat-ci


jlebon:
test comment

cevich:
[(1) [CN] Replace Travis w/ RedHat-CI for testing docker-autotest itself](https://trello.com/c/I8Ym5GMN/466-1-cn-replace-travis-w-redhat-ci-for-testing-docker-autotest-itself)

.redhat-ci.yml
```diff
@@ -0,0 +1,62 @@
+---
+
+# Example format:
+# https://github.com/jlebon/redhat-ci/blob/master/sample.redhat-ci.yml
+
+# Internal Job (for realtime monitoring):
+# https://hurl.corp.redhat.com/eb0faa3
+
+container:
+    image: docker.io/stackbrew/centos:latest
```
jlebon: I think you want just `centos:latest` here right?

cevich: same-same-only-different 😃

jlebon: Just going off of what [the page](https://hub.docker.com/r/stackbrew/centos/) says (I hadn't heard of stackbrew before):

> Deprecated; use 'centos' from https://hub.docker.com/_/centos/ instead

The `centos` image is the official one. I'm guessing the stackbrew one was available before `centos` existed?

cevich: ahh interesting.  Naw, it's part of their build-system, has always existed as far as I know.  It's not a big deal, can switch to the shorter form for conformity-sake.

PR #714, opened 2017-04-28
Fix Broken BZ search by DEFAULTS section

cevich:
When SafeConfigParser is initialized with a defaults dict,
it results in production of a ``[DEFAULTS]`` section when written.
For some reason this prevents properly connecting to bugzilla.
Possibly because the default keys are applied globally, to all sections.

Fix this by not passing a defaults dict, and instead hard-coding the
expected format (including sections) of this file.  This prevents
output of a ``[DEFAULTS]`` section and the BZ search happens normally.

### DIFF
--- a/control
+++ b/control
@@ -170,18 +170,29 @@ class ControlINI(ConfigParser.SafeConfigParser, Singleton):
 
     def __init__(self):
         # Inject defaults dict into ancestor's initialization
-        super(ControlINI, self).__init__(self.defaults(), allow_no_value=True)
+        super(ControlINI, self).__init__(allow_no_value=True)
         self.optionxform = str  # support case-sensitive options
-
-    def defaults(self):  # Part of the ConfigParser interface
-        """Return a dictionary containing the instance-wide defaults."""
-        return dict(include='', exclude='', subthings='',
-                    pretests='pretests', subtests='subtests',
-                    intratests='intratests', posttests='posttests',
-                    url='', username='', password='',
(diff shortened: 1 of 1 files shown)


cevich:
Sadly, I have no better idea why this fixes the problem.  I compared all the keys/values from before/after and there are no differences.  The reason this problem slipped through is, I have BZ scanning turned off on the docker-autotest CI job (to allow us to see all results).

control
```diff
@@ -170,18 +170,29 @@ class ControlINI(ConfigParser.SafeConfigParser, Singleton):
...
+                                                   username='',
+                                                   password='',
+                                                   excluded='',
+                                                   key_field='',
+                                                   key_match=''),
+                                     Query=dict(product='',
+                                                component='',
+                                                status='')).iteritems():
+            for option, value in options.iteritems():
+                if not super(ControlINI, self).has_section(section):
```
edsantiago: Why is this inside the second for loop?

cevich: In order to slow things down measurably yet imperceptibly 😀

cevich: ``--force`` pushed w/ fix.  Thanks 😄

cevich:
> I wish there was a simple way to write the defaults in .ini format, for legibility. This is a reasonable 

There is, it's in ``config_defaults/control.ini`` 😬 

The ugly one is only there to enforce the structure and empty values.  Not sure how to express the structure any better, ya, it's a compromise.

PR #715, opened 2017-05-03
Deferred deletion

edsantiago:


cevich:
OMG! This is HUIGE, no way I can review this with only one cup of coffee!

edsantiago:
Please note: three separate commits, each of them reviewable.

cevich:
[(5) [CN] Docker integration tests for dm.use_deferred_removal and dm.use_deferred_deletion options.](https://trello.com/c/BeawcX2a/30-5-cn-docker-integration-tests-for-dm-use-deferred-removal-and-dm-use-deferred-deletion-options)

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,104 @@
...
+import collections
+import subprocess
+
+
+class DockerInfo(collections.MutableMapping):
+    """
+    Parser of 'docker info' output
+    """
+
+    def __init__(self, info_string=None, docker_path=None):
```
edsantiago: Fixed: it is now `DockerInfo(object)`

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,104 @@
...
+        if info_string is None:
+            if docker_path is None:
+                docker_path = 'docker'
+            info_string = subprocess.check_output(docker_path + ' info',
+                                                  shell=True,
+                                                  close_fds=True)
```
edsantiago: "recursive initialization": I don't understand what this means. I don't see recursion anywhere. All `__init__` does is run `docker info` and preserve the values.

"docker_path": all I did was copy from `dockertest/output/dockerversion.py` which has the same docker_version in __init__. I don't see a strong reason for it, but can imagine it being useful for testing. If I remove it, should I remove it from `dockerversion.py` also?

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,104 @@
...
+        self.info_string = info_string
+        self.table = self._build_table(info_string)
```
edsantiago: I don't understand your phrasing. Are you saying I _should_, or _should not_, call `super`? Right now there is no `super`.

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,104 @@
...
+            # Almost every line will be Foo: Bar, but 'Insecure Registries:'
+            # is followed by a simple list of IPv4 netmasks
+            if ': ' in line or line.endswith(':'):
+                (key, value) = [e.strip() for e in line.split(':', 1)]
+            else:
+                key = line.strip()
+                value = ''
+            if line.startswith(' '):
+                if not current_key:
+                    raise IndexError("Internal error: indented output line"
```
edsantiago: Fixed. Removed "Internal error" from the string. Changed to `SyntaxError` instead; it seemed most appropriate than RuntimeError.

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,104 @@
...
+        a further set of key/value pairs. Return both.
+        """
+        if key in self.table:
+            return self.table[key]
+        return None
+
+    def sub_fields(self, key):
+        """
+        Fetcher for multi-valued docker info elements. Always returns
+        a dict, even if empty.
```
edsantiago: I guess I didn't communicate well enough the last time.

__Every__ `docker info` output line is of the form Key: Value, where Key and Value are strings. Always. That makes it natural to access docker info via accessors like `get` or dict lookup.

__Some__ `docker info` elements include an indented set of further Subkey: Value pairs; these are associated with a top-level Key such as Storage Driver, Plugins, Containers. Again, only some. 

The principal use of DockerInfo() is for a caller to ask: what is the value of DockerInfo(X), for X being Storage Driver, Docker Root Dir, Registry, whatever. Thus the principal accessing mode should make this easy. If the principal access mode returns a dict, callers will have to know how to access the principal Interesting Thing. To put it in code terms:

    now:  if DockerInfo()['Storage Driver'] != 'devicemapper' ...
    ugh:  if DockerInfo()['Storage Driver']['__stringvalue'] != 'devicemapper' ...

I want to make the standard case easy, and the sub-keys case accessible to those who need it. I don't want to make the standard case clunky. Does that make sense?

dockertest/output/dockerinfo_unittests.py
```diff
@@ -0,0 +1,106 @@
...
+                'Network': 'host null bridge overlay',
+            },
+            'ID': '5UCC:ANAG:4BIE:2KK6:VGP4:XKVO:5HB5:RPOA:PFLJ:HXRF:FCDV',
+            'Insecure Registries': '',
+            'Insecure Registries...': {
+                '127.0.0.0/8': '',
+            }
+        }
+
+        self.maxDiff = None
```
edsantiago: Fixed: added a comment.

subtests/docker_cli/info/info.py
```diff
@@ -35,82 +33,32 @@ class info(subtest.Subtest):
...
-                                     " '%s' from docker info with no previous"
-                                     " unindented lines.", line)
-                if current_key not in out:
-                    out[current_key] = {}
-                out[current_key][key] = value
-            else:
-                out[key] = value
-                current_key = key + '...'
-        return out
+        self.stuff['cmdresult'] = DockerInfo()
```
edsantiago: Fixed. Good catch, thanks.

subtests/docker_cli/info/info.py
```diff
@@ -35,82 +33,32 @@ class info(subtest.Subtest):
...
-        info_map = self._build_table(outputgood.stdout_strip)
+        super(info, self).postprocess()
+        info_table = self.stuff['cmdresult']
 
         # We support multiple storage drivers. Each one has a different
         # set of key/value settings under 'info'; so each one has a
         # dedicated helper method for validating.
-        storage_driver = info_map['Storage Driver'].lower()
+        driver_name = info_table['Storage Driver']
+        if driver_name is None:
```
edsantiago: Fixed. Good catch.

subtests/docker_cli/info/info.py
```diff
@@ -35,82 +33,32 @@ class info(subtest.Subtest):
...
+            raise DockerTestFail("'docker info' did not return"
+                                 " a value for 'Storage Driver'")
+        self.loginfo("Storage Driver = %s", driver_name)
```
edsantiago: Yeesh. That would be a lot of output, most of it unuseful. My purpose here is to recognize that `info.py` cares greatly about the Storage Driver aspect and runs different tests depending on its value. A person debugging the output would find it helpful to see that value, especially if there's a bunch of "Data file" / Metadata file errors but the system is supposed to be configured with overlay2.

subtests/docker_cli/info/info.py
```diff
@@ -35,82 +33,32 @@ class info(subtest.Subtest):
...
         try:
-            handler = '_postprocess_' + storage_driver
-            getattr(self, handler)(info_map['Storage Driver...'])
+            handler = '_postprocess_' + driver_name.lower()
+            getattr(self, handler)(info_table.sub_fields('Storage Driver'))
         except AttributeError:
-            raise DockerTestFail("Unknown storage driver: %s" % storage_driver)
-        except KeyError:
```
edsantiago: I don't understand this. I removed the `KeyError` handler because I consider this possibility too unlikely, and not worth the clutter. If it ever happens IRL, the test will fail anyway. Does that address your concern?

cevich:
[(2) (CN) Deferred Deletion ADEPT peons](https://trello.com/c/TDuEhBtZ/475-2-cn-deferred-deletion-adept-peons)

dockertest/output/dockerinfo.py
cevich: Uggg, PEBUAK, my bad.  You **SHOULD** call ``super(...)__init__()`` (first draft did not).

dockertest/output/dockerinfo.py
cevich: Blah, terrible choice of words...let me explain another way...

Having ``info_string`` optional and ``_build_table()`` private seems to defeat the purpose of defining a new type of thing.  If a class is used, then obtaining the data, and processing it, and presenting it shouldn't be tightly-coupled to creation.  It makes it very difficult to sub-class w/o re-writing most of the entire thing, so why have a class in the first place?

e.g. If I want to subclass ``DockerInfo`` to parse the output in a different way, because of different command-line options, I'd have to override ``__init__()``, `` _build_table()``, and maybe even ``sub_fields()``.  Basically the entire class.

(Small Mountain):  I'm not in a hurry, but if you are...maybe this "class" should really just be a function?

(Challenging Expedition):  To grow, and become powerful in the ways of the For^H^H^H OOP...the presentation-aspect shouldn't be tightly-coupled to both creation and processing.  Kind of like a **miniature** version of the MVC-like patterns: [MVP](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter) or [MVA](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93adapter)

e.g.: (one possibility)

```python

class DockerInfo(object):
    """
    Parser of 'docker info' output
    """

    def __init__(self, info_string):
        self.info_string = info_string  # The model

    @classmethod
    def MakeDockerInfo(cls, docker_path=None):
        return cls(subprocess.check_output...))

    def build_table(self):  # The adapter or presenter
        ...

    def __getitem__(self, key):  # The view
        ...
```

Now the data is close to all operations, yet I'm free to subclass and alter any individual aspect w/o needing to completely re-write the others.

Does that help/make more sense?

dockertest/output/dockerinfo.py
cevich: When you took notice of the excessive number of comments/docs needed for some part of ADEPT, it raised the "too complex" red-flag for me.  Take a step back, look at how much documenting is needed here, and think how the code can be simpler.

> Key and Value are strings. Always

yet

> Some docker info ... set of further Subkey

There are no overlapping keys, so why not just represent it that way?  e.g.

```python
def build_table(self):
    table = {}
    for line in self.info_string.splitlines():
        if ': ' in line or line.endswith(':'):
            (key, value) = [e.strip() for e in line.split(':', 1)]
        else:
            key = line.strip()
            value = ''
        key = key.lower().replace(' ', '_')  # All same-same
        if key in table:
            raise KeyError("Duplicate key %s detected, shoot Chris" % key)
        table[key] = value
    return table
```

Then, if you want to make callers life easier for some cases, add a method (or function):

```python
dm_keys = ('pool_name', 'pool_blocksize', ...)

def devicemapper_table(self):
    return dict(zip([(k, self[k]) for k in self.dm_keys]))
```

If in the future I get shot...well...then I won't really much care what happens from the grave 😀

cevich: err, ``zip`` isn't needed.
    ``return dict([(k, self[k]) for k in self.dm_keys])``

subtests/docker_cli/info/info.py
cevich: True, if it's a key error it will show the value of the key in the exception.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
+r"""
```
cevich: Gahhh, sorry, somehow didn't notice this new test code...

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        storage_driver = dockerinfo['Storage Driver']
+        if storage_driver != 'devicemapper':
+            raise DockerTestNAError("This test is only applicable when using"
+                                    " the devicemapper storage driver; host"
+                                    " system is using %s" % storage_driver)
+
+        docker_cmdline = dockertest.docker_daemon.cmdline()
+        if "dm.use_deferred_deletion=true" not in str(docker_cmdline):
+            raise DockerTestNAError("deferred deletion not enabled in"
+                                    " docker daemon command-line options.")
```
cevich: Hrmmm, two ways to skip isn't too bad, but these checks could easily grow in number and complexity.  Since they throw a "skip" it could easily cause us to miss a failure where the test should have run, but didn't because some complex set of skip-checks broke (for some future case).

What you have here for now is fine, it just makes me think...

...off-topic: It seems we could use a  special set of sub-tests that are simply "disabled by default".  Another example is run_volumes (which we just skip everywhere ATM).  Maybe we could have. a "subtests/special" that we don't even look at (in ``control`` file) unless they're referenced explicitly (by ``include`` and/or ``subthings``). Then none of those subject tests would need to worry about system-setup, restoration, or skip-checks.  We just allow them to fail naturally because they're only run when specifically requested by human or automation - implying setup was already done.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        Stop our idle container, by removing its semaphore file.
+        Wait until container is truly gone, as reported by docker ps -a
+        """
+        os.unlink('DELETE-ME')
+
+        dc = self.stuff['dc']
+        c_name = self.stuff['container_name']
+        c_gone = lambda: dc.list_containers_with_name(c_name) == []
+        self.failif(not utils.wait_for(c_gone, timeout=5),
+                    "Timed out waiting for container to exit")
```
cevich: How about attempting to kill this guy if it exists, during ``cleanup()``?

(otherwise, letting ``garbage_check`` grab it is probably okay)

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @property
+    def _container_mountpoint(self):
+        """
+        Given a container name, return its host-accessible root filesystem
+        """
+        subargs = ['--format', '{{.GraphDriver.Data.DeviceName}}',
+                   self.stuff['container_name']]
+        inspect_cmd = mustpass(DockerCmd(self, 'inspect', subargs).execute())
+        devmap_entry = inspect_cmd.stdout.strip()
```
cevich: Wheel reinvention? How about (not this dense):
``DockerContainers(self).json_by_name(self.stuff['container_name'])['GraphDriver]['Data']['DeviceName']``

Then we don't have to worry about ``docker inspect`` returning ``<no value>`` (it doesn't exit non-zero for all missing data).

cevich: (updated comment above, forgot the [0])

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        os.chdir(self._container_mountpoint)
+
+        # Determine baseline count of deferred deletions. We expect 0,
+        # but let's not panic if it isn't.
+        defer_count_initial = self._defer_count()
+        if defer_count_initial != 0:
+            self.logwarning("Deferred-Delete count from 'docker info'"
+                            " is %d (I expected 0). This may mean your"
+                            " system has still-undeleted containers.",
+                            defer_count_initial)
```
cevich: ``garbage_check`` would take care of this...if you're looking for some lines to delete 😀

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+            raise DockerTestNAError("deferred deletion not enabled in"
+                                    " docker daemon command-line options.")
+
+    def run_once(self):
+        super(deferred_deletion, self).run_once()
+
+        self._start_idle_container()
+
+        # cd into its directory. This prevents devicemapper from reaping it.
+        os.chdir(self._container_mountpoint)
```
cevich: Impatient, short-tempered future maintainers may become frustrated that  ``_container_mountpoint`` doesn't appear when they scroll up.

Since it's only referenced once, does it really make sense to be a property?  Would it be simpler to ``self.start_idle_container()`` from ``initialize()`` and then ``self.stuff['container_mountpoint'] = self.idler_mountpoint()``?

I dunno, it's okay either way, okay, forget I even brought it up, damn, too late.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        defer_count_initial = self._defer_count()
+        if defer_count_initial != 0:
+            self.logwarning("Deferred-Delete count from 'docker info'"
+                            " is %d (I expected 0). This may mean your"
+                            " system has still-undeleted containers.",
+                            defer_count_initial)
+        self.stuff['defer_count_initial'] = defer_count_initial
+
+        # Tell the container to exit
+        self._stop_idle_container()
```
cevich: This should probably also go into ``cleanup()``, in case something above breaks.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        super(deferred_deletion, self).postprocess()
+
+        defer_count_initial = self.stuff['defer_count_initial']
+        self.failif_ne(self._defer_count(),
+                       defer_count_initial + 1,
+                       "Deferred-Delete count while cd'ed into container")
+
+        # cd out of the container filesystem. devmapper will notice and
+        # will clean up, but it may take a while. Allow up to 30 seconds.
+        os.chdir('/')
```
cevich: This needs to go into ``cleanup()``

cevich: (also)

subtests/docker_cli/deferred_deletion/deferred_deletion.py
cevich: (note made of "off-topic" idea)

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        return mountpoint
+
+    @staticmethod
+    def _defer_count():
+        """
+        Returns the integer count of deferred deletions, as obtained
+        from docker info. We assume DockerInfo() is never cached.
+        """
+        storage_fields = DockerInfo().sub_fields('Storage Driver')
+        return int(storage_fields['Deferred Deleted Device Count'])
```
cevich: Okay, looks good'nuf.  I do think there should be a minimal ``cleanup()`` that tries some sort of half-arse job of recovering.  So, maybe if ``cd /``, ``rm -f DELETE-ME`` (ignore errors), ``self.stuff['dc'].clean_all()``?  Then it doesn't assume ``garbage_check`` is even enabled or running.

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        return {}
+
+    def __getitem__(self, key):
+        """
+        Convenience method: allows treating results as a dict
+        """
+        return self.get(key)
+
+    def __repr__(self):
+        return '{}("{}")'.format(self.__class__.__name__, self.info_string)
```
cevich: I don't mean to be a hardass about adding API stuff.  Rather it's that in my experience, it's _far_ easier on future maintainers if it's done "flexibly" the first time, rather than try and duct-tape it later.  Consider, "later" may come when there are many more dependent callers.  I know you know, the burden of fixing library code AND all it's callers, and then testing all that, and hoping you never have to see the code again.

If instead there's at least a small level of indirection, double-buffering, and/or getters/setters type things, then it's much easier to hide future fixes w/o having them affect the callers.  Doesn't have to be perfect or cover 100%, just the big "could-change" stuff.  It also supports the rare (but tricky) cases where a test needs to sub-class and override a little part, to do something special that will probably never need to ever happen elsewhere.

This is also why there's much more strict pylint checking on ``dockertest`` over than subtests.

cevich:
Okay, review is really done now 😀 (sorry I totally missed the actual test the first time).   It's a second or third-draft right out the gate, unlike the sub-beta-alpha 0th-draft crap I toss at you 😀

I've got a test-job all lined up and almost ready so we can do the setup in ADEPT and see how this (and other) tests fare in Jenkins land.

cevich:
jenkins, test this please.

cevich:
...okay, [how about this compromise(whole file - untested)](https://gist.github.com/cevich/e95e738b853ce1aeb5f2f4a5b83f081f):

- API is unchanged (keep the class vs re-write as function)
- Parsing logic is unchanged
- Parsing is much less viscus (more modular)
- Instance and parsing is stateful and cached
- I'm not married to any names or new unittests

dockertest/output/dockerinfo.py
cevich: [...continued](https://github.com/autotest/autotest-docker/pull/715#issuecomment-299876300)

edsantiago:
New diffs submitted. This PR has grown beyond the scope of sane review, so I'm trying to address all review comments in this one response.

> [ polymorphic, class, function, MVP, etc ]

I just cribbed from `dockertest/output/dockerversion.py`. I was just trying to keep things simple. There is no reason for anyone ever to subclass this nor dockerversion.py. It's just an attempt to run 'docker info' and provide that to interested parties. That said, I incorporated your 'compromise' suggestions. I also cleaned up the usage so there's only one accessor, `get()`, invoked with one or two elements. I hope that's ok.

> calling super.__init__

`dockerversion.py` doesn't. If I add a super to this one, I should probably add it to dockerversion.py. I would prefer not to do it to either, because these really are just functions shoehorned into OO form. They don't need complicated OO framework.

> flattening the docker info results

Sorry, I can't do that. I believe it would lose valuable information. And then trying to undo that by adding a hardcoded 'dm_keys' list -- that would introduce far more complexity, fragility, duplication than I can justify.

> skipping

Yes; this is a hard problem. I'm not convinced that ADEPT solves it, because as we keep finding out ADEPT configs are fragile and hard to maintain. Tests get enabled/disabled with little visibility. For now I prefer using TEST_NA because it makes future analysis easier, i.e. there's a record of when a test fails to run and why.

> dc.json_by_name()

Interesting. Thanks for the pointer.

> "garbage_check would take care of this"

I think there's a misunderstanding somewhere. I don't see how garbage_check could possibly have anything to do with any of this. The situation I'm warning about here is an unlikely one: deferred count being nonzero at the start of this test. I think it would mean that there are deleted containers whose mount space is still in use. Maybe a failed deferred_deletion from a previous test. I don't know. It just seems worth bringing to the attention of a test reviewer.

> container_mountpoint()

It's tempting to move this into `_start_idle_container()`. It would save duplication of dc and c_name. But the cost is increasing the complexity of `_start_idle_container`. Unrelated functionality. I prefer small easy-to-understand functions. I choose to leave it as-is.

> cleanup()

Whoops. Added. Thanks.

cevich:
Oh, fantastic Ed, I'm really happy to see you've carried this forward!  I was afraid you'd collided with an stationary object.  I'm happy to see that's not the case...

> grown beyond the scope of sane review

...I've had __way__ too much coffee this morning.  I'm recognizing my inability to respond in a useful way that's not likely to make that (scope) problem worse 😖

Don't block on me though, feel free to start something else (or play) until I can get my mind right to get this finalized and merged.

ohhh! look! Free Starbucks!

/me slaps sense into himself

Food first...maybe that'll help...

cevich:
> New diffs submitted. This PR has grown beyond the scope of sane review, so I'm trying to address all review comments in this one response.

Okay cool, I don't mind either way.  I read via osmosis, so *shrug*, whatever easier to write for you 😀

> I just cribbed from dockertest/output/dockerversion.py. I was just trying to keep things simple. There is 

Aye, and I think that dockerversion class is/was designed badly, so didn't want to repeat those mistakes.

> no reason for anyone ever to subclass this nor dockerversion.py. It's just an attempt to run 'docker info' and provide that to interested parties. 

No reason you can think of now...hence, why I suggested it should be at least a little OOPie, in case someone needs it later.  But I gotcha, and agree, no need to go crazy.

> That said, I incorporated your 'compromise' suggestions. I also cleaned up the usage so there's only one accessor, get(), invoked with one or two elements. I hope that's ok.

That's fantastic!  I can't wait to see...

> 
>     calling super.init
> 
> dockerversion.py doesn't. If I add a super to this one, I should probably add it to dockerversion.py. I would prefer not to do it to either, because these really are just functions shoehorned into OO form. They don't need complicated OO framework.

It doesn't?  Damn, it's more broken that I thought 😦 .  It's not _that_ important for Object() derivatives, unless/until a subclass does multi-inheritance.  Which isn't a problem now, and can easily be fixed by a ``#FIXME`` - which **is** important if somebody decides to crib that class again.  I'll take care of it...okay, done.

> Sorry, I can't do that. I believe it would lose valuable information. And then trying to undo that by adding a 

Okay, you seem to feel it's really important, so I can flex here, and trust your judgment.

> 
>     skipping
> 
> Yes; this is a hard problem. I'm not convinced that ADEPT solves it, because as we keep finding out 

It has to solve this, that's it's job.  At the same time, I'm loath to work on "fragility" because:

* It needs to be a "team" effort (working on that)
* Simplified ADEPT is getting really close (and needs team input/work)
* Project (name starts with "F") is my current priority (per management)

> For now I prefer using TEST_NA

Oh, don't get me wrong, we need ``TEST_NA``  **ALSO**.  Visibility is important **and** ADEPT needs to do it's job because Docker Autotest is a _horrible_ (high level) system-automation platform.  This is a solvable problem, we're working on it...

>     "garbage_check would take care of this"
> 
> I think there's a misunderstanding somewhere. I don't see how garbage_check could possibly have anything to do with any of this. The situation I'm warning about here is an unlikely one: deferred count being nonzero at the start of this test. I think it would mean that there are deleted containers whose mount space is still in use. Maybe a failed deferred_deletion from a previous test. I don't know. It just seems worth bringing to the attention of a test reviewer.

Nope we have understanding.  That comment was more in relation to my other comment about ``cleanup()`` (in the test) that you've answered below.  No worries...

> 
>     container_mountpoint()
> 
> It's tempting to move this into _start_idle_container(). It would save duplication of dc and c_name. But the cost is increasing the complexity of _start_idle_container. Unrelated functionality. I prefer small easy-to-understand functions. I choose to leave it as-is.

okay, np, was just an idea/suggestion (mine are ultra-cheap/disposable).  I'm super flexible about test-code, even in non-python, so long as it's not completely bonkers or dangerous.

> 
>     cleanup()
> 
> Whoops. Added. Thanks.

Cool...lessee, sounds very promising...

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    Storage Driver -> storage_driver
+    """
+    return key.lower().replace(' ', '_')
+
+
+class DockerInfo(object):
+    """
+    Parser of 'docker info' output
+    """
```
cevich: Because of the way the docs-parser works, the ``__init__()`` arguments should go into the class docstring.  I'm not going to block the merge on having this though...

dockertest/output/dockerinfo.py
```diff
@@ -0,0 +1,114 @@
...
+        normalized_key = _normalize(key)
+        if sub_key is None:
+            return self.info_table[normalized_key]
+        sub_fields = self.info_table[normalized_key + '...']
+        if len(sub_key) == 0:
+            return sub_fields
+        return sub_fields[_normalize(sub_key)]
+
+    def __repr__(self):
+        return '{}("{}")'.format(self.__class__.__name__, self.info_string)
```
cevich: Okay, good, good, let's run with this version...

dockertest/output/dockerinfo_unittests.py
```diff
@@ -0,0 +1,126 @@
...
+ Pool Blocksize: 524.3 kB
+ Data file: %s
+ Library Version: 1.02.135-RHEL7 (2016-11-16)
+Logging Driver: journald
+Plugins:
+ Volume: local lvm
+ Network: host null bridge overlay
+ID: 5UCC:ANAG:4BIE:2KK6:VGP4:XKVO:5HB5:RPOA:PFLJ:HXRF:FCDV
+Insecure Registries:
+ 127.0.0.0/8""" % ''
```
cevich: Curiosity:  Why the ``127.0.0.0/8""" % ''`` (empty substitution token) on assignment?  (not a merge-blocker)

dockertest/output/dockerinfo_unittests.py
```diff
@@ -0,0 +1,126 @@
...
+                         expect['containers...']['paused'],
+                         'subelement accessor: single element')
+
+        # reproducer
+        self.assertEqual(str(actual), 'DockerInfo("%s")' % info_string,
+                         "string reproducer")
+
+
+if __name__ == '__main__':
+    main()
```
cevich: okay, this is fine too (not looking closely since the unittest passes)...

dockertest/output/unseenlines_unittests.py
```diff
@@ -4,12 +4,44 @@
...
+setattr(mock('autotest.client.shared.error'), 'TestFail', Exception)
+setattr(mock('autotest.client.shared.error'), 'TestNAError', Exception)
+setattr(mock('autotest.client.utils'), 'PlaceHolder', Exception)
 
 
 class UnseenLinesTestBase(unittest.TestCase):
 
     def setUp(self):
-        from UnseenLines import UnseenLines
+        from dockertest.output import UnseenLines
```
cevich: Oops...thanks!  🥘

run_unittests.sh
```diff
@@ -7,7 +7,7 @@ export PYTHONPATH=$(dirname $0):$PYTHONPATH
 
 echo ""
 echo ""
-for unittest in dockertest/*_unittests.py
+for unittest in $(find dockertest -name '*_unittests.py')
```
cevich: Ooooohhh, that's why.  Good catch!

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,160 @@
...
+        from docker info. We assume DockerInfo() is never cached.
+        """
+        return int(DockerInfo().get('storage_driver',
+                                    'deferred_deleted_device_count'))
+
+    def cleanup(self):
+        super(deferred_deletion, self).cleanup()
+        os.chdir('/')
+        if 'container_name' in self.stuff:
+            self.stuff['dc'].clean_all(self.stuff['container_name'])
```
cevich: good good...

dockertest/output/dockerinfo_unittests.py
edsantiago: It gets inserted here:

    Data file: %s

Without that `%s` there would be a trailing space at the end of this line, and my editor doesn't allow trailing whitespace (neither does git, if you configure it correctly).

cevich: Oooohhh, okay, so it's actually testing something useful then, gotcha.  Couldn't tell if it was a typo or not (they all look the same to me) hahhah 😀

PR #721, opened 2017-05-19
deferred_deletion test: rewrite it

edsantiago:
Invalid assumption: we cannot rely on MountFlags=slave
being absent from docker systemd unit file. That's
only necessary for liverestore, not for deferred deletion.

Deferred deletion still works with MountFlags=slave,
we just can't access the container's filesystem from
inside the host. We have to nsenter docker's mount
namespace. And to do that we need a helper script.

Also: add one more prerequisite: check for the existence
of the sysctl knob file, and N/A if it's missing. This
covers the RHEL 7.2 case in which docker-storage-setup
optimistically and blindly enabled deferred_deletion.


subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -107,55 +111,17 @@ def _start_idle_container(self):
...
-        """
-        return int(DockerInfo().get('storage_driver',
-                                    'deferred_deleted_device_count'))
-
     def cleanup(self):
         super(deferred_deletion, self).cleanup()
-        os.chdir('/')
         if 'container_name' in self.stuff:
-            self.stuff['dc'].clean_all(self.stuff['container_name'])
+            self.stuff['dc'].clean_all([self.stuff['container_name']])
```
cevich: You caught me on this one, so now, *tag* your it!

Gotta do something like:

```python
if 'dc' in self.stuff and self.config['remove_after_test']:
    self.stuff['dc'].clean_all(self.stuff.get("container_name", []))
```

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -58,45 +71,36 @@ def initialize(self):
                            "docker daemon command-line options",
                            DockerTestNAError)
 
+        # FIXME: what if future kernels enable by default and remove the knob?
```
cevich: Won't happen in RHEL, once it's added, it's there for the entire release.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -58,45 +71,36 @@ def initialize(self):
...
+        sysctl_knob = "/proc/sys/fs/may_detach_mounts"
+        self.failif(not os.path.exists(sysctl_knob),
+                    "sysctl knob %s does not exist; this system is"
+                    " not likely to support deferred deletion" % sysctl_knob,
+                    DockerTestNAError)
```
cevich: This _could_ happen if ``/proc`` isn't mounted for some reason, so no prob. w/ checking.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -9,25 +9,37 @@
...
 ----------------------
 
 #. Run a docker container
-#. Find out the local (host) mountpoint of its root filesystem. cd into it.
-#. Read the Deferred-Deletion count from 'docker info'.
-#. Remove the container.
-#. Read the Deferred-Deletion count; confirm that it has grown by one.
-#. cd back out of the container rootfs
-#. Read the Deferred-Deletion count; confirm that it has gone back down.
+#. Sigh. Run a test script inside docker mount space that will:
```
cevich: No need to "Sigh" in documentation 😆   People probably won't know what you're upset about in that context.  As for the test, also no need, this is a perfectly valid strategy.  It also provides a good model for doing these kinds of tricky things for testing other aspects.

subtests/docker_cli/deferred_deletion/test-deferred-deletion.sh
```diff
@@ -0,0 +1,86 @@
...
+cd /
+
+wait_until=$(expr $SECONDS + 30)
+while [ $SECONDS -le $wait_until ]; do
+    now_count=$(deferred_count)
+    if [ $now_count -eq $starting_count ]; then
+        exit 0
+    fi
+    sleep 0.1
+done
```
cevich: This could be done in one line with ``( sleep 30s && kill $$ ) &``, no?

Not critical, what you have here is fine.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
edsantiago: Good to know, thanks!

subtests/docker_cli/deferred_deletion/deferred_deletion.py
edsantiago: I'm confused. Under what circumstances would I not want to remove the container? And, under what circumstances would "container_name" be in `self.stuff` but not "dc"?

subtests/docker_cli/deferred_deletion/test-deferred-deletion.sh
edsantiago: I don't think I understand that. In particular I don't see how it would give a useful error message on timeout; I'm also not sure it's more readable. Maybe I'm missing larger context.

cevich:
(Canceled ADEPT run from ``fixup!``).  CI Results are clean here, ignore that ``rhel-deferred-deletion-7.4.subtests.docker_cli.run_flags`` failure.  That peon will be going away soon.

subtests/docker_cli/deferred_deletion/test-deferred-deletion.sh
cevich: Nope, it's not more readable, what you have is fine.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
cevich: Sorry for the confusion.  the ``self.config['remove_after_test']`` support is a convention followed by every other test.  It's there to support runs of multiple tests over and over to check for leaks and to assist in debugging runs of an individual tests where you need to inspect the resulting container/image.  Esp. e.g. in this case, if your script doesn't terminate properly or you need to get into the container to see what's happening.

``container_name`` can be in ``stuff`` w/o ``dc`` due to future maintainer being a human, changing something, and forgetting to fixup ``cleanup()``.

This way is more forgiving of mistakes, but I'm also not going to make a fuss about it.  Checking for ``dc`` is really the only critical thing.

cevich:
Killing ADEPT CI test for this and merging...

Thanks Ed!

BTW: Consider adding (or as Dave) a Trello card for this work.  Even if it's a one or two-pointer, they do add up at the end of the sprint.

edsantiago:
Trello card: done. Thanks for the reminder.

PR #724, opened 2017-05-31
add test for docker user namespace remap

Edward5hen:
Request is from: https://trello.com/c/FLk16Cu7/457-ci-add-tests-for-userns-remap
The original acceptance wants to test file owners inside and outside the container.
But this feaure supporting user namespace remapping is designed to provide an
isolated user env, the files mounted from host will not be accessible. So I change
the test points to verify the process owner, which is often seen in most user guides.


Edward5hen:
[(?) [CI] Add tests for --userns-remap](https://trello.com/c/FLk16Cu7/457-ci-add-tests-for-userns-remap)

cevich:
@Edward5hen Awesome, thanks, adding you to the bot's whitelist...

cevich:
bot, add author to whitelist

cevich:
@Edward5hen okay there you go, ``Red Hat CI`` will run for you in addition to ADEPT (next time you push).  Please look at & fix the things ``Red Hat CI`` found.  Thanks again.

cevich:
Oh, I'm happy to help too if you need it, just ask.

subtests/docker_cli/namespaces/user_namespace.py
```diff
@@ -0,0 +1,138 @@
...
+        # Add subordinate uid and gid range
+        self._change_sub_uid_gid('add')
+
+        # Set max_user_namespaces
+        self._change_max_userns('add')
+
+    def _change_sub_uid_gid(self, action):
+        if action == 'add':
+            if os.path.exists('/etc/subuid'):
+                shutil.copyfile('/etc/subuid', '/etc/subuid.bk')
```
edsantiago: The standard way to do this is `os.rename`: that preserves mtime, mode, owner, and all file attributes. Then, when finishes, `os.rename` it back into place.

cevich:
@Edward5hen if you need anything done test-system-setup wise, the place for that is in ADEPT, and I'm happy to add what you need, just ask.

Edward5hen:
@edsantiago Thanks for your review and the suggestion.
I thought we would like to test the basic function of this feature, so I changed the test steps to verify the process owner which might be more clear and most user guides tell. Sorry about the misread. For strictly follow the acceptance mentioned in the card, how about below steps?
> The docker daemon successfully restarts after adding ``--userns-remap=<name>`` to OPTIONS in /etc/sysconfig/docker or docker-latest

* `echo “dockremap:100000:65536" > /etc/subuid`
* `echo "dockremap:100000:65536” > /etc/subgid`
* `echo 1 > /proc/sys/user/max_user_namespaces`
* add OPTION `--userns-remap=default` in `/etc/sysconfig/docker`
* `systemctl restart docker`

> Within a container, files from the host appear to be owned by root.

* `mkdir /test`(to avoid [rhbz1375880](https://bugzilla.redhat.com/show_bug.cgi?id=1375880))
* `chown 100000:100000 /test`
* `touch /test/from_host.txt`
* `chown 100000:100000 /test/from_host.txt`
* `docker run --rm -v /test:/test centos ls -l /test`

> On the host, files from the container owned by the remapped ID

* `docker run --rm -v /test:/test centos touch /test/from_ctn.txt`
* `ls -l /test`

I think above steps should work for any linux distro that supports user namespaces, no matter which storage is used, like [here](https://github.com/moby/moby/issues/21130) said.  But unfortunately, when I tried this with the latest RHEL7.4, it fails at checkpoint 2. If I misunderstand something, please point it out. 
Btw, do you have some unclean way with docker-storage-setup for this? Thanks.

@cevich thanks for your careful reminds!

cevich:
> I thought we would like to test the basic function of this feature

Good idea.

> docker daemon successfully restarts

Remember for ``cleanup()``, must recover from failure at any time, even "in-between" lines.  Make sure you keep backup copies of modified files, "move" changed files into place (instead of in-place edit), and remember there's no recovery from any failures after ``cleanup()``.

Also, I suggest reading the initial value of ``/proc/sys/user/max_user_namespaces`` and restoring that in ``cleanup`` instead of just blasting a ``0`` into it (that assumes it _was_ zero to begin with).

> docker or docker-latest

``dockertest.docker_daemon.which_docker()`` will tell you this.

> ``“dockremap:100000:65536"``

That (and the ``subgid``) numbers (from and to) would make excellent configuration options.  Then you can reference the "from" configuration instead of hard-coding '100000' everywhere.

> ``mkdir /test``

I suggest creating this as ``os.mkdir(os.path.join(self.tempdir, 'test'))``  That parent directory (``self.tempdir``) is guaranteed to exist, and also get automatically cleaned up after the test.

In general, your plan seems viable and probably about as simple as you can make it (it's a challenging subject on it's own).

cevich:
@Edward5hen After speaking with the team, there's another BIG gotcha here.  @edsantiago said he mentioned it as well (but I can't find the comment):  Before you're ``systemctl restart docker`` step, the ``/var/lib/docker`` and/or ``/var/lib/docker-latest`` need to be wiped out, and storage re-configured with ``container-storage-setup``  😖   Both of which are out-of-scope for Docker Autotest.

Probably the easiest way forward here is if the test is disabled by default, and we coordinate configuring and running it under a special environment / ADEPT job.  Similar to what we do for ``liverestore`` testing.  While initially that means it'll be an expensive test to run, it does present a significant opportunity for adding related tests (to run under the same environment).

That said, it's also up to you if/how you want to continue forward.  I'll certainly help as much as I can with the ADEPT-side of things, creating a special job and what not.  Nobody likes to say it, but another possibility is to abandon this, leaving the card for "later".  Then take up a simpler/easier new-test card in the meanwhile.

Let us know what you'd like to do.

cevich:
**Re: additions to ADEPT:**  The easiest thing would be to just do _all_ the setup steps in an ADEPT role, then enable this one test to run.  So your code (when test is enabled) can just assume the system is already setup and configured for userns, and the values from it's ``config_file.ini`` are correct/match the system.

Edward5hen:
@cevich Currently, I have known nothing about **ADEPT**(sad to say). I find the project you own, the documentation seems few. Is there some other place I can study this tool/platform? Or if it takes much time to study, maybe abandon this PR and take up a simpler new test is a better choice.

cevich:
>  I find the project you own, the documentation seems few

No kidding :D  I'm working on (first-draft almost finished) a updated and simplified version that will have documentation.  For now, care/feeding of legacy ADEPT is on my shoulders, so literally whatever you need it to do, I'll simply make that happen.

Even completely ignoring ADEPT for now, I agree with what Ed said at the beginning...this is a very difficult subject and test to tackle all on it's own.  I think there's no shame in closing this for now (it can always be re-opened later, by you or anyone else).  At the least, you are now much more aware of the complexity to test such things.  That will leave you better prepared for next time.

edsantiago:
@Edward5hen I'm sorry for replying so late. A few belated comments.

First: you don't need to know much about ADEPT. Basically it's just the provisioning system. docker-autotest is the set of tests, ADEPT is what creates a virtual host, installs and configures docker, and then runs docker-autotest. ADEPT knows how to create a RHEL 7.3 system, or a RHEL-7.4 with docker-latest, or (what @cevich was mentioning above) a RHEL 7.4 system with docker and with `/proc/sys/user/max_user_namespaces=100000` and with `/etc/sub[ug]id` files already created and with `--userns-remap=default` already in `/etc/sysconfig/docker`. That means your test would be much simpler: it doesn't have to do any of the complicated (and not-actually-working-the-way-you-think-it-is) setup and cleanup.

I think your test is a valuable one, and a good learning experience if you want to work on docker-autotest. I would encourage you to continue working on it. But as per the above, don't worry about setting up the environment: just tell @cevich what you need, and then concentrate on just writing the test code. You will want to add a check in `initialize()` to exit with `TestNA` (not applicable) if the right conditions aren't met; we can help you with that.

I hope this helps.

Edward5hen:
@edsantiago Thanks for your brilliant encouragement and explanation. Now I probably know what ADEPT is, it's an ansible playbook to help autotest-docker provision the environment stuff. I'm willing to contribute to docker-autotest. I'm reopening this and focusing on the test side.
So please @cevich help to add below environment setup things from ADEPT side:
* lets use rhel7.4 with docker or docker-latest
* make `/proc/sys/user/max_user_namespaces=100000`
* `/etc/sub[ug]id=dockremap:100000:65536`
* enable docker sysconfig option with `--userns-remap=default` and successfully restart the service.

Thanks a lot.

cevich:
>So please @cevich help to add below environment setup things from ADEPT side:

(Sent you an e-mail about this also)

Is this an urgent need for you?

My priority ATM is to complete 'simplified ADEPT' and document it.  I _can_ update legacy-adept now, but it's extra work to rebase those changes forward (into simplify).  There is a near-term plan to begin a transition (Red Hat internal) to simplified-ADEPT in the next few weeks.  That means my tertiary-priority is to have both (legacy and simplified) running and reporting, in parallel for a little bit (weeks).

Point being, I want to be sensitive to your testing needs, but would prefer to defer work on this for a few weeks, until Simplified ADEPT has entered transition.  OTOH, if this is a priority-test for you, let me know and I'll get the changes into legacy ADEPT made ASAP.

Edward5hen:
@cevich Like Dave said in the email, this is not urgent. We can hold off for a few weeks until the transition work is done:)

cevich:
jenkins, test this please

cevich:
Closing due to lack of activity.  Please re-open if this is important.

PR #726, opened 2017-05-31
Test running systemd in container.

adelton:


cevich:
@adelton okay, that's two PRs in one day...adding you to the whitelist 😀

cevich:
Now when you open a PR, we'll actually run all the tests from the PR's HEAD.  This is just for helping check the PR itself, and so it only runs on RHEL + RHELAH.  But it's good for getting a sense if things are expected.  You can also kick-off this job manually by commenting...

cevich:
jenkins, test this please.

cevich:
bot, add author to whitelist

cevich:
Debug log for your test: https://hurl.corp.redhat.com/67926d5

FWIW: I just followed the 'details' link, then went under 'Build Artifacts'.  Navigated down through ``results/<hostname>//docker/subtests/docker_cli/systemd_in_container.<number>``.

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,50 @@
...
+# and the first field is a security context (user:role:type:level); verify
+# that command generates exactly one line of output; extract the <type>
+# and <level> fields; and check against a list of one or more expected
+# values. Assume that there's no overlap between valid types & ranges.
+check_output() {
+    local name="$1"; shift
+    local command="$1"; shift
+
+    $command < /dev/ptmx > /tmp/systemd$$.log 2>&1 &
+    sleep 30
```
edsantiago: This adds 30 seconds to every autotest run, even if the container starts in two seconds. See below for a version that loops and exits more promptly. It's also inline.

```bash
...
image="$1"

# Start a container running systemd. Clean it up when we exit.
cid=$(docker run --detach -ti $image /usr/sbin/init)
if [ $? -ne 0 ]; then
    echo "FAIL: could not start docker container"
    exit 1
fi
trap 'status=$?; docker rm -f $cid >/dev/null; exit $status' 0

# Exit as soon as we see the following indication of success
expect="Started Journal Service"
wait_until=$(expr $SECONDS + 30)

while [ $SECONDS -le $wait_until ]; do
    if docker logs $cid | grep -q "$expect"; then
        echo "PASS"
        exit 0
    fi

    # Abort if container is no longer running
    status=$(docker inspect --format '{{.State.Running}}' $cid)
    if [ "$status" != "true" -o $? -ne 0 ]; then
        echo "FAIL: docker container exited unexpectedly"
        echo "docker logs:"
        docker logs $cid
        exit 1
    fi

    sleep 1
done

echo "FAIL: did not find expected string '$expect' in docker logs:"
docker logs $cid

exit 1
```

adelton:
> A few suggestions inline, and one overall: there's already a systemd test, did you consider making this a sub-subtest of that?

My understanding of that directory is that it's about running docker containers as systemd services. While what I try to test is running systemd within containers. It seems like a different exercise.

subtests/docker_cli/systemd_in_container/systemd_in_container.py
```diff
@@ -0,0 +1,50 @@
...
+from dockertest.output.dockerversion import DockerVersion
+
+
+class systemd_in_container(subtest.Subtest):
+
+    def initialize(self):
+        # See Prerequisites (above)
+        DockerVersion().require_server("1.12")
+        self.stuff['result'] = None
+        self.stuff['di'] = DockerImages(self)
```
adelton: Frankly, this is just a copy of subtests/docker_cli/selinux_labels/selinux_labels.py without me giving much thoughts to how things are being done. So whatever is the best practice these days, let's do it.

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo "Usage: $0 IMAGE-NAME"
+    exit 1
+fi
+image="$1"
+
+# Exit status. Default 0 (success), but set to 1 on any subtest failure
+rc=0
+
+# Main code. Run a command whose output is expected to be a single line,
+# and the first field is a security context (user:role:type:level); verify
```
adelton: Doh, this just proves it's a copy. ;-) Will push fix.

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,50 @@
...
+# that command generates exactly one line of output; extract the <type>
+# and <level> fields; and check against a list of one or more expected
+# values. Assume that there's no overlap between valid types & ranges.
+check_output() {
```
adelton: I hope to add many more tests like this for systemd in the future, this is just to get us started.

Is the container name an issue? We can make it a parameter to the function.

subtests/docker_cli/systemd_in_container/test.sh
adelton: I see your point. The problem is that we might want to check for multiple strings in the output and some of them might already be available and some not yet ...

cevich: Many other tests do something very similar by monitoring the container's logs for a keyword.  In fact, it's used so frequently, ``dockertest.output.wait_for_output()`` was written a loooooong time ago.   There's also ``dockertest.AsyncDockerCmd.wait_for_ready()`` for handling this.

Actually, that makes me think.  Maybe this would be easier if you split up the script by phase,  e.g. execute one from e.g. ``initialize()``, another from ``run_once()``, and maybe another from ``cleanup()``.  That would allow for more modular completion-testing of the previous phase, either from script(s) or w/in python (either way is fine).

BTW, we recognize python isn't everyone's cup-o-tea, doing this in bash or any other language available on an Atomic host is totally cool.  Happy to help add necessary hooks to the python module if needed as well.

subtests/docker_cli/systemd_in_container/systemd_in_container.py
```diff
@@ -0,0 +1,50 @@
+r"""
```
cevich: Something to think about:

This test (overall) seems much more related to confirming container-content and possibly behavior/interactions with the host.  I'm wondering if maybe it shouldn't get it's own subtest category entirely, so it/they are distinct from ``docker_cli`` specific tests.

I'm totally cool if you want to make a ``subtests/docker_containers/`` (or similar name) for this now or later.  Especially if you have plans to add related tests in the future.  Though I also understand this may be speculative work, to get a feel for how Docker Autotest works, and if it would work for your needs.  No problem there.

subtests/docker_cli/systemd_in_container/test.sh
edsantiago: >The problem is that we might want to check for multiple strings in the output and some of them might already be available and some not yet .

If so, the inner loop can be rewritten or refactored as a helper function. The important point is, please don't use `sleep 30`.

subtests/docker_cli/systemd_in_container/test.sh
edsantiago: >  Is the container name an issue? We can make it a parameter to the function.

If you look at my suggested clean rewrite, you'll see that container name is actually unnecessary. You can just run with `--detach` and get the CID. That way you don't have to worry about a subroutine improperly knowing the container name.

If you decide to go with a container name anyway, then yes, please, encapsulate it so that the subroutine finds out the name properly. And please remember to add a cleanup.

edsantiago:
>> there's already a systemd test, did you consider making this a sub-subtest of that?
>
> My understanding of that directory is that it's about running docker containers as systemd services. While what I try to test is running systemd within containers. It seems like a different exercise.

Point taken. I withdraw my comment. Thank you.

edsantiago:
Hi Jan, I hope you haven't abandoned this. I look forward to seeing your next iteration.

If I may be so bold as to make a suggestion: what would you think about searching for the string `\[FAILED\]` in the output, and failing if it's present? There's actually a great real-world example of this, running systemd in docker when user namespaces are enabled:

    [  OK  ] Reached target Local File Systems.
             Starting Rebuild Journal Catalog...
             Starting Load/Save Random Seed...
    systemd-random-seed.service: main process exited, code=exited, status=1/FAILURE
    [FAILED] Failed to start Load/Save Random Seed.
    See 'systemctl status systemd-random-seed.service' for details.
    Unit systemd-random-seed.service entered failed state.
    systemd-random-seed.service failed.
    [  OK  ] Started Rebuild Hardware Database.
    ...

cevich:
@edsantiago - FWIW, this is all "our" code, so there's nothing wrong with sending Jan a PR against the base for this one (``adelton:systemd_in_container``).  That will help encourage progress or even finish it for him (if you're so inclined).  Even if he doesn't accept it, or partially accepts it, it gives him code he can act-on which is generally less time-intensive than parsing English 😀

adelton:
New attempt after long pause -- sorry for the delay.

Also rebased on master -> 82ed4127894b654605acb48b82b2571950116ec7.

subtests/docker_cli/systemd_in_container/test.sh
adelton: Attempt without `sleep 30` is now in 82ed4127894b654605acb48b82b2571950116ec7.

subtests/docker_cli/systemd_in_container/systemd_in_container.py
adelton: I'm fine with either way as well. Whatever is easier for you guys to review. I don't hide the fact that this is edited copy of `subtests/docker_cli/selinux_labels/` so I left it next to it for a start.

adelton:
> If I may be so bold as to make a suggestion: what would you think about searching for the string \[FAILED\] in the output, and failing if it's present? There's actually a great real-world example of this, running systemd in docker when user namespaces are enabled:

Well, the question is whether it is a bug in docker (or SELinux policy on the host) or something which should be addressed in the images, and to what extend this autotest-docker should cover images issues as well. I've actually reported the `systemd-random-seed.service` issue:

* https://bugzilla.redhat.com/show_bug.cgi?id=1467263

so we are waiting for fixed image to be released.

Yes, eventually, when we see that things are clean in the images, we probably should check for no `FAILED` in the output, as well as checking `docker exec systemd$$ systemctl status | grep 'State: running'. My only worry is, when we see things regressed and new version of RHEL or Fedora image failing, will we be able to disable the check temporarily (so that such a bug does not hide other errors) and reenable it once fixed image is out? I am not really clear about the CI setup and practices.

adelton:
I can see "Red Hat CI" failing with

https://s3.amazonaws.com/aos-ci/ghprb/autotest/autotest-docker/82ed4127894b654605acb48b82b2571950116ec7.0.1502192178557251311/setup.log

which ends with

```
--> Processing Dependency: python-astroid >= 1.4.5 for package: python2-pylint-1.6.5-4.el7.noarch
--> Finished Dependency Resolution
Error: Package: python2-pylint-1.6.5-4.el7.noarch (epel)
           Requires: python-astroid >= 1.4.5
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

May I assume that is some fluke in the CI setup?

adelton:
@cevich, thanks for the review. I hope 82ed4127894b654605acb48b82b2571950116ec7 is a move in the right direction. I don't seem to be able to comment on your review comments directly in this GitHub UI.

adelton:
bot, test pull request once

cevich:
jenkins, test this please.

cevich:
@adelton sorry no response, our team is up in Westford for a f2f this week (i.e. multi-day meetings).  I'm happy to see you picking this back up though.  We'll give more/better feedback soon as we can.

cevich:
CI be bustid' (technical term)

cevich:
Note: bz 1438359 is breaking "Red Hat CI"

adelton:
But that bugzilla says that it had the fix pushed four months ago (in spite of its status still being MODIFIED).

cevich:
Hmmm, any better ideas?  It's still breaking for me, running CentOS + EPEL container in papr.  I even tried epel-testing 😦

cevich:
Turned on another PR tester, lessee if that one works...

cevich:
Other tester broken as well 😖 Will have to test manually when we get back from the F2F.

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,65 @@
...
+            break
+	fi
+    done
+
+    return $lrc
+}
+
+# Actual checks
+
+docker run -d --rm -ti --name systemd$$ $image /usr/sbin/init
```
edsantiago: `Conflicting options: --rm and -d`

Please remove --rm

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,65 @@
...
+container_id=$( docker ps -q -f name=systemd$$ )
+check_output "systemd$$" \
+             "systemd in container" \
+             "Set hostname to <$container_id>\." \
+             "Started Journal Service"
+docker rm -f systemd$$
+
+docker run -d --rm -ti --name systemd$$ -h www.example.test $image /usr/sbin/init
```
edsantiago: Same thing: remove --rm

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,65 @@
...
+        i=0
+	while [ $i -lt $timeout ] ; do
+            if docker logs "$container" | grep -q "$expect" ; then
+                echo "PASS $name = $expect"
+                break
+	    else
+	        if [ $i -lt $timeout ] ; then
+                    echo "Did not find the output in attempt $i, will retry"
+	        fi
+	    fi
```
edsantiago: Missing `i=$(expr $i + 1)`

I would also suggest adding `sleep 1`, because that would make it a real timeout (as it is, it's just a retry counter and it's possible for all ten iterations to finish before systemd has started.)

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,65 @@
...
+    for expect in "$@"; do
+        echo "Checking for $expect"
+        i=0
+	while [ $i -lt $timeout ] ; do
+            if docker logs "$container" | grep -q "$expect" ; then
+                echo "PASS $name = $expect"
+                break
+	    else
+	        if [ $i -lt $timeout ] ; then
+                    echo "Did not find the output in attempt $i, will retry"
```
edsantiago: This doesn't really add much to the log. Actually, since it takes several seconds for systemd to come up, it detracts greatly. I would suggest just removing this block entirely.

cevich:
> Better, but I feel surprised (and disappointed) that you never even tried to test this on a local installation.

I'm guilty here as well since it seems our Papr setup is broken (CentOS) and no (old) ADEPT jobs are running.  I for one would trust results from automated-ci much more than a local setup.

@edsantiago Do you have any suggestions at least on the Papr problem?  Check the logs, you'll see the issue right away.  Maybe we should switch it to use Fedora instead of CentOS?

edsantiago:
No good suggestions, sorry: I don't know my way round CentOS. But: is it possible to get pylint from a CentOS repo instead of EPEL? Is it possible to yum-install an explicit version? (That assumes that this problem is a new one caused by a recent update to EPEL pylint). Also on the assumption that a recent pylint update broke EPEL, is it likely to be caught and fixed soon?

A final personal opinion: CI is not the place to test or debug code. CI is the last gating point, and as developers we should strive to submit already-tested code.

adelton:
> Better, but I feel surprised (and disappointed) that you never even tried to test this on a local installation.

Sorry abou that. I did test it locally, just not the "negative" situation when things would fail and uncover the missing increment.

subtests/docker_cli/systemd_in_container/test.sh
adelton: Hopefully, with that `sleep 1` it might make a bit more sense.

subtests/docker_cli/systemd_in_container/test.sh
adelton: Nod, that's what I meant to do there and I messed up my rebasing.

Hopefully fixed now (and rebase on master) -> dc49efac40fde25290c06fa4146912fa0e6fbdab.

subtests/docker_cli/systemd_in_container/test.sh
adelton: Removed in 366b8dd7bd9df97ae51abca1ab0744aa2b34f484.

subtests/docker_cli/systemd_in_container/test.sh
adelton: Removed in 366b8dd7bd9df97ae51abca1ab0744aa2b34f484.

adelton:
Hopefully addressed all the concerns, rebased on master -> 366b8dd7bd9df97ae51abca1ab0744aa2b34f484.

adelton:
To minimize the delay when `docker run` already fails (for example the image does not exist), I've check for `docker run`'s exit status. -> 3bbf6aa73ba1b57592aebf3818601474183ca8bb.

cevich:
> No good suggestions, sorry:

Not true!  You've made some excellent suggestions (below):

> is it possible to get pylint from a CentOS repo instead of EPEL? Is it possible to yum-install an explicit version? (That assumes that this problem is a new one caused by a recent update to EPEL pylint). Also on the assumption that a recent pylint update broke EPEL, is it likely to be caught and fixed soon?

Yes, it's a recent update to EPEL that caused this and the problem only exists in EPEL.  There's a "fixed" bug, that doesn't actually fix the problem.  But you gave me an idea...I can't use yum to install a specific version directly, but I could point it at a URL of an older version, perhaps that would work...(working on it).

> CI is not the place to test or debug code.

For small things sure, I agree.  For this, with multiple distros. and multiple environments (docker/docker-latest, etc) relying on CI is for sure more efficient use of developer-time.  For example, you can work on one of many other ongoing tasks while you wait for CI results.  No reason to spend developer time on things that can be automated by (cheap) machine time.

Though we could use some magic to allow passing ``--args`` in to the CI runs (to enable/disable specific tests for example).

cevich:
jenkins, test this please.

cevich:
@adelton Ignoring the VMs that bombed out completely, your test did run successfully on RHEL Atomic.  Here are the debug files  https://hurl.corp.redhat.com/e8edb2f  from the 
develop_dockerautotest-ghprb run

adelton:
That would be a good news right? ;-)

cevich:
Possibly...you didn't expect me to actually LOOK at the debug files, ya? 😀 

Really it just means I've finished jamming in a temporary PR-testing job that runs all tests on all platforms for PRs.  Every-time you update the code in this PR, it'll re-run the tests automatically.  Oh...wait...did you want a new puppy-dog too?  Sorry, I forgot.

---

Over here: https://github.com/cevich/autotest-docker-enabled-product-testing/pull/169 (if it works and I merge it), will allow safely passing ``--args`` to the actual test VMs ``autotest-local`` command.  In a nutshell, that means you can limit the run to specific tests.   That should shave 20-30 minutes off of the CI job's runtime.  For example, in this PR, by adding into the description text (anywhere):

```
[--args=docker_cli/systemd_in_container]
```

(Once I merge that ADEPT PR)

subtests/docker_cli/systemd_in_container/test.sh
```diff
@@ -0,0 +1,73 @@
...
+    local lrc=0
+    for expect in "$@"; do
+        echo "Checking for $expect"
+        i=0
+        while [ $i -lt $timeout ] ; do
+            if docker logs "$container" | grep -q "$expect" ; then
+                echo "PASS $name = $expect"
+                break
+            else
+                if [ $i -lt $timeout ] ; then
```
adelton: Removed, rebased on master -> 3b72a8e054baa84d25fb57a2fd046bb73c92ab67.

cevich:
Excellent, thanks Ed.  If it LGTU it LGTM also 😄 

@adelton In the ADEPT CI job (for this PR), the rhel7-4...lvm failure is expected, as are the other test failures except for:  [fedoratomic26-docker-latest.subtests.docker_cli.systemd_in_container](https://hurl.corp.redhat.com/b33d75e).  Is that something you were expecting?

Also, many of the CI jobs run all tests by default, on all platforms (RHEL, Fedora, Atomic, docker, and docker-latest).  Are there any limits that should be imposed WRT this new test, and/or any special environments it should be run in (custom default test images, for example)?

adelton:
> In the ADEPT CI job (for this PR), the rhel7-4...lvm failure is expected, as are the other test failures except for: fedoratomic26-docker-latest.subtests.docker_cli.systemd_in_container. Is that something you were expecting?

I wonder if it can be SELinux related. Could we try

```
setsebool -P container_manage_cgroup 1
```

on that machine?

> Also, many of the CI jobs run all tests by default, on all platforms (RHEL, Fedora, Atomic, docker, and docker-latest). Are there any limits that should be imposed WRT this new test, and/or any special environments it should be run in (custom default test images, for example)?

In ideal case, all base images that have systemd installed should be able to run just fine.

cevich:
> I wonder if it can be SELinux related. Could we try

Shouldn't be a reason to wonder, we [capture ``audit.log`` and ``messages`` for](https://hurl.corp.redhat.com/be39371) every test, looking...

> ``setsebool -P container_manage_cgroup 1``

Should that be set for all of F26+ with docker-latest?

Let me see if I can reproduce the failure in a manual CI run...

cevich:
@adelton Ya, I think you're right, in ``audit.log`` I see:

```
type=AVC msg=audit(1502872056.507:8809): avc:  denied  { mounton } for  pid=1800 comm="systemd" path="/proc" dev="proc" ino=1 scontext=system_u:system_r:container_t:s0:c331,c427 tcontext=system_u:object_r:proc_t:s0 tclass=dir permissive=0
...
type=AVC msg=audit(1502872059.519:8816): avc:  denied  { write } for  pid=1800 comm="systemd" name="release_agent" dev="cgroup" ino=7 scontext=system_u:system_r:container_t:s0:c331,c427 tcontext=system_u:object_r:cgroup_t:s0 tclass=file permissive=0
type=SYSCALL msg=audit(1502872059.519:8816): arch=c000003e syscall=2 success=no exit=-13 a0=5630772dbfd0 a1=80101 a2=0 a3=7c items=0 ppid=1759 pid=1800 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="systemd" exe="/usr/lib/systemd/systemd" subj=system_u:system_r:container_t:s0:c331,c427 key=(null)
type=PROCTITLE msg=audit(1502872059.519:8816): proctitle="/usr/sbin/init"
type=AVC msg=audit(1502872059.519:8817): avc:  denied  { write } for  pid=1800 comm="systemd" name="docker-72d497fa292a703bb860872a42f9c64354602e09656470519381cfb1c311110b.scope" dev="cgroup" ino=2335 scontext=system_u:system_r:container_t:s0:c331,c427 tcontext=system_u:object_r:cgroup_t:s0 tclass=dir permissive=0
```

Trying a manual CI run with ``setsebool -P container_manage_cgroup 1``...

cevich:
... okay @adelton I was able to reproduce the exact same failure.  Trying your suggested fix...

cevich:
...fix confirmed, 100% pass now.  Okay, I'll merge this PR and open a new one for ADEPT to make sure this selinux bool gets flipped for this specific VM.

Thanks again for all your work on this, please feel free to mail or ping us if you need anything else 😀

PR #727, opened 2017-06-02
Docs: Update quickstart and test-selection section

cevich:
The quickstart was very old, and not representitive of modern output.
Also cleanup docs pertaining to selecting/including/excluding tests.
Make the ``--args`` command-line option usage, and the ``control.ini``
usage more clear.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -87,19 +87,18 @@ Prerequisites
 
 *  Platform Applications/tools
 
-    *  Autotest 0.16.0 or later (preferred).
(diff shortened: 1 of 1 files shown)


cevich:
N/B: Powers-that-be specificity requested a quick-start that checks out the latest branch, and runs a simple, always-passes type of test.  While I was at it, I also cleaned up the ``--args`` and ``control.ini`` sections.

To build HTML docs, type '`make`` (assuming you have sphinx installed).  For more info...see the docs :D

PR #728, opened 2017-06-05
kill deferred deletion tests

edsantiago:
Turns out to be complicated, and the feature will not be
enabled by default. No point in testing.


cevich:
Oh, thanks @edsantiago I appreciate your taking care of this.

PR #730, opened 2017-06-08
Refix dd

cevich:


edsantiago:
Please don't do this. I need to reexamine how this should be done.

cevich:
This PR doesn't break or fix anything in the DD-test that wasn't already working, or broken.  It only removes a test that was for sure broken and brings back one that shouldn't have been removed.

cevich:
jenkins, test this please.

PR #731, opened 2017-06-08
Simplify log_versions

cevich:
Previously versions of the various docker RPMS was collected in two
places. This is redundant possibly ambiguous, and unnecessary.  The
"active" version of docker is also meaningless since subtests are free
to restart at will.  It would only represent the active version when the
pretest runs.  Better to leave making any declaration of "active"
version up to any subtests where it matters (none currently).

### DIFF
--- a/config_defaults/log_versions.ini
+++ b/config_defaults/log_versions.ini
@@ -1,3 +1,3 @@
 [log_versions]
 #: CSV of rpm names to store in sysinfo/key_rpms
-key_rpms = docker-selinux,container-selinux,runc,atomic,skopeo
+key_rpms = docker,docker-latest,docker-selinux,container-selinux,runc,atomic,skopeo
--- a/pretests/log_versions/log_versions.py
+++ b/pretests/log_versions/log_versions.py
@@ -18,7 +18,6 @@ from autotest.client import utils
 from dockertest import subtest
 from dockertest.dockercmd import DockerCmd
 from dockertest.output import DockerVersion, mustpass
-from dockertest.docker_daemon import which_docker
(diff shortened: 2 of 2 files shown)


edsantiago:
> The "active" version of docker is also meaningless since subtests are free to restart at will

This isn't quite true. Subtests can restart, possibly with different docker options, but there is currently no mechanism for a test to switch between docker and docker-latest. This would be such an incredibly bad idea that I don't foresee us ever adding such functionality.

cevich:
Which docker is currently running is the one it says when it prints out the version.

edsantiago:
Can you rephrase that please? I don't understand it.

I will submit a PR to reinstate docker_rpm_active because my jenkins-analyze script relies on it.

cevich:
[(2) [CI] Collect set of critical rpm NVRA's per run.](https://trello.com/c/JS4KG9H5/499-2-ci-collect-set-of-critical-rpm-nvras-per-run)

edsantiago:
For benefit of anyone finding this thread in the future: what @cevich refers to in his above comment is that one could read `sysinfo/docker_version` and then match that to either `docker` or `docker-latest` in the file `sysinfo/key_rpms`. This doesn't actually work, because docker and docker-latest may be the same major version (e.g. different 1.12.x builds).

In the absence of `sysinfo/docker_rpm_active`, one way for a reader to determine the active docker is to fetch `sysinfo/sysconfig.tar.gz`, extract it into a tempdir, read the extracted `sysconfig/docker` file, and check the `DOCKERBINARY=` line near the bottom.

PR #733, opened 2017-06-12
Reenable deferred_deletion test

edsantiago:
Per vgoyal: the functionality is still important, let's not
allow it to regress. This reintroduces the test, with a
few minor changes from before:

  * Remove the check for devicemapper: docker-storage-setup
    only enables deferred_deletion on devicemapper. Even
    the broken RHEL 7.2 version.
  * Don't just check for the existence of /proc/sys/fs/may_detach_mounts;
    read it and confirm that it's enabled.
  * Only do the /proc/sys/fs/may_detach_mounts check on RHEL.
    This lets the test run on Fedora, which never had the
    kernel bug requiring that sysctl switch.

### DIFF
--- a/subtests/docker_cli/deferred_deletion/deferred_deletion.py
+++ b/subtests/docker_cli/deferred_deletion/deferred_deletion.py
@@ -1,0 +1,145 @@
+r"""
+Summary
+---------
+
+Test deferred deletion: device mapper will not delete a docker
+filesystem if it's still in use.
+
+This test is only applicable if docker daemon is run with:
+
+    --storage-opts dm.use_deferred_deletion=true
+
+(this is detected and enabled by the docker-storage-setup script)
+
+That suffices on Fedora. On RHEL it's more complicated: the 7.2 version
+of docker-storage-setup blindly enabled the option even though it
+didn't actually work due to a kernel bug. 7.4 introduced a kernel with
+a fix, but disabled by default. We therefore need to differentiate
+between systems with (1) enabled incorrectly and correctly. We do
+so by checking that the sysctl knob /proc/sys/fs/may_detach_mounts
+(introduced in kernel 3.10.0-632) exists and is 1.
+
+We abort with TestNAError if any precondition fails.
+
+Note that we need to do all the important work from inside a
+helper script, because our autotest process will only have access
+to the container's mounts if MountFlags=slave is *absent* from
+the docker systemd unit file; by default that option is present,
+and it is not clear if/when it will be removed. To make sure
+we have access to the container's mounts we need to nsenter
+the docker daemon's mount namespace, and we can't do that within
+autotest (using ctypes.CDLL and setns) because autotest is
+multithreaded and setns() doesn't allow that.
+
+Operational Summary
+----------------------
+
+#. Run a docker container
+#. Run a test script inside docker mount space that will:
+#. - Find out the local (host) mountpoint of the container's root filesystem.
+#. - cd into that directory
+#. - Read the Deferred-Deletion count from 'docker info'.
+#. - Remove the container.
+#. - Read the Deferred-Deletion count; confirm that it has grown by one.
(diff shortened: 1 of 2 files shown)


edsantiago:
Most useful way to diff is probably:

    $ git pr 733
    $ git diff 41ebca6 subtests/docker_cli/deferred_deletion/deferred_deletion.py

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,145 @@
...
+
+Note that we need to do all the important work from inside a
+helper script, because our autotest process will only have access
+to the container's mounts if MountFlags=slave is *absent* from
+the docker systemd unit file; by default that option is present,
+and it is not clear if/when it will be removed. To make sure
+we have access to the container's mounts we need to nsenter
+the docker daemon's mount namespace, and we can't do that within
+autotest (using ctypes.CDLL and setns) because autotest is
+multithreaded and setns() doesn't allow that.
```
cevich: A bit much for a summary, eh?  See if you can move the bulk of this under an ``Operational Detail`` heading.

Background:  A long time ago, in a galaxy far, far, far, far away.  We were asked to bundle each test's summary in with it's results.  A lot of work was done to the documentation processing modules to enable doing this.  I'm willing to bet that at some point, having test results w/ summaries for reporting purposes will come again.  However, this will only remain possible if the test summary sections are kept relatively short and summary-ish 😀

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,145 @@
...
+    def initialize(self):
+        super(deferred_deletion, self).initialize()
+
+        docker_cmdline = dockertest.docker_daemon.cmdline()
+        self.failif_not_in("dm.use_deferred_deletion=true",
+                           str(docker_cmdline),
+                           "docker daemon command-line options",
+                           DockerTestNAError)
+
+        self._rhel72_bug_workaround()
```
cevich: Good idea to have this as a separate/private function.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
```diff
@@ -0,0 +1,145 @@
...
+        # RHEL
+        sysctl_knob = "/proc/sys/fs/may_detach_mounts"
+        self.failif(not os.path.exists(sysctl_knob),
+                    "sysctl knob %s does not exist; this system is"
+                    " not likely to support deferred deletion" % sysctl_knob,
+                    DockerTestNAError)
+
+        with open(sysctl_knob, 'rb') as fs_may_detach_mounts:
+            enabled = int(fs_may_detach_mounts.read().strip())
+            self.failif(not enabled,
```
cevich: optional: Can ``failif_ne()`` be used for these checks at all?  Sometimes these knobs end up having more values added on later.  Not very likely, mostly thinking that the output from that function may save some error-message writing (here).

cevich:
Hrmm, dunnalike how review comments kinda get hidden...re-posting:

> Mostly approve 😀 Just the docstring thing needs fixup, code-wise, I'll stand by the PR's CI results.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
edsantiago: Done. Thanks for the background! It makes good sense to keep Summary brief; I'll keep that in mind for future tests.

subtests/docker_cli/deferred_deletion/deferred_deletion.py
edsantiago: `failif_ne()` was my first choice but I just wasn't happy with the error message. Although writing maintainable code is my second-highest priority, the end consumer (person running the tests) is my principal priority. This is the clearest, most time-saving message I could come up with.

PR #734, opened 2017-06-13
New test: live-restore

edsantiago:
Start a container, restart the docker daemon, make sure our
container is still listed in 'docker ps' and still running.

Introduces new systemd_show() function in docker_daemon,
used to obtain the value of 'KillMode'. Refactored
docker_daemon.pid() to use this new function.


edsantiago:
I have not tested this with containerized docker. That's next, but it won't be until late this afternoon/evening.

cevich:
Is there a card for this?  If so, smash the 'github' button (in the card) and link it to this PR please.  Also, thanks for writing this test, taking a quick-peek now...

dockertest/docker_daemon.py
```diff
@@ -153,14 +153,22 @@ def restart():
...
+    Returns the value as a string.
+    """
+    cmd_result = _systemd_action('show --property=%s' % prop)
+    stdout = cmd_result.stdout
+    if not stdout.startswith(prop + '='):
+        raise RuntimeError("Unexpected output from %s: expected %s=XXXX,"
+                           " got '%s'" % (cmd_result.command, prop, stdout))
+    return stdout[(len(prop) + 1):].strip()
+
+
```
cevich: Consider adding a basic unittest for callers of ``_systemd_action()``.  The function is just a wrapper for ``utils.run``, so it's easy to mock.  Then all the callers can be checked easily with input, output, expected mapping..

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,185 @@
...
+---------
+
+Test live-restore: running container should survive docker daemon restart.
+
+Operational Summary
+----------------------
+
+#. Run a docker container
+#. Restart docker daemon
+#. Verify that container is listed under 'docker ps' and is still running.
```
cevich: (Haven't read below) How does the test prove the container was running while the deamon was not.  In other words, that restarting the daemon wasn't also responsible for starting the container (i.e. maybe the default --restart argument changes)

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,185 @@
...
+possibility of leaving our system in an undefined state if something
+goes wrong between stop and start.
+
+Prerequisites
+-------------
+
+This test is only applicable if docker daemon is run with --live-restore
+option; we abort with TestNAError if it isn't.
+
+live-restore also requires `KillMode=process` in the docker systemd
```
cevich: Doesn't it also require that ``MountFlags=slave`` is absent?

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,185 @@
...
+                    "Last-output time from container (%s) is >2 seconds"
+                    " off from system time" % time.ctime(container_time))
+
+    def _container_log(self):
+        """
+        Returns container log as a list of strings
+        """
+        c_name = self.stuff['container_name']
+        result = mustpass(DockerCmd(self, "logs", [c_name]).execute())
+        return result.stdout.strip().split("\n")
```
cevich: The kernel (and even the VM's host) [clocks are non-monotonic](https://en.wikipedia.org/wiki/Monotonic_function#See_also).  There's no guarantee that can't flow backward or repeat jiffies in totally unpredictable ways.  Consider if ``NTP`` or ``chrony`` happens to tweak backwards a few milliseconds, or changes time-zones immediately before a second's boundry.

The simple fix here is to make the time checks both ordered and interval-based (instead of absolute).  So, have it dump the value of a counter that increments on 100ms intervals (``sleep 0.1s`` for example).  Otherwise this test could fail in an unpredictable and extremely difficult to diagnose way.

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,185 @@
...
+                           "docker daemon command-line options",
+                           DockerTestNAError)
+
+        # rhbz1424709#c40 : live-restore fails unless KillMode=process
+        killmode = docker_daemon.systemd_show('KillMode')
+        if killmode != 'process2':
+            self.logwarning("systemd KillMode is '%s' (expected 'process')."
+                            " This test will probably fail." % killmode)
+
+        self.stuff['dc'] = DockerContainers(self)
```
cevich: This needs to be coordinated with https://github.com/cevich/autotest-docker-enabled-product-testing/pull/142 or similar/related otherwise it ain't gettin' tested in CI.

edsantiago:
[[QE] Test cases for --live-restore feature](https://trello.com/c/aKwN2Axb/474-qe-test-cases-for-live-restore-feature)

dockertest/docker_daemon.py
edsantiago: I assume you mean systemd_*show*, not _action. Good thinking. It'll take me a while because python testing always takes me a while; I'll have something ready tomorrow.

subtests/docker_cli/liverestore/liverestore.py
edsantiago: It checks that the container is now a child of init, not of dockerd. I suppose I could preserve the container PID and make sure it doesn't change. I'll add that for next revision.

subtests/docker_cli/liverestore/liverestore.py
edsantiago: It doesn't look like it. I've done all my testing with `MountFlags=slave` and all tests (including my manual ones) pass. I think it does require `fs.may_detach_mounts=1` but I haven't actually checked.

subtests/docker_cli/liverestore/liverestore.py
edsantiago: No coordination needed, since the test will recuse itself with `NA` if docker daemon isn't run with --live-restore

subtests/docker_cli/liverestore/liverestore.py
edsantiago: Time zone: irrelevant, we're talking epoch. NTP: the `abs()<2` check accounts for a reasonable margin of error. If NTP does a more-than-1-second adjustment, the system is broken, and other tests will probably break. I'm not sure it's worth spending time on a problem this unlikely.

dockertest/docker_daemon.py
cevich: Naa, I do mean ``_systemd_action()`` (the private function).  If you mock that, then testing all the other functions which call it should be easy.  My thought was to have a test method run through a list of mappings for function-name, input-arguments, expected return, and expected wrapper-arguments (to the mocked ``_systemd_action()``.  Just a cheap idea though, do whatever is easier

> python testing always takes me a while; I'll have something ready tomorrow.

Mannn, don't you know it!  Testing is gotta be like 10-times harder than writing the subject-code itself, no doubt!  Take your time, there's no rush.  I don't care if unittest code is pretty, so long as it gets the job done. Fugly and duct-tape covered is perfectly acceptable 😀

subtests/docker_cli/liverestore/liverestore.py
cevich: Naa, testing the process hierarchy is more than good'nuf, forget my comment.

subtests/docker_cli/liverestore/liverestore.py
cevich: Huh, interesting.  It's painful how hard 'tis to get reliable details like this sometimes.

subtests/docker_cli/liverestore/liverestore.py
cevich: It's not the chance of it occurring I'm worried about, it's the resulting hair-loss trying to figure out what happened just by looking at the logs.  VMs can be paused, moved across hosts, NTP, leap-seconds, etc.  there are any number of ways the absolute-time can change, it's not a reliable measure.  

The other reason I suggested it is: I think it would make the test code simpler.  It only needs to verify the series instead of trying to make sense of timestamps across domains (host vs container).  It also enables sub-second intervals, helping jobs complete faster.   e.g.

``docker run ... "for (( N=0; 1; N++ )); do sleep 0.1s; echo $N; done"``

and

```python
out_seq_start = result.find('READY\n') + len('READY\n') + 1
out_seq = [int(i.strip()) for i in result[out_seq_start].strip().splitlines()]
self.failif_ne(out_seq, range(len(out_seq)), "non-linear output sequence")
```

cevich: Even better would be to have it write to a file in a volume-mount, and confirm the sequence is running while the docker daemon is stopped.  Anyway, I suppose what you have is fine, just a suggestion that might make it simpler.

dockertest/docker_daemon_unittests.py
```diff
@@ -109,5 +136,48 @@ def test_bad_input_mismatched_quotes(self):
...
+        # It also checks the docker daemon command line, because it has to
+        # distinguish between dockerd itself and dockerd run under runc
+        # (as a container). The second fakerun_setup() simulates the
+        # output of 'ps' on our pid.
+        fakerun_setup(stdout="MainPID=12345\n")
+        fakerun_setup(stdout="/usr/bin/dockerd --add-runtime ...\n")
+
+        self.assertEqual(docker_daemon.pid(), 12345, 'daemon pid')
+
+
```
cevich: Righton, good'nuf.

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,204 @@
...
+        log_0 = self._container_log()
+
+        # Wait up to 5 seconds for output to change
+        end_at = time.time() + 5
+        while time.time() <= end_at:
+            log_1 = self._container_log()
+            if log_1 != log_0:
+                break
+            time.sleep(1)
+
```
cevich: Compromise:   Dump ``log_0`` and ``log_1`` contents into ``self.resultsdir`` for later examination (non-monotonic time problems)?

edsantiago:
@cevich welcome back! I've addressed your concerns, PTAL and merge if you think it's OK

cevich:
@edsantiago caught your comment, won't forget about this (unless I forget about this).

dockertest/docker_daemon.py
```diff
@@ -126,11 +126,17 @@ def which_docker():
...
     """
     docker = 'docker'
-    with open('/etc/sysconfig/docker', 'r') as docker_sysconfig:
-        for line in docker_sysconfig:
-            if line.startswith('DOCKERBINARY='):
-                if 'docker-latest' in line:
-                    docker = 'docker-latest'
+
+    # /etc/sysconfig/docker doesn't exist on the CI host on which
+    # we run github unit tests
```
cevich: Hrmmm, or it's not called docker, or it's not installed, or ... lots of ways this could break 😢 

This feels wrong to me: Changing behavior in code when it _guesses_ it's running under test.  What happens when it guesses wrong...it silently fails.  VW had a similar problem with their TDIs and the EPA, maybe a bad analogy, but same idea.

dockertest/docker_daemon.py
```diff
@@ -126,11 +126,17 @@ def which_docker():
...
+    try:
+        with open('/etc/sysconfig/docker', 'r') as docker_sysconfig:
```
cevich: One idea could be to mock the open call (which I'm told is hard).  Another idea is to indirect reference the open-call, to support unit-tests modifying it:

```python
OPEN_FN = open
...

    def which_docker():
        ...
        with OPEN_FN(...):
            ...
```

dockertest/docker_daemon.py
```diff
@@ -126,11 +126,17 @@ def which_docker():
...
+            for line in docker_sysconfig:
+                if line.startswith('DOCKERBINARY='):
+                    if 'docker-latest' in line:
+                        docker = 'docker-latest'
+    except IOError:
+        pass
```
cevich: Another idea is kinda duct-tape-ish but addresses the "silence" problem:

```python
        ...
        except IOError:
            logging.logwarning("I think I'm running under test, so I'm ignoring the failure to open /etc/sysconfig/docker")
        ...
```

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,200 @@
...
+        docker_daemon.restart()
+        t1 = time.time()
+        if t1 - t0 > 30:
+            self.logwarning("docker restart took %d seconds", t1 - t0)
+
+    def _start_container(self):
+        """
+        Start a container. We only need it briefly, until we (the test,
+        running in host space) can cd into its root filesystem. Container
+        will spin idly until a semaphore file is removed.
```
cevich: Nowhere in this function is anything acting on a lock ifle, should this part of the docstring be moved elsewhere?

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,200 @@
...
+        # Wait up to 5 seconds for output to change
+        end_at = time.time() + 5
+        while time.time() <= end_at:
+            log_1 = self._container_log()
+            if log_1 != log_0:
+                break
+            time.sleep(1)
+
+        self.failif(len(log_0) == len(log_1),
+                    "Container log line count did not grow")
```
cevich: I think in all cases, it would be helpful to have the output for later reference.  Could we do something like this:

```python
        ....
        self.logdebug(self._container_log())
        self.failif(len(log_0) == len(log_1),
        ...
```

subtests/docker_cli/liverestore/liverestore.py
```diff
@@ -0,0 +1,200 @@
...
+        # Run 'docker ps', confirm that container is listed and is active
+        dc = self.stuff['dc']
+        docker_ps = dc.list_containers_with_name(self.stuff['container_name'])
+        self.failif(not docker_ps,
+                    "Container not in 'docker ps' output after restart")
+        self.failif_not_in("Up ", docker_ps[0].status,
+                           "container status as reported by 'docker ps'")
+
+        # Confirm that container is still running (spitting out dates)
+        self._verify_that_date_changes()
```
cevich: Hmmm, just occurred to me a simpler and faster state-full liveliness test could be:

#. Record the contents of ``/proc/<pid>/stat``.
#. Stop / start daemon
#. Compare before/after ``stat`` fields 22 and 14:   ``starttime_before == starttime_after && utime_before < utime_after``

The you only need the busy-loop running inside the container, and don't need logging or timeout-loop or date checks at all.  It's basically impossible for the process PID to remain the same, and also have the starttime match and utime advance (CPU ticks).

Up to you, but I think it would be simpler.

edsantiago: stat fields 22 and 14:

    # ln /bin/cat /bin/"kitty cat"
    # /bin/kitty\ cat /proc/self/stat
    6258 (kitty cat) R 29302 6258 29302 34820 6258 4202496 218 0 0 0 [...]

(Not that RH is likely to ship an executable with spaces in the name).

The problem, though, is that this isn't the purpose of the test. I don't actually care that the container is running: I care that the container is running and that docker can talk to it (via docker logs). 

Thanks for your feedback. I can't find any ways to improve this test right now in the ways you request while also keeping it clean. I will revisit it next week and see if I have new insights.

cevich: Probably why the name is in parenthesis, though you could count the fields from the end as well to avoid extra parsing.  An implied use-case of live-restore is also that the container continues to do work while disconnected from the docker daemon.  IMHO, checking the CPU counters for the process is a simpler proof than date-based check.  OTOH, it's such a fine-line, it's not worth making a fuss over, what you have is fine too.

cevich:
Re: "try and try and try all day still nogo" :(

Assuming you're referring to the time vs stat check.  Okay, sounds like you gave it an honest effort.  Sorry I didn't anticipate it would be such a PITA.   I can bend, and learn to live with the old/working check.  If you like, let's call it imperfect but good enough for merge.  Unless you want to keep working on it / are almost done.

edsantiago:
>Assuming you're referring to the time vs stat check. 

I still don't understand what you mean by "time" (as in "time check"). I got rid of the time code almost a month ago (June 14). Now the only check is that the number of output lines increases, which I thought was safe enough and good enough.

dockertest/docker_daemon.py
edsantiago: Fixed in upcoming rebase.

dockertest/docker_daemon.py
edsantiago: I still can't understand the point of testing `_systemd_action()`. It has no useful inputs or outputs. It does one thing only, which is wrap `utils.run`. Any test would just be a test of `fakerun()`. What am I missing?

subtests/docker_cli/liverestore/liverestore.py
edsantiago: My bad: I copy-pasted from another subtest and did not update the dock string. Fixed in upcoming push.

subtests/docker_cli/liverestore/liverestore.py
edsantiago: I'm not sure what this accomplishes. There is no longer any issue with monotonic time, because the `date` output is ignored. I just care about increasing number of output lines.

dockertest/docker_daemon.py
cevich: Nono, I said _mock_ ``_systemd_action()``, not test it.  Doesn't matter tho, in the other PR you mocked ``utils.run`` which has same effect.  No worries then.

subtests/docker_cli/liverestore/liverestore.py
edsantiago: I still really just don't understand the point of this or why it's better than checking stdout. I will come back to it tomorrow and try yet again.

cevich:
> "time" (as in "time check")

This method (line 118): 
``_verify_that_date_changes()``

https://github.com/autotest/autotest-docker/pull/734/files#diff-07050262d07960cac8ba95a389b8419aR118

> which I thought was safe enough and good enough.

It is, I think my confusion is on the method's name and docstring.  Feel free to fix those if you must, but I'm not going to hold this up if you don't.

cevich:
> I will come back to it tomorrow and try yet again.

Great, I've got my finger on the merge button :D

edsantiago:
I had to force-push because I rebased (to pick up the which_docker changes). Tested on RHEL 7.4 with docker and docker-latest, all tests passed; tested on rHEL 7.3 (failed with TEST_NA as expected).

cevich:
Excellent, thanks for all your not-easy work on this, merging...

PR #736, opened 2017-06-28
Workaround docker-latest build re: BZ 1403326

cevich:
Docker ``1.12+`` regresses WRT this bug in that it adds additional
escape-codes to docker build output.  This is being (correctly)
caught by the ``OutputGood()`` check, however the bug seems
unlikely to ever be fixed.  Unfortunately, removing the check
isn't enough, because many sub-subtests also rely on regular-expression
matching against output.

For example, searching for ``^'Schazam!$'`` where the build output is:

```
Step 4/6 : RUN echo "Schazam!"
 ---> Running in cc59317feb51

[91m[0mSchazam!
 ---> 8a359f725df3
Removing intermediate container cc59317feb51
```

Would fail to match due to escape codes at the start of the important
line.  This commit adds a configurable option, defaulting to "off" which
enables checking for escape codes in ``OutputGood()``.  It also removes
the start-of-line match token from all the regex's, but leaves the
end-of-line token.  The assumption here is that escape codes will never
appear at the end of the line being matched (possibly not true).

### DIFF
--- a/subtests/docker_cli/build/build.py
+++ b/subtests/docker_cli/build/build.py
@@ -201,10 +201,15 @@ class postprocessing(object):
 
     def basic_postprocess(self, build_def, command, parameter):
         del parameter  # not used
+        # June 2017: docker-1.13 ignores $TERM for build sub-command.
+        #            Jambs ANSI codes in output, BZ1403326
+        skip = ['nonprintables_check']
+        if self.config['enable_nonprintables_check']:
+            skip = None
         if command == 'positive':
             # Verify zero exit status and healthy output
             opg = OutputGood(build_def.dockercmd.cmdresult,
-                             ignore_error=True)
+                             skip=skip, ignore_error=True)
             if opg:
                 return build_def.dockercmd.cmdresult.exit_status == 0
             else:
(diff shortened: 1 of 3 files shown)


subtests/docker_cli/build/build.py
```diff
@@ -201,10 +201,14 @@ def parse_postprocess_commands(self, build_def):
 
     def basic_postprocess(self, build_def, command, parameter):
         del parameter  # not used
+        if self.config['enable_nonprintables_check']:
+            skip=['nonprintables_check']
```
edsantiago: reversed test sense

edsantiago: Okay, with the conditional fixed, tests pass cleanly and I think this is a reasonable workaround for the ugly situation.

Because the normal case is to disable the nonprintables check, I think the code would be much more readable and maintainable as:

    # June 2017: docker-1.13 includes ANSI codes in output, with no way to disable
    skip = ['nonprintables_check']
    if self.config['enable_nonprintables_check']:
        skip = None

This weights the default case more heavily than the exception.

edsantiago:
Weird. I got the wait_for_ready email but never saw one about fixing the test sense. Anyhow, LGTM, tests now running on my systems. Thanks.

PR #737, opened 2017-07-06
wait_for_ready(): better handling of finished containers

edsantiago:
In main timer loop, check container-running status (via self.done).
Terminate early if container is not running and we don't see the
expected READY string. This prevents a 10-minute wait on a
container that has exited with an error.

As long as we're in here, move the sleep() further down, to
the end of the loop. Not because it saves us 0.2 seconds but
because it makes the code flow more readable.


edsantiago:
In real-world terms: the `run_volumes_selinux` test always fails with user namespaces turned on (known problems). This fix saves 20 minutes of waiting for never-even-started containers.

dockertest/dockercmd.py
```diff
@@ -338,16 +338,18 @@ def execute(self, stdin=None):
...
         :raises DockerExecError: on timeout.
         """
         if timeout is None:
             timeout = self.timeout
         end_time = time.time() + timeout
-        while time.time() <= end_time:
-            time.sleep(timestep)
+        done = False
+        while time.time() <= end_time and not done:
+            done = self.done
```
cevich: Why not just ``while time.time() <= end_time and not bool(self.done)``?

edsantiago: Race condition: the container could have emitted READY and exited during the `sleep()`.

PR #738, opened 2017-07-10
which_docker(): reimplement using systemctl

edsantiago:
We now scan the output of 'systemctl list-units', looking
for one of our known docker engine services (docker,
docker-latest, container-engine) in 'running' state.

This is a much cleaner, safer implementation: it gives us
the actual name of the running systemd service, not a
close-enough guess.

Also: add unit tests


edsantiago:
This addresses one of your review comments in the liverestore test. My intention is to rebase the latter, using this new functionality for passing selftests.

dockertest/docker_daemon.py
```diff
@@ -117,20 +118,25 @@ def version(self):
...
-                if 'docker-latest' in line:
-                    docker = 'docker-latest'
+
+    # Known docker-daemon services as of July 2017
+    docker_services = ('docker', 'docker-latest', 'container-engine')
+
+    # List systemd units, look for "xx.service loaded active running"
+    re_svc = re.compile(r'^([\w-]+)\.service\s+loaded\s+active\s+running\s')
+    list_units = 'systemctl list-units --full --type=service --state=running'
+    units = utils.run(list_units).stdout.strip().split("\n")
```
cevich: ``units = utils.run(list_units).stdout.strip().splitlines()``

cevich:
One fine-letter fix than this is good to merge.  Nice work, Thanks!

PR #739, opened 2017-07-11
Remove packaging subtest

edsantiago:
per rhbz#1432520 the golang issues with stripping debuginfo
have been resolved. docker now ships stripped, with a separate
debuginfo package, like normal packages.

I considered flipping the sense of this test, but it's best
to just delete it outright. Any changes in stripped/not-stripped
package state will be caught by rpmdiff.


cevich:
Yep, agreed.

PR #740, opened 2017-07-14
new oci-umount test

edsantiago:
oci-umount is a new (July 2017) docker plugin that unmounts
certain filesystems before a container is started. This test
looks for the presence of /etc/oci-umount.conf and, if it
exists, runs a container with a number of bind mounts and
ensures that no path mentioned in the .conf file is mounted
in the container.

Tested on RHEL 7.4, docker-1.12.6-45.git1680dd8.el7, works
fine.

Tested on RHEL 7.3, docker-1.12.6-32.git88a4867.el7, skips
with Test_NA. Tested by copying /etc/oci-umount.conf from
a RHEL 7.4 box, and test fails with long list of filesystems
that should not be mounted (i.e. it works as expected).

Tested on RHEL 7.4 with containerized docker 1.12.6-8. Test
does not actually pass, but it may be a problem in the
containerized docker build (older oci-umount). Am submitting
for review despite that as-yet unresolved problem.

### DIFF
--- a/config_defaults/subtests/docker_cli/run_volumes.ini
+++ b/config_defaults/subtests/docker_cli/run_volumes.ini
@@ -1,5 +1,5 @@
 [docker_cli/run_volumes]
-subsubtests = volumes_rw, volumes_one_source
+subsubtests = volumes_rw, volumes_one_source, oci_umount
 #: These must be setup prior to testing
 __example__ = host_paths, cntr_paths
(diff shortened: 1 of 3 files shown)


cevich:
LGTM.

edsantiago:
[(3) Test oci-umount as packaged with docker](https://trello.com/c/3V8FPYpQ/581-3-test-oci-umount-as-packaged-with-docker)

PR #741, opened 2017-07-20
new test: oci_umount_bz1472121

edsantiago:
Intended as a short-lived test, to be removed maybe Sept 2017.
Purpose is to confirm that this regression is fixed on all
distributed builds of docker.

This tests for a corner case in oci-umount that triggers
a SEGV. Test fails (correctly) in docker-1.12.6-47.git0fdc778.el7,
passes in -48.


cevich:
Re: Just opened https://github.com/autotest/autotest-docker/pull/742

Lol, well it's a good thing we didn't do exactly the same thing 😄 

Okay, taking a look at this, though maybe it can make use of some of the new machinery in #742?

cevich:
Okay, LGTM.  Let me know what's best WRT #742 if merge-order makes any difference.

PR #742, opened 2017-07-20
Collect oci_umount package version

cevich:


cevich:
>     the point of the caches was to make these globals across an entire run. So when one subtest calls DockerVersion(), then another subtest does, we don't have to waste effort rerunning anything.
>     docker version is useless. You cannot reliably get the RPM version from it. (try it with docker-latest)

Naa, the cache's only help with repeated calls from the same instance.  It could be made class-level cache, but still would only help in the same process-space.  Remember Autotest forks for each subtest.  Perhaps calling it a cache is a mistake, it's really more like a buffer.

> docker version is useless.

Ahh, okay, ya didn't really look to much into it.  My goal was to grab something quick that might work.  Perhaps we need a more generic "requires_package" thingies then.  That way we can check e.g. the oci_umount version directly.  

> completely different NVRs for RHEL 7.3 and RHEL 7.4, not to mention Fedora.

That's just configuration stuff, ADEPT is more than happy to manage tracking those for us.  It's already got to fill in a ``subtests.ini`` template, adding another section for "expected version" isn't a big deal really.

cevich:
Okay, I'll probably drop this then, and implement a package NVR checker then.  I'm pretty sure this isn't the only case where having one would be handy.

cevich:
@edsantiago --force push just the version-collection part.

PR #743, opened 2017-07-21
Update DockerVersion with proper cache

cevich:
Previously this class's methods duplicated a lot of code because
values were buffered as instance attributes.  However some values
were actually cached as class attributes.  Fix both by moving the
buffer items into a class-attribute dictionary.

De-duplicate all the separate client/server methods to use common
private-methods.

Add a method for flushing the cache, to help support unittesting.

Update unittests to flush cache prior to each test.

[--args=docker_cli/version]

### DIFF
--- a/dockertest/output/dockerversion.py
+++ b/dockertest/output/dockerversion.py
@@ -214,7 +214,6 @@ class DockerVersion(object):
         """
         return self._version('server')
 
-
     @staticmethod
     def _require(wanted, name, other_version):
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago Split this off from the other PR, and re-implemented the storage so it's actually caching values now for usage within the same process.

cevich:
Closing/Reopening to force CI to re-run.

PR #744, opened 2017-07-21
Add class to cache NVRA queries from rpm command

cevich:
Add a read-only dict-like interface for querying RPM NVRA's by package
name.  Support caching of previous queries within the class so all calls
in the same process-space can benefit.  Add unittests for exercising the
new class's major features.

[skip-ci]

### DIFF
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -288,3 +288,98 @@ class CanonicalDistro(str):
     def is_enterprise(self):
         """STUB"""
         return 'rhel' in self or 'centos' in self
+
+
+class RDFilter(Mapping):
+
+    """
+    Filter-map storage type for RPM NVRAs and CanonicalDistro strings
+
+    :param args: Same meaning as for ``dict()``
+    :param dargs: Same meaning as for ``dict()``
+    """
+
+    # Internal storage, do not use
+    _mapping = None
+
+    #: Class required for nvra keys, expected to be compatible
+    #: with the RPM class.  If all fields are "", only
+    #: a distro. match will be considered.
+    nvra_type = RPMs.nvra_type
+
+    #: Set of acceptable distro-list values.  Where
+    #: ``None`` value matches the nvra key on any distro.
+    acceptable_distros = set(CanonicalDistro.distros) + set([None])
+
+    def __new__(cls, *args, **dargs):
+        if args or dargs:
+            argtest = dict(*args, **dargs)
+            for nvra, distros in argtest.iteritems():
+                if not isinstance(nvra, cls.nvra_type)
+                    raise ValueError("Initializer key '%s' is not a"
+                                     " '%s' or subclass." % (nvra, cls.nvra_type))
+                for distro in distros:
+                    if distro not in cls.acceptable_distros:
+                        raise ValueError("Initializer key '%s' value '%s', is"
+                                         " not in '%s'."
+                                         % (nvra, distro, cls.acceptable_values))
+        return super(Mapping, cls).__new__(*args, **dargs)
+
+    def __init__(self, *args, **dargs):
+        # N/B: Instances are immutable!
+        self._mapping = dict(*args, **dargs)
+
+    # Simply query internal storage to fill out ABC requirements
+    __getitem__ = lambda self, key: self._mapping.__getitem__(key)
+    __iter__ = lambda self: self._mapping.__iter__()
+    __len__ = lambda self: self._mapping.__len__()
+    __contains__ = lambda self, key: self._mapping.__contains__(key)
+
+    @classmethod
+    def from_csv(cls, nvra_distro_csv, nvra_delim=';', sub_delim='#'):
+        """
+        Return a new instance produced by parsing rpm_distro_csv.
+
+        :param nvra_distro_csv: A CSV list of ``delimiter`` separated key/value
+                                pairs.  Keys must be acceptable ``cls.nvra_type``
+                                initializers.  Values must be in
+                                ``cls.acceptable_values``.
+        # FIXME: More docs  (delims must never be valid distro or rpm names)
+        """
+        # count ; in "name;version;release;arch" the most complicated way possible
+        n_nvra_delim = len(cls.nvra_type._fields) -1
+        # Prune empty values and values w/o required delimiters
+        is_valid = lambda val: (val.strip() and
+                                val.find(nvra_delim) == n_nvra_delim and
+                                val.find(sub_delim) == 1)
+        # Dict-like's can be initialized from lists of key,value tuples
+        nvra_distros = [sub_val.split(delimiter, 1)
+                        for sub_val in [val.strip()
+                                        for val in rpm_distro_csv.split(',')
+                                        if is_valid(val)]]
+        # Catch eval() exceptions here instead of in __new__
+        try:
+            # Given nt = cls.nvra_type
+            #
+            # from_csv("foobar;42;;#fedora,baz;;;#rhel,;;;i386")
+            #
+            # returns
+            #
+            # [(nt('foobar', '42', '', ''), 'fedora'),
+            #  (nt('baz', '', '', ''), 'rhel'),
+            #  (nt('','','',''), 'i386')]
+            #
+            args = [(eval('cls.nvra_type(*%s)' % k.split(nvra_delim, n_nvra_delim)), str(v))
+                    for k, v in nvra_distros]
+        except TypeError:
+            raise TypeError("Error instantiating %s(%s): %s"
+                            % (cls.nvra_type))  # FIXME: finish this
+        # __new__ does final value validation
+        return cls(*args)
+
+    def fail_if_nvra_or_distro(self, extra_msg='', exception=RuntimeError):
+        """
+        Raise exception on distro and/or rpm nvra match w/ details + extra_msg
+        """
+        # Walk self.iteritems(), toss exception if match found
+        pass # STUB
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+
 """
 Low-level/standalone host-environment handling/checking utilities/classes/data
 
@@ -185,3 +186,105 @@ class RPMS(Mapping):
         """Flush cache of RPM NVRAs to force fresh queries on access"""
         cls._cache = dict()
         cls._alldone = False
+
+
+class CanonicalDistro(str):
+
+    """
+    Represents a canonical/standardized name for the OS distribution
+
+    :param object: Optional, same as ``str()`` built in.  If non-empty,
+                   take ``str(object`` as the definitive distro. name,
+                   bypassing automatic detection.
+    """
+
+    #: Ordered list of class-method names to use for distro-detection.
+    #: Each takes a dictionary representing the current detection state
+    #: (See as ``matches`` below), and is expected to return
+    #: a string representing the detected distro. name, None, or ''.
+    distros = ("fedora", "rhel", "centos")
+
+    #: Mapping of distro names to lower-case distro-method result strings
+    #: if auto-detection was used (i.e. no initialization arguments passed)
+    matches = None
+
+    def __new__(cls, *args, **dargs):
+        _object = cls._handle_object(*args, **dargs)
+        if _object:
+            return super(CanonicalDistro, cls).__new__(cls, str(_object).lower())
+        else:
+            _matches = cls.investigate()
+            distro_matches = set(_matches.values())
+            if len(distro_matches) > 1:
+                raise ValueError("Multiple distros detected: %s" % _matches)
+            elif len(distro_matches) != 1:
+                raise ValueError("No distros detected: %s" % _matches)
+
+            # immutables can't rely on __init__(), it must be done here
+            new_cd = super(CanonicalDistro, cls).__new__(cls, distro_matches.pop())
+            new_cd.matches = _matches  # __init__ can't handle this
+            return new_cd
+
+    @classmethod
+    def _handle_object(cls, *args, **dargs):
+        # Sigh, the str() builtin really does use a parameter named 'object' :_(
+        # Deal with this here, so __new__ only needs to worry about
+        # an optional string initialization parameter.
+        if len(args) > 1 or len(dargs) > 1:
+            raise TypeError("__init__() takes at most 1 argument (%d given)"
+                            % (len(args) + len(dargs)))
+        elif len(dargs) == 1:
+            if 'object' in dargs:
+                _object = dargs['object']
+            else:
+                raise TypeError("__init__() got unexpected keyword argument %s"
+                                % dargs.popitem()[0])
+        elif len(args) == 1:
+            _object = args[0]
+        else:
+            _object = ''  # default
+        return _object
+
+    @classmethod
+    def investigate(cls):
+        """
+        Return mapping of distro names to distro class-method results.
+        """
+        _matches = {}
+        for distro in cls.distros:
+            distro_match = getattr(cls, distro)
+            match = str(distro_match(_matches.copy())).lower()
+            if match in ('', 'none'):
+                continue
+            _matches[distro] = match
+        return _matches
+
+    # Method names must exactly match the cls.distros list.
+    # Called in order defined by cls.distros list.
+    # Use/Modification of ``matches`` argument is optional.
+    # Each must return a string, that will update ``matches`` for that distro.
+
+    @staticmethod
+    def fedora(matches):
+        del matches  # Optional, not used here
+        return None
+
+    @staticmethod
+    def rhel(matches):
+        del matches  # Optional, not used here
+        return None
+
+    @staticmethod
+    def centos(matches):
+        del matches  # Optional, not used here
+        return None
+
+    # General instance helper/convenience methods
+
+    def is_atomic(self):
+        """STUB"""
+        return 'atomic' in self
+
+    def is_enterprise(self):
+        """STUB"""
+        return 'rhel' in self or 'centos' in self
--- a/dockertest/environment.py
+++ b/dockertest/environment.py
@@ -9,6 +9,8 @@ Low-level/standalone host-environment handling/checking utilities/classes/data
 
 import warnings
 import subprocess
+from collections import Mapping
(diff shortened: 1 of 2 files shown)


cevich:
This is a working-prototype, my thinking is to add a method/function/class somewhere that will make use of this ``RPMS`` class, similar to how ``output.dockerversion.DockerVersion.requure_foo()`` operates.  In other words, you'll be able to initialize it with a list of acceptable distro + package NVRAs, then call something like ``requirements_met()`` (which will return or throw ``TestNAError``).  Something like that.

Oh, and BTW, that ``environment`` module has no autotest/docker autotest dependencies (on purpose) so it can be loaded/used from other scripts.  

Open to ideas/thoughts on implementing the ``requires(rpm)`` thingie I've gotta write next (including letting anyone else write it 😀)

dockertest/environment.py
```diff
@@ -79,3 +81,101 @@ def selinux_is_enforcing():
...
+class RPMS(Mapping):
+    """
+    Read-only mapping of rpm names to ``value_type`` instances
+
+    :param args: Same as standard ``dict()`` positional arguments.
+    :param dargs: Same as standard ``dict()`` keyword arguments.
+    """
+
+    #: Resentation for values, the first index must be the package name.
+    value_type = namedtuple('value_type',
```
edsantiago: Would you consider renaming this to `nvra`? "value_type" doesn't really convey a sense of what this holds.

dockertest/environment_unittests.py
```diff
@@ -0,0 +1,110 @@
...
+
+    exception = None
+    callcount = 0
+    command = None
+    name = None
+    version = '1.0'
+    release = 'a'
+    arch = '8088'
+    output_fmt = '%s#%s#%s#%s\n'
+    allnvras = tuple()
```
edsantiago: I think this should probably be `None`, but I do kind of like the idea of initializing it with a dummy value as a hint to the maintenance programmer. In that case, though, it should be `[tuple()]` (list), shouldn't it?

dockertest/environment.py
cevich: I was originally thinking of having this be a base-class, then specializing it as ``RPMS``.  That's why there's this (and other) generic names used.  Yes, happy to rename, I'm not married to much at all here.

dockertest/environment_unittests.py
cevich: Initializing a class attribute to a non-imutable is a no-no in python, I believe pylint will make quite a bit of noise about it.  It would be analogous to initializing a C pointer to some pre-determine address in a general program (not something specialized like a hardware driver or the kernel) instead of using ``malloc()`` or it's friends.  It causes lots of headaches, esp. in testing.  The best place for such things are in ``__new__()`` (for classes) and ``__init__()`` (for instances), sometimes both.

In this case, since there's really [only a single/special use for this](https://github.com/autotest/autotest-docker/pull/744/files#diff-185f8e9777f5988a94866cb9868e305bR93), maybe I should just move that, there.

cevich:
Updates for both concerns/suggestions applied in 669c0f7 (and I'm still not married to the names).

cevich:
^^^^ pylint fix

dockertest/environment_unittests.py
```diff
@@ -0,0 +1,109 @@
...
+    def __call__(self, command, *args, **dargs):
+        del args  # don't care
+        del dargs  # also don't care
+        self.callcount += 1
+        self.command = command
+        # Don't modify the input container value
+        self.name = list(command).pop()
+        if self.exception:
+            self.exception = None
+            raise self.exception("Raised by FakeCheckOutput")
```
edsantiago: I assume you meant to preserve `self.exception` before clearing it

dockertest/environment_unittests.py
```diff
@@ -0,0 +1,109 @@
...
+            self.environment.RPMS.nvra_type = original_type
+
+    def test_callout(self):
+        original_type = self.environment.RPMS.nvra_type
+        try:
+            self.environment.RPMS.nvra_type = staticmethod(lambda *args: tuple(args))
+            rpms = self.environment.RPMS()
+            sfco = self.fakecheckoutput
+            self.assertEqual(rpms['foobar'], (sfco.name, sfco.version, sfco.release, sfco.arch))
+            self.assertEqual(self.fakecheckoutput.callcount, 1)
```
edsantiago: why not sfco? (Here and line 87)

dockertest/environment_unittests.py
cevich: eww, how'd that get in there 😕 Thanks!

cevich:
fixed ^^^^^

dockertest/environment.py
```diff
@@ -79,3 +81,102 @@ def selinux_is_enforcing():
...
+                           % (' '.join(cmd),
+                              xcept.output.strip(),
+                              xcept.returncode))
+        for value in output.strip().splitlines():
+            try:
+                # Use a field-delimiter incompatible with any field value
+                new_nvra = cls.nvra_type(*value.strip().split('#'))
+                new_name = new_nvra[0]  # documented in class attr. defs. above
+                cls._cache[new_name] = new_nvra
+            except (TypeError, ValueError), xcept:
```
edsantiago: I see `TypeError` coming from `namedtuple`, but what throws `ValueError`?

Also, could you add a test for this? (I thought it would be simple-- `self.fakecheckoutput.output_fmt = "%s%s%s%s\n"` -- but I can't get that to work. It doesn't override the '#' one.)

dockertest/environment.py
```diff
@@ -79,3 +81,102 @@ def selinux_is_enforcing():
...
+    @classmethod
+    def _store(cls, name):
+        cmd = list(cls.rpmcmd) + [name]
+        xcept_fmt = ("%s did not return expected line-delimited"
+                     " items of hash-delimited fields as"
+                     " output: %s: %s")
+        try:
+            output = cls.execute(cmd)
+        except subprocess.CalledProcessError, xcept:
+            raise KeyError(xcept_fmt
```
cevich: Ok

dockertest/environment.py
cevich: > I see TypeError coming from namedtuple, but what throws ValueError?

IIRC, that comes from nvra_type if a field is missing (so, ``rpmcmd`` is wrong)

> Also, could you add a test for this?

If I just delete it (also) then no test is needed 😀

dockertest/environment_unittests.py
cevich: Hmmm, maybe a typo, lessee...yep.  Thanks!

cevich:
Addressed all concerns ^^^^^

> but I can't get that to work. It doesn't override the '#' one.

Got it working, it's misleading because the line gets parsed into a one-item-tuple.  I added a class attribute that records the expected size of ``nvra_type`` before the unittests swaps it for a tuple.  That allows dropping a ``ValueError`` if the ``split()`` length doesn't match expectations.  Not perfect, but should catch innocent blunders 😀 Added a unittest for this.

cevich:
OIC, nvm, I'm guessing you probably tried changing ``output_fmt`` on the ``FakeCheckOutput`` class, but it needs to be changed on the instance that's created in ``setUp()``.   In any case, it's fixed and better (IMHO) now with a unittest to make sure.

dockertest/environment.py
```diff
@@ -79,3 +81,95 @@ def selinux_is_enforcing():
...
+    def _store(cls, name):
+        cmd = list(cls.rpmcmd) + [name]
+        xcept_fmt = ("%s did not return expected line-delimited"
+                     " items of hash-delimited fields as"
+                     " output: %s")
+        output = cls.execute(cmd)
+        for value in output.strip().splitlines():
+            # Use a field-delimiter incompatible with any field value
+            new_nvra = cls.nvra_type(*value.strip().split('#'))
+            if len(new_nvra) != cls.nvra_len:
```
edsantiago: I don't think this can happen. This is what I referred to earlier when I mentioned TypeError:

    >>> t = namedtuple('nvra_type',('n','v','r','a'))
    >>> s = 'abc#def#ghi#jkl#mno'
    >>> x = t(*s.strip().split('#'))
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    TypeError: __new__() takes exactly 5 arguments (6 given)

cevich: It can, if the storage class (``nvra_type``) doesn't behave like the default ``namedtuple``.  This is done in nearly every unittest, where ``nvra_type`` is overloaded to be a regular tuple.  Probably a more complete way to validate the instantiation parameters,  would be to have the checker itself be mutable.  In anycase, I think that can be a detail left up to someone else to fix if ever needed.

cevich:
@edsantiago thanks for taking a peek.  Any ideas/suggestions on implementing the ``requires(rpm)`` thingie? Is that something you'd like to tinker with?

(my stomach's full of ADEPT-docs and sphinx, so not-that-hungry-really) 😀

edsantiago:
> Any ideas/suggestions on implementing the requires(rpm) thingie? Is that something you'd like to tinker with?

No and no. As I've mentioned before, I don't think docker-autotest is the right place for this checking.

cevich:
> I don't think docker-autotest is the right place for this checking.

Sure, that's fine, but no reason to let that assumptions get in the way of innovation.  That's a sure-fire way to shut down collaboration, which is useful beyond just the code, merged or not.

Okay, pressing onward then...

edsantiago:
Eek. I have no desire to stifle innovation: I simply have no interest in spending my time on something I've already looked into and can't find a sane way to implement. I am, however, eager to see what other minds come up with.

cevich:
OIC, so maybe "Creatively Exhausted" is maybe a better description.  Have you ever read "The Artists Way"?  At our level, the technical details are very mundane, its better to tap into the software-as-art mentality.  That book is teaches several methods for unblocking creativity, especially for writers.  Highly recommended, even if you only read the first few chapters 😀 

Perhaps this PR is an attempt that will fail.  I have no problem with that, because somewhere along the way of writing failures, the better-solution-spark will fire.  Maybe not even for me, but it will fire.  If it hasn't, it just means I/we haven't failed enough yet 😀

cevich:
...okay back on topic...

I'm looking at one possible destination:  A sub/sub-subtest wants to call ``something.requires(something or somethings)`` where ``something`` is maybe a mapping of platform-name to minimum package versions?  e.g. ``something.requires(fedora=[foo-1.0, bar-1.2], rhel=[foo-0.9, baz-13])``.

Another, possibly more flexible idea could be to take a CSV of comparison strings, and use ``eval()`` to parse them. e.g. ``something.requires(fedora="foo == '1.0', bar >= '1.2'", rhel="foo > '0.9', baz <= '13'")``

OFC the parameters could come from configuration or whatever, not important yet.

In all cases, it seems like some kind of simple distro-detection would be useful, so let's go there next...

cevich:
^^^ incomplete implementation of ``environment.CanonicalDistro()`` immutable.  Needs adding ``fedora()``, ``rhel()``, and ``centos()`` classmethods.  Or, throwing it all away and starting over with something simpler.

edsantiago:
AUGH! WAIT! Can you just merge this as it was (harmless), and start a new PR? Otherwise this has the potential to grow beyond anything that can sanely be reviewed.

cevich:
I suppose, though I kind of don't like the idea of having dead-code if it's avoidable.  Hmmm, I could make a new PR that depends on this one, that's no prob....

cevich:
--force push

cevich:
...blah, it is a prob, b/c it's a separate branch.  Can't you just look at the one new commit and review that?  In fact, it's not even worthy of a full review, just a quick sanity-check is fine.

cevich:
Yep, just checked.  If you click the 'commits' tab, and just pick one of them, it's possible to review them individually: https://github.com/autotest/autotest-docker/pull/744/commits/bbed735f67fd0febd4208d940b9e440a44be99e4

I'm not going to touch that for the rest of today, got other stuff to work on.

edsantiago:
> If you click the 'commits' tab, and just pick one of them, it's possible to review them individually

I know, and I can do the same with git-pr (which I strongly prefer). It's just that heaping on more and more changes on one single PR makes it difficult for future maintainers, too: they will have one enormous merge instead of multiple smaller ones.

Anyhow, that said, I need to switch gears. I've spent many hours over two days trying to understand this latest change and still can't figure out how it's meant to be used or why. I need to move on to work for our Westford meeting.

For the sake of future maintainers, could I please ask you to rename variables as follows?

    detectives -> distros
    victims -> matches
    evidence -> hint (or something actually descriptive; I have no clue what this is supposed to be)
    spyglass -> distro_match

And perhaps also add @staticmethods for each of fedora, rhel, centos into environment.py so there's a reasonable predefined set of helpers?

I also made big changes to unittests.py, changing foo/bar/baz to fedora/rhel for consistency and sanity, but I can't figure out how to push those changes onto this PR. I would encourage you to simplify some of those tests, adding real-world examples that could illuminate the intention of this class.

Thank you!

cevich:
Not really, I normally straight-merge instead of squash-merge, so the individual commits will all stay intact.  

> And perhaps also add @staticmethods for each of fedora, rhel, centos into environment.py so there's a reasonable predefined set of helpers?

OIC where the confusion is...sorry I wasn't clear, yes this is far from complete.  I'm pushing the little half-baked, incremental pieces up early, draft-stage stuff, not polished and finished.  In fact it's pretty raw.  Just in case there's some giant glaring design-problem I've overlooked.

Yeah, don't spend hours and hours reviewing, that's way to much.  Quick 5-minute glance, if that.  There are more puzzle-pieces needed, that's why you can't see the complete picture, it ain't there 😀 

> For the sake of future maintainers, could I please ask you to rename variables as follows?

...and for your sake :D  Please consider all my names as "placeholders".  In fact, perhaps I should just start using strings of random letters and numbers 😀 I'm seriously not married to any of them.  I just pick stuff that makes sense to me at the moment, not for the long-term.  I'm always happy to change them, esp. when your names usually make more sense anyway. 

> I can't figure out how to push those changes onto this PR.

Oh, that would be a big help!  This is a key thing to know!  It's how multiple people can serialize their changes up into a single (upstream) PR.  The secret is:

Push the changes up as per usual, but before submitting the PR, change the base-fork (at the top, left-side) to my repo. (``cevich/autotest-docker``) and the branch to ``require_rpm_nvr`` (just as it says on the RHS, top of this PR).  The right-side should be your repo/brnach.

In other words, you're opening up a PR from your copy (on the right),  against my copy (on the left).  Effectively, it's a request to change my base-branch that was used to make this PR.  Then I review those changes, merge them, and *poof*, they show up here with your name on 'em.

I won't change or push anything here until you figure it out, so there's a stable HEAD.    Ping me if you still can't make it obey.

cevich:
Oops, shoulda squash-merged, oh well.  Anyway, there ya go, your updates above ^^^^

cevich:
@edsantiago ^^^ some changes...sorry, probably more than you'd like 😢 I think I need more coffee...

cevich:
Hint: if you absolutely HATE what I did in that ``fixup!`` you can revert as part of another PR to my branch 😁  We can always do a final rebase/squash away the unnecessary commits, then --force push as a final step before merge (i.e. so ``fixup!`` and ``squash!``, ``rebase`` commits "go away").

cevich:
/me takes hands off steering wheel, removes it, and passes it over to @edsantiago 😀

dockertest/environment.py
```diff
@@ -79,3 +82,197 @@ def selinux_is_enforcing():
...
+    distros = ("fedora", "rhel", "centos")
+
+    #: Mapping of distro names to lower-case distro-method result strings
+    #: if auto-detection was used (i.e. no initialization arguments passed)
+    matches = None
+
+    def __new__(cls, *args, **dargs):
+        _object = cls._handle_object(*args, **dargs)
+        if _object:
+            return super(CanonicalDistro, cls).__new__(cls, str(_object))
```
edsantiago: I'm not sure what the purpose of this is. It also breaks unit tests. Could you try again?

cevich: oh ya, I never updated the unit tests.  Sure I can get them working again...

Purpose: This is preserving the base-class (``str``) API.  It's analogous to any other ``super`` call, except it's polymorphic on the class-object itself.  Having it supports, e.g.:

```python
foobar = CanonicalDistro('foobar')
if foobar.is_enterprise():
    do_something_useful()
```

cevich:
@edsantiago ^^^ there ya go, unittests pass now (didn't run pylint).  Okay, keeping my hands off the wheel again.

edsantiago:
Okay... I'm still not clear on what problem this will solve, but it passes selftests so let's just merge it and then build on it as needed. LGTM. Thanks!

cevich:
Okay Cool, will do.

It's not necessarily solving an immediate problem really, it's more for conforming to the base-class API.  So if the caller already knows the canonical distro name, and wants to make a CanonicalDistro instance because some useful instance-method exists, there's no need to waste time/resources calling the discovery-methods again.  The instance can simply be initialized straight from the string, because by definition, we've made CanonicalDistro a "type" of string.

cevich:
Next steps seem to be:

- Write a few/all distro methods (esp. RHEL)
- Bring ``CanonicalDistro`` into use by ``SubBase.failif_not_redhat()``
- Make sure that works with existing tests (i.e. run them)
- Decide if an intermediate ``environment.something`` class is needed...
- ...then add ``SubBase.require_packages()``.
- add unittests for ``SubBase.require_packages()``

cevich:
@edsantiago I just rebased, only the last commit is new (and incomplete/unfinished).

TODO:

* Complete ``RDFilter`` class
* Add unittests for RDFilter class
* Bring CanonicalDistro into use by SubBase.failif_not_redhat()
* Add to ``Subtest`` and ``SubSubtest`` ``initialize()`` methods
  ``RDFilter.from_csv(self.config.get('rdfilter_csv', '').fail_if_nvra_or_distro(exception=DockerTestNAError)``

The last bullet allows adding the filter to any subtest or sub-subtest via the standard configuration interface, and does nothing if the key isn't present.

The unittests for ``RDFilter`` should be pretty easy since it has a standard dict-like interface, and all interaction points can be easily instrumented w/o shenanigans.

cevich:
abandonware

PR #746, opened 2017-07-26
Add supporting class for testing with bash

cevich:
Sometimes we're in a hurry, and just need to toss a quick and
simplistic test using only shell-scripts.  Support this
with a new ``BashSubtest`` class and example implementation
in ``subtests/bashexample/``.

[skip-ci]

### DIFF
--- a/config_custom/control.ini
+++ b/config_custom/control.ini
@@ -1,0 +1,83 @@
+# Autotest-control configuration, not consumed by dockertest or subtests
+# directly.  Though job.resultdir will contain an updated reference copy.
+
+[Control]
+
+# Subtests/sub-subtest names to run (CSV, in order)
+# **instead** of searching the subtests directory.
+# Any also specified by the --args i=<csv> sub-option and
+# are not on this list, will prefix resulting list (also in order)
+
+# e.g. "docker_cli/version"
+include =
+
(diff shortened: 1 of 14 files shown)


cevich:
**N/B:** Untested, just an early prototype of an idea.

edsantiago:
Obvious first question: how does one script pass state to the next?

cevich:
> Obvious first question: how does one script pass state to the next?

Simple, it's left up to the script author.

This isn't about supporting polished, long-term maintainable tests, it's only providing an extremely low entry-barrier for situations/people where that's helpful.

cevich:
--force push got it actually working with some more sane-logging.

cevich:
**N/B**: To be clear, NONE of the changes from ``FUDGING FOR A PR/CI TEST - DO NOT MERGE`` are going in.  That was just a look-and-feel test.

---

After playing with that ``dummy/test``, it seems alright.  It's easy to copy-paste from the ``bashexample``, and it supports the ini-configuration right out-of-the-box.  IMHO, it's possibly something engineering or an "outsider" would appreciate, for the occasional rattle-trap / quickie-contribution.

cevich:
abandoned, can re-open if ever needed.

PR #747, opened 2017-07-27
ci: rename to new name

jlebon:
Rename file to new name and update link to docs.


edsantiago:
bot, test pull request

(Not that I see anything that would get tested anyway. LGTM)

PR #748, opened 2017-07-27
Get rid of useless warning message

cevich:
An pedantic-check for a non-float value from a test's configuration
prints a warning that's not very useful in most cases.  Kill it.


cevich:
Ya, I tried figuring it out too.  Not sure if just killing it will work on it, hopefully it will.  If not, maybe I go back and read the old commit message and try and figure why I added it to begin with 😖

cevich:
*shrug* Hmm, it must be okay.  Great.

PR #750, opened 2017-08-03
Template for creating new tests, cheaply

edsantiago:
This provides a new bash-script scaffolding that developers
can grab, modify, and submit to QE for inclusion in autotest.
The scaffolding provides logging, failure checks, helper
functions for testing command output, and various examples.

The idea is to publish this in an easily-accessible location,
possibly part of the Trello QE checklist. Developers can
grab this as part of a bug/feature fix, quickly add a
sanity test for their code, and submit the script to QE.
QE can then use as-is or, if warranted, rewrite into
the docker-autotest framework.

Initial commit is entirely docker-centric because docker is
the only tool we currently test. We need to have some
conversations about that.

[skip-ci]

### DIFF
--- a/subtests/example/example.sh
+++ b/subtests/example/example.sh
@@ -1,0 +1,149 @@
+#!/bin/bash
+#
+# FIXME - add a one-line description of what this script is supposed to test
+#
+# FIXME - possibly add a more verbose multi-line description here. Not
+# strictly necessary but possibly helpful to future maintainers. You
+# could, e.g., include a BZ ID and/or a description of the behavior
+# this test script is intended to check.
+#
+
+###############################################################################
+# BEGIN setup
+#
+# You probably don't want to touch anything here, but you may want
+# to skim it to see the available helper functions
+#
+
+# Don't touch. This lets the docker() wrapper log debug statements
+# to original stdout, while also preserving a context-specific stdout
+exec 5>&1
+
+# In case you need a temp dir, e.g. for bind mounts or Dockerfiles
+TMPDIR=$(mktemp -d --tmpdir docker-test-sample.XXXXXXXX)
+if [ -z "$DEBUG" ]; then
+    trap 'cd /;/bin/rm -rf $TMPDIR' 0
+fi
+
+# Docker image on which to test. This will probably be overridden in ADEPT.
+IMG=centos
+
+die() {
+    echo "$*" >&2
+    exit 1
+}
+
+# Helper for the helpers: returns a full path to a stdout or stderr log file
+LOG_COUNTER=0
+commandlog() {
+    printf "%s/log.%s.%02d" $TMPDIR "$1" $LOG_COUNTER
+}
+
+# When debugging failed tests, it is crucial to know the exact commands
+# that were run. This is a trivial wrapper on the docker command. It:
+#
+#    * emits its full arguments to stdout, in a way that the docker-autotest
+#      caller will detect and log for human readers;
+#    * invokes docker with those arguments, but logs stdout and stderr
+#      to temp files
+#    * outputs stderr, if any, to caller's stderr
+#    * outputs stdout to caller's stdout
+#    * returns with docker's exit status
+#
+# FIXME-QE: refactor to make it reusable for kpod, buildah; not just docker
+#
+docker() {
+    echo "DEBUG:docker $@" >&5
+
+    LOG_COUNTER=$(expr $LOG_COUNTER + 1)
+
+    log_stdout=$(commandlog stdout)
+    log_stderr=$(commandlog stderr)
+
+    command docker "$@" >$log_stdout 2>$log_stderr
+    docker_status=$?
+
+    cat $log_stderr >&2
+    cat $log_stdout
+
+    return $docker_status
+}
+
+# Helper function: tests for presence of given string in docker output
+# stream (default: stdout), aborts with diagnostic if absent.
+expect_string() {
+    local expect=$1
+    local stream=${2:-stdout}
+    local logfile=$(commandlog $stream)
+
+    grep "$expect" $logfile && return
+
+    die "TEST_FAIL:Did not find expected string '$expect' in $stream '$(<$logfile)'"
+}
+
+# Helper function: opposite of the above. Tests for presence of given string
+# in docker output stream (default: stdout), aborts if string is present.
+die_if_string_present() {
+    local string=$1
+    local stream=${2:-stdout}
+    local logfile=$(commandlog $stream)
+
+    found=$(grep "$string" $logfile)
+    test -z "$found" && return
+
+    die "TEST_FAIL:Found unwanted string '$string' in $stream: '$found'"
+}
+
+# FIXME-QE: if needed, add equivalents for regex
+# FIXME-QE: if needed, add non-die equivalents so further tests can proceed
+# FIXME-QE: if needed, add docker_require(), return TEST_NA if version < wanted
+
+# END   setup
+###############################################################################
+# BEGIN single docker command
+#
+# This section is for tests that can be run with a single docker command
+# that terminates immediately and whose output or exit status can be
+# checked. See later below for commands that require a detached docker
+# command that you can interact with.
+
+# Run a command that's expected to terminate, e.g.:
+#
+#   mkdir $TMPDIR/mysubd; docker -v $TMPDIR/mysubd:/xyz run $IMG ls -lZd /xyz
+#   docker ps -a
+#   docker --userns=host run $IMG /bin/crashme
+#   storage_stuff=$(docker info | grep -A3 'Storage Driver')
+#   ...
+
+docker run --rm -v $TMPDIR/nonexistent-subdir:/mumble $IMG date || die
+
+# Check for errors in output; also make sure our desired output is present
+die_if_string_present "Error running date command"
+expect_string "Today is the first day of the rest of your life"
+
+# END   single docker command
+###############################################################################
+# BEGIN detached docker command
+#
+# Use this for starting a detached container, then interacting with it.
+# For instance: sending signals, or engaging with a server.
+
+CID=$(docker run -d $IMG bash -c 'while :;do date; sleep 2;done')
+
+# FIXME: interact with container
+count1=$(docker logs $CID | wc -l)
+sleep 5
+count2=$(docker logs $CID | wc -l)
+
+if [ $count1 -eq $count2 ]; then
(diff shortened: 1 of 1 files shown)


subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
...
+    local string=$1
+    local stream=${2:-stdout}
+    local logfile=$(commandlog $stream)
+
+    found=$(grep "$string" $logfile)
+    test -z "$found" && return
+
+    die "TEST_FAIL:Found unwanted string '$string' in $stream: '$found'"
+}
+
```
cevich: Having separate definitions of all the above functions embedded here, means fixes/additions will be missing from *all* copies.  I'd prefer them to be defined in some central place, then sourced into this script instead.

subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
...
+# You probably don't want to touch anything here, but you may want
+# to skim it to see the available helper functions
+#
+
+# Don't touch. This lets the docker() wrapper log debug statements
+# to original stdout, while also preserving a context-specific stdout
+exec 5>&1
+
+# In case you need a temp dir, e.g. for bind mounts or Dockerfiles
+TMPDIR=$(mktemp -d --tmpdir docker-test-sample.XXXXXXXX)
```
cevich: This should probably be exported

subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
+#
+# FIXME - add a one-line description of what this script is supposed to test
```
cevich: How about:

```
# FIXME - add a one-line description of what this script is supposed to test
# then go to near the bottom of this file to write you test.
```
or something to that effect.  Prevents needing to read the entire script to understand what needs to be done, and where.

cevich:
So the idea is, this script can be run stand-alone, and then later, an autotest-guru groks and converts it into an actual subtest?

Assuming so, I do like the core-idea here: Having something that doesn't require much learning, and can be tossed together quickly.  However, I think that without translation into a subtest, it will be very low-fidelity to the environment that matters: _CI_

#746 (while not perfect, nor complete) would at least provide CI-environment feedback, because it would run by the Docker Autotest PR-tester.

**How could something similar be done with a script like this, to provide rapid feedback from within an automated-environment?**

subtests/example/example.sh
edsantiago: Great idea. So done.

subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
...
+###############################################################################
+# BEGIN setup
+#
+# You probably don't want to touch anything here, but you may want
+# to skim it to see the available helper functions
+#
+
+# Don't touch. This lets the docker() wrapper log debug statements
+# to original stdout, while also preserving a context-specific stdout
+exec 5>&1
```
edsantiago: So done.

subtests/example/example.sh
edsantiago: Probably not: there is no need for any subprocesses to access it. However, TMPDIR is absolutely the wrong variable name choice: that's a reserved name! I've changed it to TEST_TMPDIR. Thanks for highlighting this.

subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
...
+
+# In case you need a temp dir, e.g. for bind mounts or Dockerfiles
+TMPDIR=$(mktemp -d --tmpdir docker-test-sample.XXXXXXXX)
+if [ -z "$DEBUG" ]; then
+    trap 'cd /;/bin/rm -rf $TMPDIR' 0
```
edsantiago: So done. Thanks, I didn't know about  EXIT.

subtests/example/example.sh
```diff
@@ -0,0 +1,149 @@
...
+fi
+
+# Docker image on which to test. This will probably be overridden in ADEPT.
+IMG=centos
```
edsantiago: As part of the conversion process: one of us in QE will take this script, incorporate it into docker-autotest, and from there use the ADEPT-defined test image.

subtests/example/example.sh
edsantiago: That goes against the entire point of this script: the idea is for a developer to grab this one single file, play with it, toss it over the net to us. They don't need to check out docker-autotest nor anything else. Just one single standalone file.

edsantiago:
>So the idea is, this script can be run stand-alone, and then later, an autotest-guru groks and converts it into an actual subtest?

Exactly. In the docker-autotest framework. Can you rephrase your "low-fidelity" comment? I don't understand how you go from "this will be written as a subtest" to "this won't be a subtest". It will be a subtest within docker-autotest, it's just that the subtestification will be done by QE.

subtests/example/example.sh
cevich: Checking out repos. and wrangling git is part of an engineers regular responsibilities.  It's so regular, in fact they'd probably check out the repo to get the script anyway.  This is not something worth arguing over, it's totally normal to expect a repo. checkout.

cevich:
> Exactly. In the docker-autotest framework. Can you rephrase your "low-fidelity" comment? I don't understand how you go from "this will be written as a subtest" to "this won't be a subtest". It will be a subtest within docker-autotest, it's just that the subtestification will be done by QE.

low-fidelity: adj. antonym of high-fidelity: https://en.wikipedia.org/wiki/High_fidelity

In all cases (even #746), having this working out-of-the-gate within a CI / Automated environment is a must-have feature.  Here's a proposal, that would allow us to have any cake AND munch them all:

*  Move these scripts [into ContainerTest (internal repo)](https://hurl.corp.redhat.com/fc42037).
*  Relaxes the review / scrutiny overhead by an order of magnitude.
*  Places it right next to the automation facilities, which include a stand-alone execution tooling.
*  We can have as many different kinds of "QuickTest" scripts as we like, all 31-flavors if need be.
*  It opens the option of having control over the environment part of QuickTest.  For example, selecting the types / configuration of hosts it should run on.  Optionally.
*  It leaves open the option of having a QE person add the test-pieces into Docker Autotest, or wherever.
*  It supports adding execution into [the new fedora CI framework.](https://fedoraproject.org/wiki/CI)

Here...I'll hack together a quick prototype...

subtests/example/example.sh
edsantiago: It's not the git repo, it's having to set up an include path so the template script can find includes, and copying those includes to a test system. The point of this exercise is to have one single file. We'll talk about it at the f2f, I guess, but I am not going to split this. It would defeat the entire purpose of having something simple.

cevich:
...prototype (internal) PR that runs your ``example.sh`` stand-alone, within an automated environment (same as used for Docker Autotest):  https://hurl.corp.redhat.com/e08c420

(doing a test-run right now)

cevich:
Two --fixup's later, and this is working for me in ContainerTest repo. (above):

```
[cevich@cevich containertest]$ cat /tmp/workspace/results/fedora-25-docker-latest/example.sh.output 
DEBUG:docker run --rm -v /tmp/docker-test-sample.UITMkZ63/nonexistent-subdir:/mumble centos date
Mon Aug  7 14:43:04 UTC 2017
TEST_FAIL:Did not find expected string 'Today is the first day of the rest of your life' in stdout 'Mon Aug  7 14:43:04 UTC 2017'
```

edsantiago:
I'm still puzzled, but I think it's because I haven't adequately explained how I see this working out in practice. I hope I can communicate better this time.

The key takeaway from my purpose is that a developer does not need to know autotest, docker-autotest, nor ADEPT. All they have to do is fill in a few comments and commands, something that can give a QE engineer a working model with which to write a docker-autotest test. Put another way: developer takes example.sh, fills it in, and emails it (not git-commit nor anything like that) to you or me. We then rewrite it into a docker-autotest subtest: find the right subdirectory for it, split it into initialize / run_once / cleanup as needed, and put it into production. The delivered example.sh will never actually be checked in or run in production.

Does that make more sense?

subtests/example/example.sh
cevich: Maybe "IMG" could be passed in as an environment-variable or as an optional parameter?

e.g. 
```bash
[ -n "$IMG" ] || IMG="centos"
```

w/n ADEPT is running this, probably passing in parameters is the best way to go.  Unf. parsing arguments w/in bash really sucks. Esp. if there're anything but trivial single-minus, or positional parameters.  Another place where having a standard "library" or include-file would be super handy.

subtests/example/example.sh
cevich: Hmm.  What about reversing the layout somehow, so the primary "stuff to do" is at the top instead of the bottom.  That would allow hiding all the common stuff (functions, etc.) from view, by require one to "scroll down".

It would also allow adding an arbitrary number of utility functions while having all the important stuff an engineer is suppose to care about, live right at the top.  Then, have the script include itself (``$0``) before calling main (somehow).  Maybe it sets a "magic" env. var then tests it, so it doesn't end up in an infinite loop.

Just an idea.

cevich:
> Does that make more sense?

I understood the first-time around 😀   I'm generally in agreement it's a useful approach.  Did you read the [description in the PR link](https://hurl.corp.redhat.com/e08c420)?  

The workflow is exactly the same but with optional side-benefits.  It's not enough to expect running this on a or some random VM, represents a reproducible environment (QE involvement or not).  Garbage-in-garbage-out applies at the macro-level as well 😀

Having the script exist w/in ContainerTest doesn't imply any understanding of ADEPT or DockerAutotest or anything else.  It simply means the environments the script runs in, are high-fidelity with the eventual (QE edited/crafted) CI environment.   So...much less "garbage".

cevich:
There, [edited PR's description](https://hurl.corp.redhat.com/e08c420) to make it even more clear.  Give it a try, it works for me.

cevich:
Okay, more thinking...closed (internal) PR, we need the example in public somewhere so it's supported by Fedora.  It Looks like putting this in our 'container runtime' or a higher-level module may be the way to go.  That solves the all the manual + automated problems nicely, and lets this remain a self-contained easy-script. 

Let's connect with @lsm5 to figure out if he's got either an official or temporary repo. space for it to live in for Fedora, and just kick the RHEL "can" down the road (until a repo. emerges)

cevich:
@edsantiago & @lsm5 is there a high/mid-level module repo. where this example script can live for use?

If not, I'm not opposed to sticking it into a 'contrib' style directory, or if that would be easier.

cevich:
Closing as trello-card archived.  Can re-open if needed.

PR #752, opened 2017-08-15
Temp. Workaround broken CentOS/EPEL pylint package

cevich:

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -34,7 +34,9 @@ packages:
     - git
     - make
     - python-bugzilla
-    - pylint
+    # Temporarily workaround broken dependency in EPEL:
+    #     pylint requires python-astroid >= 1.4.5 but is missing (BZ 1479018)
+    # - pylint
     - python-pep8
     - python2-mock
     - python-sphinx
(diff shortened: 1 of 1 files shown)


.papr.yml
```diff
@@ -55,6 +55,8 @@ env:
 # present, tests will run after a successful build. Full
 # UTF-8 values are supported.
 tests:
+    - yum install -y --skip-broken pylint
+    - yum downgrade -y pylint
```
edsantiago: This is the sort of thing where I like to add big screaming comments explaining what's going on, why, and when it should be time to revisit / review / remove this. Likewise for the commented-out pylint above. Aside from that, LGTM but we'll have to see if it works.

cevich: Yes, this is just a 'see if it works' thing.  Skookum comments (with spell-checking) can come later 😁

cevich:
This is the thing that breaks:

``--> Processing Dependency: python-astroid >= 1.4.5 for package: python2-pylint-1.6.5-4.el7.noarch``

cevich:
Okay, @edsantiago I ``--force`` pushed with better comments, and it's working.  Though...uggg I never spell checked 😖

PR #756, opened 2017-08-28
oci-umount: handle blank lines, wildcards

edsantiago:
oci-umount in docker-1.12.6-55 introduces two wrenches: comment
and/or blank lines, and the use of '*' as a wildcard:

    # blah blah

    # blah
    /var/lib/docker/containers/*

Comments and empty lines are easy: ignore them.

Wildcards: will always be at the end, in the form 'dir/*',
meaning 'dir is OK but unmount anything underneath'. This
too is easy: strip off the asterisk but leave the trailing
slash. This will let us alert if we see dir/subdir mounted
but will not trigger on plain dir.

[--args=docker_cli/run_volumes/oci_umount]


cevich:
Did you see the 'fedoratomic26-docker-latest.subtests.docker_cli.run_volumes' failure?

Also, I'll edit the description here so only the run_volumes test runs in CI...

cevich:
jenkins, test this please.

edsantiago:
> Did you see the 'fedoratomic26-docker-latest.subtests.docker_cli.run_volumes' failure?

No... I actually never look at CI results any more. They almost always fail, and are really painful to track down, and every time I've tried it has been something unrelated to my change. I run my own tests on my own VMs, and I trust those.

I just waited for this CI run, and see the run_volumes failure on fedatomic, but unfortunately without a way to ssh into the VM I have no way to track it down further.

edsantiago:
I've spun up a Fedora Atomic 26 virt, set up my tests on it ... and the failure seems to be because the docker rpm (specifically, oci-umount) is too old. So this is actually an expected failure.

cevich:
> No... I actually never look at CI results 
> ...
> I've spun up a Fedora Atomic 26 virt

Our directive is: **If it's not automated, it's broken.**  There's no more room for fly-by-night manual testing outside of controlled environments, it doesn't scale, and ends up wasting everybody's time (one way or another).

[Instructions on manually running any automated test are well documented.](https://hurl.corp.redhat.com/c9ffead)  If you can't figure it out, ask for help.  If it's broken, fix it.  If you can't fix it, ask for help.  Simple as that.

edsantiago:
> Our directive is: If it's not automated, it's broken. 

Of course. That's our goal but we're a long way from it still. For instance, if I understand your implication correctly, you're requesting that this PR be blocked until it passes on all ADEPT hosts. Is that correct? If so, can you recommend a way to do that? We don't, at this time, have a mechanism for skipping or ignoring tests on known-bad environments. If I've misunderstood your comment, can you please clarify?

cevich:
> Of course. That's our goal but we're a long way from it still.

Mostly.  To clarify: Yes, I'd like the Docker Autotest CI job to pass at least the affected test(s) on a PR, on all hosts.  We don't get to 100% PASS without "continuous" efforts on either the tests, configuration, or the environment.  I cannot do that alone, it won't scale.  If something's bustid for a PR, it's gotta be addressed so the failures don't simply continue to propagate through automation.  As always, I am more than willing to help.

edsantiago:
So in essence we have tests that are breaking because one host has a known-bad RPM; and we have no way of updating that RPM. This is exactly what #679 was intended to address. I don't know any other way to address it.

cevich:
> a known-bad RPM

Hmmm...interesting problem.  There's got to be a solution since the inputs that break it are known....<idea brewing>...

cevich:
@edsantiago remind me, why the known-wonky RPM can't be detected from w/in the test (besides being a PITA)?

cevich:
thinking: There's a 10-ton hammer on line 150: https://github.com/cevich/autotest-docker-enabled-product-testing/blob/master/kommandir/peon_setup.yml#L150

As a short-term fix, we could toss some scripting in there to cause this test to skip on hosts with the afflicted RPM installed.  Could something like that work?

edsantiago:
>remind me, why the known-wonky RPM can't be detected from w/in the test (besides being a PITA)?

It just offends me deeply: adding temporary code that runs `utils.run("rpm -q docker")`, compares it to the exact version on one particular ADEPT host, then remembering to remove that code (via another PR) later on. It clutters up the code and the git history. I can write horrible code at times but I'm not willing to go that far.

>As a short-term fix, we could toss some scripting in there to cause this test to skip

IIUC that would have to be some sort of `sed -i` that edits `control.ini` and adds to the exclude list? That too seems fragile.

I'm going to take some time next week to revisit #679 and see if there's some other way to do it.

cevich:
> That too seems fragile.

I'm totally cool with short-term fragile fixes.  My motivation here is getting this specific PR merged.  There's no reason to hold up a PR if we can find a quick/dirty solution.  That gives us time to consider a better more permanent fix while realizing the benefits of having a running test.

> code that runs utils.run("rpm -q docker")

Even doing it w/in a test isn't _that_ bad, it just becomes dead-code once releases move past that specific version.  We have plenty of other places that check version-numbers.  Doing this at the package-level isn't really all that different.

cevich:
Long-term fix: IIRC, #695 was even closer than #679 since I added unittests.  Might be faster to pick up / fix- that one instead.

cevich:
@edsantiago after more pondering, I updated #744 after realizing a possible fatal flaw to the after-the-fact approach (#695 or #679 or others):

     An unfixable bug which is fatal to the system (or any other tests) will trigger Armageddon.

Like Nuclear deterrents, a low risk but devastating consequences.  The only mitigation is to not run that test.  That puts us back in the seat with the up-front approach (#744), which also preserves all result/reality fidelity.

edsantiago:
@cevich IIRC this one is awaiting the approval of #761. I'm done for the week but if you get a chance tomorrow to approve #761, can you follow up with this one? You may need to add certain magic lines to the known_failures.txt file; unfortunately I don't know what those are because I can't get to the ADEPT results any more (timeout waiting for jmaster). Thanks in advance.

cevich:
@edsantiago thanks for the reminder

cevich:
@edsantiago all is merged and I confirmed the known_failures.txt file is being copied.  You have commit access to open PRs and merge them straight to master (hopefully after the CI tests pass):

https://github.com/cevich/ADEPT/tree/master/kommandir/roles/autotested/files

edsantiago:
jenkins, test this please

edsantiago:
jenkins, test this please

edsantiago:
[force-pushed rebase onto master, which now includes the overlook_known_failures code]

edsantiago:
@cevich it failed because of this:

    SubBase: Skipping known_failure check: [Errno 2] No such file or directory: '/var/lib/autotest/client/tests/docker/config_custom/known_failures.txt'

Is there an easy way to tell if this CI environment includes the latest ADEPT code for injecting `known_failures.txt`?

cevich:
> Is there an easy way to tell if this CI environment includes the latest ADEPT code for injecting 

Damn.  Taking a peek...crap crap...something's broken: https://hurl.corp.redhat.com/4088e1c  File should be there but isn't :S  -sigh-  It was working when in my ADEPT PR (I checked).  Okay, digging...

cevich:
jenkins, test this please.

cevich:
jenkins, test this please.

cevich:
Ahh, okay! now should be working...

```
00:00:27.473 Reset autotest to origin/master commit a5cf4bff6d2a280e18174ae2a105a74f5ba57c3e
00:00:27.673 Reset adept to origin/master commit 5547d7f573c408014dfb23902506e5196d0ec6a8
```

cevich:
w00t!

PR #757, opened 2017-08-29
Workaround: cope with changes in ls output

edsantiago:
coreutils 8.25(?) introduces unexpected behavior changes in
output: many file paths are now in quotes:

    - ls: cannot access /foo/bar: No such file or directory
    + ls: cannot access '/foo/bar': No such file or directory

We now accept either form.

### DIFF
--- a/subtests/docker_cli/run_attach/run_attach.py
+++ b/subtests/docker_cli/run_attach/run_attach.py
@@ -19,8 +19,11 @@ from dockertest.dockercmd import DockerCmd
 from dockertest.images import DockerImage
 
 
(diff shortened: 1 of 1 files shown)


cevich:
Do we really care about checking quotes?  I think only the ENOENT is the important part...nevermind, I see your comment now.  okay, LGTM.

To make CI run **_MUCH_** faster, be sure to stick the magic string in the PR description.  For example, in this one you could use ``[--args=docker_cli/run_attach]`` so it only runs just this one test.  It accepts [any argument string as described in the docs.](http://docker-autotest.readthedocs.io/en/latest/#selecting-sub-test-and-sub-subtest-modules)

PR #758, opened 2017-08-29
run_cgroup_parent test: don't require a controller

edsantiago:
Fedora 26 /proc/1/cgroup includes the line '0::/init.scope',
i.e. the second field is empty. Our code was not expecting
that. Deal with it.

[--args=docker_cli/run_cgroup_parent]


cevich:
jenkins, test this please.

cevich:
I think we can call this a CI "PASS" despite ``rhelatomic7-4-docker.pretests.docker_test_images`` dying due to timeout (we can't control that).  All other hosts ran perfectly.

PR #761, opened 2017-09-13
Overlook known problems in docker builds

edsantiago:
Certain subtests are known to fail under certain specific
docker builds. This is an inevitable part of life in a
fast-paced development environment.

Test failures are a distraction to humans: they waste time
and energy, often multiple days in a row because it's hard
to remember which failures we've seen before. This PR
introduces a lightweight mechanism for ignoring known
failures, by checking (subtest, docker-nvra) against
an easily-maintained list. When we find a match, we
change subtest failures to WARN and sub-subtest to PASS.

The key is an input file, known_failures.txt, inserted
into the config_custom subdirectory by ADEPT or other
test scaffolding. On any subtest or subsubtest failure
we read this file and cross-check the failing test
against the list.


cevich:
Oh cool @edsantiago, you got it working!  I'm pretty busy today, but will make room to squeeze this in.  I'm only going do a really quick review/test b/c we've already got lots of history with this code, no need to re-hash...

edsantiago:
Thanks! The subsubtest code is pretty much the same; the big difference is subtests: in `cleanup()` I check for a recent exception and, if it's a `DockerXxxx` one, and subtest/nvr are on the known-failures list, override the `exit_status` field in the exception.

dockertest/subtestbase.py
```diff
@@ -102,10 +161,84 @@ def cleanup(self):
         Always called, before any exceptions thrown are re-raised.
         """
         self.log_step_msg('cleanup')
+        self.overlook_known_failures()
```
cevich: @edsantiago did you happen to check if there are any sub/sub-subtests which override ``cleanup()`` but don't call ``super()``?

dockertest/subtestbase.py
```diff
@@ -102,10 +161,84 @@ def cleanup(self):
...
+        """
+        If we've caught a DockerXxxx exception, see if it's a known problem
+        with this combination of subtest & docker NVR. If it is, change the
+        exception status from FAIL to WARN.
+        """
+        e_type, e_value, _ = sys.exc_info()
+
+        # A string comparison lets us catch any DockerXxxxx exception,
+        # e.g. DockerTestFail, DockerOutputError
+        if 'Docker' not in str(e_type):
```
cevich: This should probably be ``isinstance(e_type, xceptions.whatever)`` no?

dockertest/subtestbase.py
```diff
@@ -102,10 +161,84 @@ def cleanup(self):
...
+        # Children of autotest TestBaseException have an exit_status field;
+        # we only bypass if it's FAIL
+        if not hasattr(e_value, 'exit_status'):
+            return
+        if e_value.exit_status != 'FAIL':
+            return
+
+        # Yep, it's a test failure
+        # FIXME: is there a way to check if this is a new exception
+        #        coming from our module, vs an old one from a previous test?
```
cevich: To the best of my knowledge, this is incredibly difficult to do because exception-handling can also raise exceptions.  There's also weird things, like if an exception is thrown from an instances ``__del__()`` method, python changes it to a warning.

I'm willing to see how this plays out in reality, what you have here may be good enough.

A more "controlled" way to handle this situation via exceptions could be decorating a few key ``SubBase`` and ``Subtest`` base-methods with ``error.context_aware()`` (see example in ``autotest/client/shared/error.py``).  For example, ``run_once()`` and ``initialize()``.  Then those methods can use ``error.context()`` to apply set some magic string, which will automatically be attached to any raised exceptions.  Finally, this function can then use ``e_value.get_context()`` to validate the magic string to know from whence the exception came 😀

dockertest/subtestbase.py
```diff
@@ -8,10 +8,69 @@
...
+        if row and not row.startswith('#'):
+            try:
+                nvra, subtest, description = row.strip().split(None, 2)
+                if subtest not in known:
+                    known[subtest] = {}
+                known[subtest][nvra] = description
+            except ValueError:
+                SubBase.logwarning("Bad row in %s: %s"
+                                   % (known_failures_path, row))
+    return known
```
cevich: Would it be a major PITA to move the above functions into the ``environment`` module?

That's really it's purpose (inspecting/doing stuff with the os-environment), moving them later certainly would be a PITA.

I can give it a go and send you a patch if you're disinclined (unless there's a compelling reason to leave them here).

dockertest/subtestbase.py
edsantiago: Grrrrr. I knew this would be complicated. I even had a long comment there but it grew beyond any hope of sanity. Yes, `isinstance()` is how I started. I gave up on it.

First: for reasons I still don't understand, sometimes a test throws `xceptions.DockerTestFail` and sometimes `dockertest.xceptions.DockerTestFail`; and `isinstance()` isn't smart enough to realize that both are the same. I would have to test for both.

Second: it's not just `DockerTestFail`. I've also seen `DockerOutputError`. I realize that a string compare is ugly but the alternative I see is a long and fragile enumeration of exceptions. Maybe there's a cleaner way that I've missed.

dockertest/subtestbase.py
edsantiago: Excellent question. No, I hadn't. I've just run `ack -A10 'def cleanup' subtests/` and it looks like there are four:

    login/login.py
    ps_size/ps_size.py
    run_cgroup_parent/run_cgroup_parent.py
    top/top.py

Shall I submit an amendment to this PR? If so, can you remind me of the proper convention for calling `super()` within cleanup? I notice that most call it at the beginning (the usual) but a few call `super()` at the end; and ISTR something special about cleanup. Do you remember what that's about?

dockertest/subtestbase.py
cevich: >  I would have to test for both.

Ooohh right, I remember that strangeness now.  We ran into that in another lifetime (PR) w/o any conclusion.  I can live with what you have, but surely if I can forget, anybody can.  Maybe make a simple comment about this as a reminder.  How about:

```python
# xceptions.DockerXYZ makes isinstance() mad sometimes, cheat by looking at the name
```

?

dockertest/subtestbase.py
cevich: The decorator-way (suggestion above) would also address the broken ``isinstance()`` thing.  Just sayin.  Don't run out and do ``@context_aware()`` unless you think it's a good idea.  I'm fine if that's a later addition, or never.

dockertest/subtestbase.py
cevich: There may be a reason why they don't call ``super()``, or it was just an omission.  IIRC, in cases where they can't use ``super()`` I'm pretty sure I made an effort to comment about that.  In those cases, they just need to call ``self.overlook_known_failures()`` on their own (assuming ``super()`` isn't applicable).  If they don't override ``cleanup()`` at all, then there's nothing to be done, they'll automatically use the base-class method.  

Yeah, best fix them  in this PR, but if you're in a pinch or any look hairy, let me know and I'll help out.

edsantiago: Sorry, I wasn't clear. My question was: do you know why some places call `super()` at the beginning, and some at the end?

    def cleanup(self):
        super(MyBase, self).cleanup()           <---- the normal way

    def cleanup(self):
        ...do some stuff...
        ...do more stuff...
        super(MyBase, self).cleanup()           <---- at the end

dockertest/subtestbase.py
cevich: FWIW: My motivation is mainly about general code-orginization but ...wait a sec...no, I'm wrong.  Really the only function that should move to ``environment`` is ``docker_rpm()``.  The ``known_failure...`` functions are more specific to subtest than environment.  My bad.

edsantiago: Whew. Okay, thanks. Moving `known_failures` is turning out more complicated than I had first thought. (And I agree, known_failures belongs closer to subtest stuff). Undoing.

cevich: Okay, glanced at the unittests and code WRT ``docker_rpm()``.  Doesn't appear like it would be all that painful to move over to ``environment.docker_rpm()``.  Is that cool?

dockertest/subtestbase.py
cevich: This could very well be me getting my wires crossed WRT C++.  There, destruction always happens in the reverse (ancestor class method called last) because memory-freeing is involved.  In python, it really only matters if ``...do some stuff...`` is known to negatively affect the state (or behavior) expected by the base-class method.

Here for example, I think it only matters if  if ``...do some stuff...`` involves something that could lead to ``sys.exit(1)`` before ``overlook_known_failures()`` happens.

Otherwise, it's pretty rare to need to not call it not "The normal way", so most likely it was my goof and/or should be commented.  I have a unittest I'm working on elsewhere, that requires having the ``super()`` call in the middle of a ``testcase.setUp()``.  So it does happen.

dockertest/subtestbase.py
edsantiago: Done.

dockertest/subtestbase.py
edsantiago: Done.

dockertest/subtestbase.py
edsantiago: Still looking into `@error.context`. Looks really intriguing but I'm not entirely sure how I could take advantage of it. Thanks for the tip.

edsantiago:
Three new commits, each addressing one of your suggestions. Delay was from running a full autotest run on my virt. All looks good.

edsantiago:
Where do you suggest adding docs?

cevich:
@edsantiago Can you write up a blurb (a paragraph or so) documenting the file-format super clearly, so idiots like me can get it right 😀   I think I can copy-paste + tweak the commit-message into the docs as well.  So with those two in hand, adding docs should be easy.  I can send U a PR-for-the-PR (unless you wanna update ``index.rst`` yourself).

cevich:
> Where do you suggest adding docs?

Ugg, you're too fast for me...lessee...

cevich:
How about a new section like:

```rst
Filtering results of known-failures
-------------------------------------
```

at the end of  this (major) section:

```rst
Docker Autotest Control
=========================
```

?

cevich:
I s'pose under the ``Hacking`` section would work too.  But I prefer it closer to the beginning, and it seems to fit next to 'selecting tests'.

edsantiago:
On it.

cevich:
Ugg, lacking the time to properly test this, but feeling guilty about holding this up, I'm just going to merge it.  If there are problems or we need to revert, that can be done later/as needed.

cevich:
[(5) Silence known CI failures](https://trello.com/c/RLseD9il/658-5-silence-known-ci-failures)

PR #762, opened 2017-09-14
Remove annoying warning

cevich:
WARNING: Test results may be externally dependent! (See debug log for details

This message only points to a possible source of problems, it's
excessive logging at the WARNING level is basically just noise.

Since there's already a more specific warning at the DEBUG level, this
one can simply be killed outright.


cevich:
Heh...thanks for the...err review 😀

PR #764, opened 2017-09-19
Followup: removing images

edsantiago:
The previous PR was buggy: docker can't remove an image by ID
if the image is tagged more than once, e.g.

    docker.io/fedora     latest     b0b140824a48
    docker.io/fedora     7          b0b140824a48

The error is:

    conflict: unable to delete b0b140824a48 (must be forced) - image is referenced in multiple repositories

Solution: only remove by ID if the image has no tag. Otherwise
(default case) remove by fullname. This may not always work;
we may need to fall back to:

    docker rmi $(docker images -f dangling=true -q)

Unfortunately, I can't manage to reliably reproduce the
case where an image has a name but no tag; this means I
can't actually test this code.

### DIFF
--- a/dockertest/images.py
+++ b/dockertest/images.py
@@ -614,8 +614,15 @@ class DockerImages(object):
         Remove an image. This is simply a convenience function so
         callers don't need to access image_obj internals.
 
+        If the image has a tag, we try removing it by name: this is
+        robust when the same image might be tagged differently.
+        If there's no tag, as can happen during an incomplete
+        build subtest, try removing by ID.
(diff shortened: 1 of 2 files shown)


cevich:
Don't feel bad about struggling with this topic.  It's a bad design of Docker IMHO and certainly not the first time the name-vs-tag thing has caused LOTS of problems with not many good solutions.

Gaaa, I took the CI server down after moving it to anothe rcloud, and lost the results...okay, well, we can always revert 😀

PR #767, opened 2018-02-07
attach subtest: minimize race conditions

edsantiago:
The attach subtest fails frequently and irreproducibly; failure mode
is a timeout waiting for container to spin up. Here we bump up the
timeout, and (where possible) use wait_for_ready() to eliminate
unnecessary delays. There are still four instances of hardcoded
sleeps; these are harder to eliminate.

Also: removed unnecessary division of 'bash_cmd' and 'cmd'.

Also: improved readability of two diagnostic messages.


edsantiago:
Medium-low priority. Code isn't perfect - I've still seen one failure in my long-running tests - but it's better than before, and maybe good enough to reduce the false positives a little more.

subtests/docker_cli/attach/attach.py
```diff
@@ -95,7 +94,7 @@ def initialize(self):
         # Runs in background
         self.sub_stuff['cmdresult'] = dkrcmd.execute(run_in_pipe_r)
         self.sub_stuff['cmd_run'] = dkrcmd
-        self.wait_interactive_cmd()
+        dkrcmd.wait_for_ready(timeout=self.config['wait_interactive_cmd'])
```
cevich: I'm okay if you want to just make ``wait_interactive_cmd`` a global default.  Seems it's used in more than one place.  I'm doubtful that being able to specify different timeout values for different tests is ever going to be useful :wink:

cevich:
@edsantiago Still "Medium-low priority"?  Any thoughts on suggestion above?  Otherwise, this LGTM.

subtests/docker_cli/attach/attach.py
edsantiago: I think this may conflict with #772. If 772 looks good, merge that, and I'll take a fresh look at this one afterward. Thanks.

cevich: Yep, was reviewing PRs bottom-up so didn't know about that one yet :grin: .  Ya, I'd like to merge the other one since it updates all the tests, the name-difference isn't important to me.  But #772 needs a quick documentation/comment update.  After merging, fixing the conflict here should be really easy.

edsantiago:
rebased, removed the `timeout` arg from the new `wait_for_ready`s, and force-pushed

cevich:
Aye, that looks much better.

cevich:
Note to self:  Using github's online editor on someone else's PR quickly results in FUBAR :grinning:

PR #769, opened 2018-02-27
dockerhelp: fix nonfunctional test

edsantiago:
The dockerhelp subtest that was supposed to be checking
invalid commands ('docker --help --quiet', 'docker foobar')
was not actually doing anything useful. Remove nonuseful
code, replacing it with a simple check for output on
stderr: we don't check content, just check that something
was emitted on stderr.

Also, for proper help commands that should succeed,
make sure exit status is 0.


cevich:
I agree, the catch-all ``negativeusage`` should be used for those negative checks...

cevich:
LGTM.

PR #770, opened 2018-03-21
OutputGood.error_check(): check *all* output

edsantiago:
Bug fix: code has only been checking the first line.


cevich:
@edsantiago LGTM, just needs a quick/easy rebase.

cevich:
(I think I did that right)

PR #772, opened 2018-03-28
wait_for_ready(): refactor

edsantiago:
add a common 'wait_ready' setting (default: 60s). Allow individual
subtest .ini files to override, and continue allowing overriding
in code itself.

Goal is to eliminate various 5- and 10-second timeouts: these are
causing sporadic and irreproducible test failures.


config_defaults/defaults.ini
```diff
@@ -22,6 +22,9 @@ docker_options =
 #: Max runtime in seconds for any docker command (auto-converts to float)
 docker_timeout = 300.0
 
+#: Time (seconds) to wait for READY from container
```
cevich: Small thing: Might be nice to future developers (stolen phrase) to mention this is not automatic, it's for use with ``dockercmd.wait_for_ready()``.

cevich: How about:

```
#: Time (seconds) to wait for "READY" from the container, when using
#: ``dockercmd.AsyncDockerCmd.wait_for_ready()``
```

cevich: Actually...in ``wait_for_ready()`` it has access to the test configuration.  Would it make sense if that method used ``self.subtest.config["wait_ready"]`` as the default value ``if timeout is None``?

cevich: Yeah, looks like it.  Right now the default is ``self.timeout`` which IIRC is the 'docker_timeout' key...that's probably NOT what we want, using your new value would be much better IMHO.

edsantiago: I'm confused. Isn't that exactly what this PR is doing?

cevich: lol durrr....just looked at code.  I totally missed that change in dockercmd.

cevich: How about:

```
#: Default timeout (seconds) to use, when calling
#: ``dockercmd.AsyncDockerCmd.wait_for_ready()`` with ``timeout=None``.
```

edsantiago: Force-pushed with updated comment.

PR #773, opened 2018-04-04
liverestore: deal with docker-containerd

edsantiago:
docker on RHEL is one systemd service; in fedora it's two.
docker-containerd is a separate unit. This wrecks an assumption
I had made in the liverestore test, that a container process PID
would be a child of dockerd. It's too cumbersome to deal with
either case, so just abandon the heredity check.

Extra wrinkle: restart docker-containerd. Necessary because
on these systems, restarting docker does not restart containerd
and hence is not a useful test of liverestore. We do a simple
systemctl restart, and deliberately ignore exit status because
we know it will fail on RHEL where there's no such unit.

Final wrinkle: restarting docker-containerd also restarts docker,
and I've seen runs where 'docker ps' in postprocess() fails
before docker is not fully started. Run 'systemctl is-active'
to make sure docker daemon is fully back before we continue.
This requires opening a formerly private function.


cevich:
(Haven't looked at code yet, only description)  Couldn't final-wrinkle fix simplify extra-wrinkle?
Meaning,  on all platforms, try restarting docker-containerd, if that fails, ignore it.  Restart dockerd, then `systemctl is-active dockerd`?

> assumption I had made in the liverestore test, that a container process PID would be a child of dockerd.

I vaguely recall there is/was another test that makes this assumption.  It may have been deleted though, it was years ago.

cevich:
Ahh, nvm, I think that would just be the same thing just in a different order.

cevich:
Okay, LGTM.

PR #776, opened 2018-05-14
selinux_labels: fix deprecated usage

edsantiago:
The docker '--security-opt' option requires 'label=xxx'; but
for historical reasons accepts the now-deprecated 'label:xxx'
format. We've been using the latter; switch to the new form.


edsantiago:
@cevich PTAL when convenient (low priority)

cevich:
LGTM.

PR #777, opened 2018-05-16
import/truncated test: more readable diagnostics

edsantiago:
If the import command *doesn't* fail, give a better
indication of why (i.e. the error output or status).
Previously it was difficult to know what had failed.


edsantiago:
@cevich PTAL; this is cherry-picked from my podman_fixes branch because it's unrelated to podman

PR #778, opened 2018-05-16
typo fix: parse_columns

edsantiago:
Fix a misspelled function name.


edsantiago:
@cevich PTAL, this should be a quick one

PR #779, opened 2018-05-16
Command-line cleanliness: use --hostname, not -h

edsantiago:
...in the systemd_in_container subtest. Primarily because
podman (correctly, IMO) uses -h to mean --help, so this
gives us podman compatibility, but also for legibility.


edsantiago:
@cevich another softball; I'm just trying to clear out the small not-really-podman changes

PR #780, opened 2018-05-17
'docker inspect': don't assume leading / in Name

edsantiago:
'docker inspect' prepends '/' to the container Name.
podman doesn't. Instead of blindly removing the first
character, use lstrip('/') to deal with either.


edsantiago:
It's been `[1:]` since the initial commit in January 2015. This is just a little safer. Thanks!

PR #781, opened 2018-05-17
rmi subtest: remove unneeded postprocess

edsantiago:
This is essentially a dup of the parent class's postprocess.
Removing in preparation for upcoming podman changes, where
slight tweaks will be needed to that postprocess.


edsantiago:
@cevich one more small one. Although this `postprocess` isn't exactly a duplicate of the parent, it's the same functionality. Removing this now is useful as cleanup whether or not we incorporate the podman changes.

PR #783, opened 2018-05-31
Fix "key rpms" generated by the log_versions

cevich:


cevich:
Confirmed the podman RPM version is being collected in ``results/default/sysinfo/key_rpms``.

PR #785, opened 2019-07-10
New subtest: cp_symlink, tests copying into container

edsantiago:
...especially into a path pointed to by a symlink.
Confirms two CVEs, plus regressions introduced in
the fixing thereof


edsantiago:
New corner cases in cp. Tested with older docker, with patched version from Giuseppe, and with podman.

edsantiago:
@cevich looks like CI is broken. Any chance of merging this in anyway?

cevich:
I'm totally not surprised.  Ya, one sec let me get that out of the way...

PR #786, opened 2019-08-27
rhbz1741718 - docker cp to /var/lib/foo -> /foo

edsantiago:
If docker storage is in /var/lib/docker/<etc> (default) and
you try to cp a file into <container>:/var/lib/myfile, docker
strips off the /var/lib/ prefix and places the file as /myfile.

New test confirmed to fail under docker-1.13.1-103.git7f2769b.el7
and to pass under hand-patched docker [1]

 [1] https://github.com/projectatomic/docker/pull/357

Since code is almost identical to cp_symlink(), I refactored
that a little bit and added some config_default settings.


cevich (approved):
LGTM

